PR #16, opened 2024-12-05
chore(deps): rustls and hashbrown

storopoli:
hashbrown to 0.15.2
rustls to 0.23.19

closes https://github.com/alpenlabs/strata-bridge-poc/security/dependabot/5
closes https://github.com/alpenlabs/strata-bridge-poc/security/dependabot/6
closes https://github.com/alpenlabs/strata-bridge-poc/security/dependabot/4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -517,7 +517,7 @@ dependencies = [
  "const-hex",
  "derive_more 1.0.0",
  "foldhash",
- "hashbrown 0.15.0",
+ "hashbrown 0.15.2",
  "hex-literal",
  "indexmap 2.6.0",
  "itoa",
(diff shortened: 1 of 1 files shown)


Rajil1213 (changes requested):
Looks like `hashbrown` still needs updating. It's not part of the diff.

In the CI, `cargo-audit` has been allowed to fail due to a dependency on `sqlx-mysql` (that we aren't even using but `cargo` still keeps in the lockfile). We need a longer term strategy for how to deal with this so that we don't miss any critical vulnerabilities. But that is for later.

For now, the only thing required in this PR is to update `hashbrown` too.

storopoli:
Thanks for pointing that out. It is now included.

Rajil1213 (approved):

PR #21, opened 2025-01-15
Implement Block Tracker

ProofOfKeags:
## Description

This PR implements a reorg resistant blockchain tracker that should correctly track the full lifecycle of any transactions that a consumer wants to monitor. The PR should be reviewed in stages, giving attention to different aspects during progression. This should help reduce the amount of unnecessary review cycles, allowing us to solve the bigger, more foundational problems first, and move onto refinement after.

- [x] Interface design
- [x] Property testing completeness
- [x] Implementation correctness (edge case handling)
- [x] Algorithmic implementation efficiency
- [x] Documentation clarity, correctness, and completeness
- [x] Code organization (which files/order of presentation within files)
- [ ] Silicon implementation efficiency

### Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update
-   [ ] Refactor

## Checklist

-   [x] I have performed a self-review of my code.
-   [x] I have commented my code where necessary.
-   [x] I have updated the documentation if needed.
-   [x] My changes do not introduce new warnings.
-   [x] I have added tests that prove my changes are effective or that my feature works.
-   [x] New and existing tests pass with my changes.

## Related Issues

STR-940


crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+
+    /// Updates the BtcZmqConfig with a zmqpubsequence connection string and returns the updated config. Useful for a
+    /// builder pattern with dotchaining.
+    pub fn with_sequence_connection_string(mut self, s: &str) -> Self {
+        self.sequence_connection_string = Some(s.to_string());
+        self
+    }
+}
+
+#[derive(Clone)]
```
storopoli, line 96: Always derive `Debug` as well.

storopoli, line 96: Always derive `Debug` as well.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+        self
+    }
+
+    /// Updates the BtcZmqConfig with a zmqpubsequence connection string and returns the updated config. Useful for a
+    /// builder pattern with dotchaining.
+    pub fn with_sequence_connection_string(mut self, s: &str) -> Self {
+        self.sequence_connection_string = Some(s.to_string());
+        self
+    }
+}
```
storopoli, line 94: This is the builder pattern. I like it. But it is missing the `.build()` method.
See https://rust-unofficial.github.io/patterns/patterns/creational/builder.html

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+
+impl<T> Stream for Subscription<T> {
+    type Item = T;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        self.get_mut().receiver.poll_recv(cx)
+    }
+}
```
storopoli, line 122: Do we need trait bounds (or trait markers) in `T`.
I'm not sure, but having a bound is always a good idea, if applicable.

crates/btc-notify/src/lib.rs
ProofOfKeags: Why does it need one?

crates/btc-notify/src/lib.rs
ProofOfKeags: I don't know why we'd want to constrain it at all. The construction works independent of T

ProofOfKeags: > having a bound is always a good idea, if applicable

Hard disagree. Code should be written in as general way as possible. The trait bound should only be invoked if the essential aspects of that trait are used. In this case, the subscription is a stream irrespective of the contents of the subscription. This is a somewhat trivial case since it's really just a trivial indirection from the underlying tokio channel but the principle still holds. Regardless of what you are subscribing to, the subscription behaves like a stream.

storopoli: Doesn't this `T` needs to be `Sync`?
Yeah your argument makes sense. We can keep `T` as it is if it works fine

crates/btc-notify/src/lib.rs
storopoli: Because then `build() -> BtcZmq`, no?

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+use tokio::sync::Mutex;
+use tokio::task::JoinHandle;
+use std::collections::BTreeMap;
+use std::collections::VecDeque;
+use std::error::Error;
+use std::ops::Sub;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::Context;
+use std::task::Poll;
```
Rajil1213, line 20: `rustfmt` should have group these more nicely, I think.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+
+/// BtcZmqConfig is the main configuration type used to establish the connection with the ZMQ interface of Bitcoin. It
+/// accepts independent connection strings for each of the stream types. Any connection strings that are left as None
+/// when initializing the BtcZmqClient will result in those streams going unmonitored. In the limit, this means that the
+/// default BtcZmqConfig will result in a BtcZmqClient that does absolutely nothing (NOOP).
+pub struct BtcZmqConfig {
```
Rajil1213, line 26: We generally structure these docs to have a single header line followed by a longer paragraph. There is also a [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph) for this that is allowed by default but something we might deny later. Tidying this up now will make our work much easier later.

The content itself is top-notch though!

crates/btc-notify/src/lib.rs
Rajil1213: This, as it stands right now, isn't exactly the builder pattern since there is no intermediate type to call `.build` upon to get the actual type that is being built. That said, you might want to check out [bon](https://github.com/elastio/bon) to reduce some of the boilerplate involved.

The nicest way to do this is to use the typestate pattern to only expose accessors if the corresponding value has been set. But that is a massive overkill for this.

crates/btc-notify/src/lib.rs
Rajil1213: Normally, I would be tempted to at least constrain the generic to `Sized`. But that's mostly because when using an unconstrained generic, I usually run into issues that is solved by constraining it with `Sized`. If that isn't the problem here, I'm leaving it unconstrained.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn setup() -> Result<(BtcZmqClient, corepc_node::Node), Box<dyn std::error::Error>> {
+        let mut bitcoin_conf = corepc_node::Conf::default();
+        bitcoin_conf.view_stdout = true;
```
Rajil1213, line 417: You should be able to get away with just setting `RUST_LOG` to `trace` for debugging purposes. Is there any reason we need the `stdout` too though?

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+
+    #[tokio::test]
+    async fn basic_subscribe_blocks_functionality() -> Result<(), Box<dyn std::error::Error>> {
+        let (mut client, bitcoind) = setup()?;
+
+        let mut block_sub = client.subscribe_blocks().await;
+        let newly_mined = bitcoind.client.generate_to_address(1, &bitcoind.client.new_address()?)?.into_model()?;
+        let blk = block_sub.next().await.map(|b|b.block_hash());
+        assert_eq!(newly_mined.0.first(), blk.as_ref());
+        drop(client);
```
Rajil1213, line 449: Is this explicit `drop` necessary? If we are doing this, should we also check if the thread has indeed been aborted?

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+                        // TODO: figure out how to make this a reference into the block we save to avoid duplicating
+                        // transaction data.
+                        raw: Some(matched_tx.clone()),
+                        block: Some(block.block_hash()),
+                        seq_received: false,
+                    };
+                    self.tx_lifecycles.insert(matched_tx.compute_txid(), lifecycle);
+                }
+                Some(lifecycle) => {
+                    match (&lifecycle.raw, lifecycle.block, lifecycle.seq_received) {
```
Rajil1213, line 331: Looks like there are a bunch of runtime invariants that we are trying to handle. The comments definitely help but is there a way to encode this at compile time instead? Perhaps with enums with struct variants?

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+            tx_subs,
+            state_machine,
+            thread_handle,
+        })
+    }
+
+    pub async fn subscribe_transactions(&mut self, f: impl Fn(&Transaction) -> bool + Sync + Send + 'static) ->
+        Subscription<(Transaction, TxStatus)> {
+
+        let (send, recv) = mpsc::channel(4);
```
Rajil1213, line 236: Is there any specific reason to go with this queue size?. Same question for the choice of `10` in the `subscribe_blocks` method below.

Rajil1213 (commented):
Left a few cosmetic nits and some questions.

I like how the subscription API is coming along. Definitely needs a bunch of tests though. There is a lot of async coordination going on that I couldn't completely reason about looking at the code alone.

crates/btc-notify/src/lib.rs
ProofOfKeags: I'll address before finality but I want to focus on things in the following order:

1. Interface design
2. Property testing completeness
3. Implementation correctness (edge case handling)
4. Algorithmic implementation efficiency
5. Documentation clarity, correctness, and completeness
6. Code organization (which files/order of presentation within files)
7. Silicon implementation efficiency

I still consider this to be in stage 1. Feel free to leave feedback on any of the stages but I'll address in roughly this order. To the extent that you can focus review on the stage we're in (currently 1 and 2), that would be ideal. If something is preventing you from properly evaluating those two, let me know.

I'll add this stage checklist to the main body of the PR.

crates/btc-notify/src/lib.rs
ProofOfKeags: No they were just best guesses that I went with to start. I'll drop a TODO in there to give a more careful thought to the magick numbers.

crates/btc-notify/src/lib.rs
ProofOfKeags: Completely agree with this push. The answer is I believe that we can. As I was going through implementation I think I figured out a way to do a CBC encoding of this to reduce and possibly eliminate the runtime invariants, but I wanted to complete a full sketch before going back and refining.

crates/btc-notify/src/lib.rs
ProofOfKeags: This was for my own ability to debug stuff since I wanted to ensure that the bitcoind side was actually publishing on the zmq interface.

Can you clarify whatyou mean by:
> You should be able to get away with just setting RUST_LOG to trace for debugging purposes.

crates/btc-notify/src/lib.rs
ProofOfKeags: This is to ensure that rustc does not get cute with doing a premature drop, causing the producer thread to terminate prematurely (by virtue of its drop trait)

ProofOfKeags: > should we also check if the thread has indeed been aborted?

That's a great property to test. I'd have to think about how to design such a test though.

crates/btc-notify/src/lib.rs
Rajil1213: That's fine. I only wanted to flag this because it's something that your editor should take care of for you instead of having to dedicate actual hours on this.

crates/btc-notify/src/lib.rs
Rajil1213: You can set the environment variable to `RUST_LOG` to `trace` and that should produce a lot more logs from `corepc-node` including the actual requests and response. This requires that you first initialize the logger which you can do via `strata_common::logging::init(...)`. More on this here: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

crates/btc-notify/src/lib.rs
Rajil1213: The simplest way to change the visible of the `JoinHandle` to `pub(crate)` and then, check for `is_canceled`.

crates/btc-notify/src/lib.rs
storopoli: Wow `bon` seems awesome :)

crates/btc-notify/src/lib.rs
storopoli: Yeah the content is REALLY good.

@ProofOfKeags check the engineering best practices for docstrings in this section of our notion doc: https://www.notion.so/Engineering-Rust-Best-Practices-14c901ba000f8025a974ec8e2a70b4af?pvs=4#14c901ba000f80b38a1fc290ef3a900a

crates/btc-notify/src/lib.rs
ProofOfKeags: Yeah I can definitely see sized being something that kicks in as a requirement at some point so you may be right.

crates/btc-notify/src/lib.rs
ProofOfKeags: I think the other test design here would be to ensure that all subscription channels have their send sides closed when the client is dropped. Because the subscription doesn't contain references to the client, it isn't automatically invalidated by lifetime analysis (which is probably a good thing). I'm hesitant to leak the JoinHandle since it breaks some of the abstraction. I'm focusing on tests right now so I'll probably have better ideas in the next day or so here.

Rajil1213 (changes requested):
Amazing work with the tests!

The API looks good now. I've left a few nits and suggested some type improvements but otherwise looks good.

crates/btc-notify/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "btc-notify"
+version = "0.1.0"
+edition = "2021"
+
```
Rajil1213, line 5: Could you add this section here as well?

```rustrust.missing_debug_implementations = "warn"
rust.missing_docs = "warn"
rust.unreachable_pub = "warn"
rust.unused_must_use = "deny"
rustdoc.all = "warn"
```
This is gonna be pretty annoying in the beginning but this is the right time to do this and it might help resolve some of the cosmetic nits from the earlier round of reviews.

Rajil1213 (changes requested):
One more comment that I forgot to add earlier.

crates/btc-notify/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+bitcoin.workspace = true
+bitcoincore-zmq = { version = "1.5.2", features = ["async"] }
+itertools = "0.14.0"
+tokio = { workspace = true, features = ["macros", "rt", "sync", "test-util"] }
+futures = "0.3.31"
+zmq = "0.10.0"
+
+[dev-dependencies]
+corepc-node = "0.5.0"
```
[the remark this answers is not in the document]
storopoli: Ideally we would just add the `corepc-node` only in `dev-dependencies` which would exclude it being compiled for `rust-analyzer` and simple `cargo {check,build}` etc.

We don't have workspace-level `dev-dependency`s hence it needs to be added to the workspace `dependency`s which will bloat the `cargo {check,build}`. I think what is being done here should be the guideline for the repo.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+                    raw: tx,
+                    block: None,
+                };
+
+                self.tx_lifecycles.insert(txid, Some(lifecycle));
+
+                // We intentionally DO NOT return the transaction here in the diff because we are unsure of what the
+                // status is. We will either immediately get a followup block or a followup sequence which will cause us
+                // to emit a new state change.
+                Vec::new()
```
storopoli, line 452: Same idea with the optional return value.
You wouldn't need to return an allocated vec.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+
+                self.tx_lifecycles.insert(txid, Some(lifecycle));
+
+                // Presence within the map indicates we have already received the sequence message for this but don't
+                // yet have any other information, indicating that this rawtx event can generate the Mempool event.
+                vec![(tx, TxStatus::Mempool)]
+            }
+            // In this case we know everything we need to about this transaction, and this is probably a rawtx event
+            // that accompanies an upcoming new block event.
+            Some(Some(_)) => Vec::new(),
```
storopoli, line 470: Same idea with the optional return value.
You wouldn't need to return an allocated vec.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+        // Set up new bitcoind and zmq client instance.
+        let (mut client, mut bitcoind) = setup()?;
+
+        // Get a new address that we will use to send money to.
+        let new_address = bitcoind.client.new_address()?;
+
+        // Mine 101 new blocks to that same address. We use 101 so that the coins minted in the first block can be spent
+        // which we will need to do for the remainder of the test.
+        let _ = bitcoind.client.generate_to_address(101, &new_address)?.into_model()?;
+        tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```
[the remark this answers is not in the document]
storopoli: Probably fine to reduce to `1`. Or even `0.5` (using `from_millis`)

storopoli (changes requested):
Some comments.
Haven't had time to go over the docs.

Focused in 

> " I’m requesting a review of the design approach, paying special attention to the TODOs and NOTEs"

crates/btc-notify/Cargo.toml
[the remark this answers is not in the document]
Rajil1213: I think `cargo` is smart enough to not include any unused dependencies even if they _are_ specified at the workspace level. If the dependencies specified are only present in the `dev-dependencies` section in each of the member crates, it should not be compiled unless we pass in the `--tests` flag. (I verified this locally but please feel free to verify this yourselves too, didn't spend too much time).

Either way, the other PR I'm working on introduces `corepc-node` as a dependency of the `test-utils` crate. So it should be fine to just use the workspace version here.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+        Subscription {
+            receiver,
+        }
+    }
+}
+
+impl<T> Stream for Subscription<T> {
+    type Item = T;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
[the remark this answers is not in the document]
Rajil1213: Ahh I see. My bad. Will resolve this.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: I mean why is this sleep necessary at all?

crates/btc-notify/Cargo.toml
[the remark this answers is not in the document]
storopoli: > If the dependencies specified are only present in the `dev-dependencies` section in each of the member crates, it should not be compiled unless we pass in the `--tests` flag. (I verified this locally but please feel free to verify this yourselves too, I didn't spend too much time).

Yes



> I think `cargo` is smart enough to not include any unused dependencies even if they _are_ specified at the workspace level.

Yes, but most of the time we do a bunch of `cargo CMD --workspace` in bash history and also in Makefiles.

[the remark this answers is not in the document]
Rajil1213: > Yes, but most of the time we do a bunch of cargo CMD --workspace in bash history and also in Makefiles.

I mean even if we have `--workspace`, it does not compile dependencies that are not used as `dependencies` in any of the member crates.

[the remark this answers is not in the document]
storopoli: I don't think so

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+    }
+}
+
+/// This structure is here so that we can keep track of messages coming in on parallel streams that are all
+/// tracking the same underlying event. Depending on the messages we receive and in what order we track the transaction
+/// all the way to block inclusion, inferring other states depending on the messages we have received.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct TxLifecycle {
+    raw: Transaction,
+    block: Option<BlockHash>,
```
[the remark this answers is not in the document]
ProofOfKeags: What purpose does adding the height serve here?

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
ProofOfKeags: Bitcoind does not fully process all 101 blocks before returning. It merely generates all of their data, but without a sleep I observed that we miss certain events on the ZMQ interface.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+use std::task::Context;
+use std::task::Poll;
+
+/// BtcZmqConfig is the main configuration type used to establish the connection with the ZMQ interface of Bitcoin. It
+/// accepts independent connection strings for each of the stream types. Any connection strings that are left as None
+/// when initializing the BtcZmqClient will result in those streams going unmonitored. In the limit, this means that the
+/// default BtcZmqConfig will result in a BtcZmqClient that does absolutely nothing (NOOP).
+pub struct BtcZmqConfig {
+    /// depth at which a transaction is considered buried, defaults to 6
+    bury_depth: usize,
```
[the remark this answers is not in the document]
ProofOfKeags: @Rajil1213 has always referred to this as "buried" and I like the term, but I can use "finality depth" and "Final" instead. I don't care so much.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+
+        let mut stream = bitcoincore_zmq::subscribe_async(&sockets)?;
+
+        let block_subs = Arc::new(Mutex::new(Vec::<mpsc::UnboundedSender<Block>>::new()));
+        let block_subs_thread = block_subs.clone();
+        let tx_subs = Arc::new(Mutex::new(Vec::<TxSubscriptionDetails>::new()));
+        let tx_subs_thread = tx_subs.clone();
+        let state_machine_thread = state_machine.clone();
+        let thread_handle = tokio::task::spawn(async move {
+            loop {
```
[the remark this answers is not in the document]
ProofOfKeags: When should we break, then? There's no sensible place to break. This loop is aborted from the outside from the `thread_handle` which is aborted when the client is dropped.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+                            block_subs_thread.lock().await.retain(|sub| sub.send(block.clone()).is_ok());
+
+                            // Now we process the block to understand what the relevant transaction diff is.
+                            sm.process_block(block)
+                        },
+                        Ok(Message::Tx(tx, _)) => sm.process_tx(tx),
+                        Ok(Message::Sequence(seq, _)) => sm.process_sequence(seq),
+                        Err(e) => {
+                            eprintln!("ERROR: {e:?}");
+                            Vec::new()
```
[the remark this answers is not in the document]
ProofOfKeags: An option will allocate similarly, will it not? It will also introduce a duplicate zero value. What is the semantic difference between Some(vec![]) and None? I don't think we should bother with this unless we find that perf becomes an issue.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+
+        Ok(BtcZmqClient {
+            block_subs,
+            tx_subs,
+            state_machine,
+            thread_handle,
+        })
+    }
+
+    pub async fn subscribe_transactions(&mut self, f: impl Fn(&Transaction) -> bool + Sync + Send + 'static) ->
```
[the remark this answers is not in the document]
ProofOfKeags: Not sure why we need it, just that rust told me it couldn't proceed without it being 'static. My guess is that we can't put it inside the Arc unless we own it, and owned values have the 'static lifetime.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+        Subscription<(Transaction, TxStatus)> {
+
+        let (send, recv) = mpsc::unbounded_channel();
```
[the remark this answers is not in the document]
ProofOfKeags: I don't understand what this is asking.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+        let details = TxSubscriptionDetails {
+            predicate: Arc::new(f),
+            outbox: send,
+        };
+
+        let mut subs = self.tx_subs.lock().await;
+        let mut sm = self.state_machine.lock().await;
+        sm.add_filter(details.predicate.clone());
+        subs.push(details);
+
```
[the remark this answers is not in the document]
ProofOfKeags: doesn't rustc do this automatically? I've seen countless times where drops are aggressively hoisted to their earliest possible time.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+        Subscription::from_receiver(recv)
+    }
+
+    pub async fn subscribe_blocks(&mut self) -> Subscription<Block> {
+        let (send, recv) = mpsc::unbounded_channel();
+
+        self.block_subs.lock().await.push(send);
```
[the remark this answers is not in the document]
ProofOfKeags: I prefer explicit drops, but the point about eagerly and explicitly dropping the lock is well-taken.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+            // backfill the block history at startup using the RPC interface, or accepting the blocks newer than the
+            // bury depth as an argument to the constructor.
+            None => {
+                self.unburied_blocks.push_front(block)
+            }
+        }
+        let block = self.unburied_blocks.front().unwrap();
+
+        // Now we allocate a Vec will collect the net-new transaction states that need to be distributed.
+        let mut diff = Vec::new();
```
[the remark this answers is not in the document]
ProofOfKeags: Maybe don't optimize without profiling? :)

crates/btc-notify/Cargo.toml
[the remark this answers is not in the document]
ProofOfKeags: Perhaps this is dumb but afaict, there is no workspace dependency for `corepc-node`

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+    // rm_filter takes a TxPredicate that was previously added via add_filter.
+    fn rm_filter(&mut self, pred: &TxPredicate) {
+        if let Some(idx) = self.tx_filters.iter().position(|p| Arc::ptr_eq(p, pred)) {
+            self.tx_filters.swap_remove(idx);
+        }
+    }
+
+    // process_block is one of the three primary state transition functions of the BtcZmqSM, updating internal state to
+    // reflect the contents of the block.
+    fn process_block(&mut self, block: Block) -> Vec<(Transaction, TxStatus)> {
```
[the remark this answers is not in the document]
ProofOfKeags: Yeah no problem.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+
+        // Subscribe to all non-coinbase transactions.
+        let pred = |tx: &Transaction| !tx.is_coinbase();
+        let mut tx_sub = client.subscribe_transactions(pred).await;
+
+        // Send a non-coinbase transaction, remembering its txid.
+        let txid = bitcoind.client.send_to_address(&new_address, bitcoin::Amount::ONE_BTC).unwrap().txid()?;
+
+        // Kick off a mining task that will mine a new block every 100ms until we tell it to stop.
+        let stop = Arc::new(std::sync::atomic::AtomicBool::new(true));
```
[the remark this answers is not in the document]
ProofOfKeags: Looks like we can but we have to make the task cooperative and check for `tokio::task::is_cancelled()`

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
ProofOfKeags: image

crates/btc-notify/Cargo.toml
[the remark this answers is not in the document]
Rajil1213: > I don't think so

Did you verify this with a simple repository? I've tried this and it does not compile dependencies that are unused (or only used as dev-dependencies).

> Perhaps this is dumb but afaict, there is no workspace dependency for corepc-node

You might need to rebase. It was merged yesterday.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: I see. Thanks for pointing that out. Could you also add a comment there?

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: Aborting feels more idiomatic than managing it yourself unless it involves a more complicated refactor here.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: Yeah I'm okay with `final` as well.

PS: I didn't know I was the only using that term. 😅

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: I think Jose is asking to change the API so that the consumer is responsible for creating a channel rather than the method creating a new channel per call -- essentially one-channel-per-consumer vs one-channel-per-subscription is the difference here. But this might entail a more substantial refactor because it is the `self` that holds the `outbox`.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: maybe adding a comment then, saying that the `loop` is broken when the `client` is dropped?

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: TIL, thanks!

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: No, that's because tokio cannot reason the compiler about lifetimes properly with respect to the work stealing. Everything needs to be `'static` to have the compiler guarantees.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: Yes, I think this is good for now. But in the future we might have some issues with performance. Then, we can think about this refactor.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: No, it doesn't.
The lock is dropped like any other borrow checked/owned value.
When it goes out of scope or when it's explicitly `drop`ped.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: Hahahaha, yeah. Good point!

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: This is very difficult to verify but doesn't rustc shorten the lifetime even without an explicit scope (`{ ... }`)?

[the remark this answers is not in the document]
storopoli: > This is very difficult to verify but doesn't rustc shorten the lifetime even without an explicit scope (`{ ... }`)?

I don't think that you have a guarantee for that.



> Put another way, when does the scope end? When a borrowed value is no longer used or when there is a `drop` call or closing of its scope with `}`?

Scope ends with `}`. Then all the values inside (that are "owned" by the scope) are "destroyed" by calling recursively the `drop` method. This is the "drop check".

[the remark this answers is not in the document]
Rajil1213: > Scope ends with }. Then all the values inside (that are "owned" by the scope) are "destroyed" by calling recursively the drop method. This is the "drop check".

You're right! Turns out `rustc` is smart about lifetimes but not about scopes. TIL.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
ProofOfKeags: Leaving as buried because I think it's a better term

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
ProofOfKeags: I think the solution to that is to allow adding filters to existing subscriptions. But I do understand now that you're saying that you want to have a monoidal structure on Subscriptions that allows for resource cleanup while preserving disjunction semantics.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
ProofOfKeags: I have observed rustc hoisting drops, which is why the explicit drops for the client are included towards the end of tests. However, I'm aware that this is probably UB and we are not *promised* that it will aggressively drop it and so taking responsibility for that seems prudent.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+        let thread_handle = tokio::task::spawn(async move {
+            loop {
+                while let Some(res) = stream.next().await {
+                    let mut sm = state_machine_thread.lock().await;
+                    let diff = match res {
+                        Ok(Message::HashBlock(_, _)) => { Vec::new() }
+                        Ok(Message::HashTx(_, _)) => { Vec::new() }
+                        Ok(Message::Block(block, _)) => {
+                            // First send the block to the block subscribers.
+                            block_subs_thread.lock().await.retain(|sub| sub.send(block.clone()).is_ok());
```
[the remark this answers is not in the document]
ProofOfKeags: I asked claude about this and its answer matches what I've seen rustc do which is that intermediate values in a dotchain are dropped at the end of the statement.

> are intermediate dotchained values dropped when the statement ends?

---

Yes, intermediate values in dot-chained expressions are dropped at the end of the statement. Here's a demonstrative example:

```rust
struct Droppable(i32);

impl Droppable {
    fn new(i: i32) -> Self {
        println!("Creating {}", i);
        Self(i)
    }
    
    fn next(self) -> Droppable {
        println!("Converting {} to {}", self.0, self.0 + 1);
        Droppable(self.0 + 1)
    }
}

impl Drop for Droppable {
    fn drop(&mut self) {
        println!("Dropping {}", self.0);
    }
}

fn main() {
    // Each intermediate value is dropped after its use
    Droppable::new(1).next().next().next();
}
```

This will output something like:
```
Creating 1
Converting 1 to 2
Dropping 1
Converting 2 to 3
Dropping 2
Converting 3 to 4
Dropping 3
Dropping 4
```

Each intermediate value is dropped as soon as it's no longer needed in the chain. This is important for memory management and also for understanding when destructors will be called.

This behavior also means you need to be careful when chaining methods that take ownership - once a value is consumed in the chain, you can't use it again later in the same chain.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+impl Drop for BtcZmqClient {
+    fn drop(&mut self) {
+        self.thread_handle.abort();
+    }
+}
+
+impl BtcZmqClient {
+    pub fn connect(cfg: BtcZmqConfig) -> Result<Self, Box<dyn Error>> {
```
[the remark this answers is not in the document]
ProofOfKeags: I actually think this is worth taking care of in this PR. is `thiserror` your preferred library for this?

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1625 @@
+#![cfg_attr(coverage_nightly, feature(coverage_attribute))]
```
Rajil1213, line 1: Are these necessary? `cargo-nextest` should already do this for you, I think.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1625 @@
...
+use bitcoin::{transaction::Transaction, Block, BlockHash, Txid};
+use bitcoincore_zmq::{Message, SequenceMessage};
+use futures::StreamExt;
+use log::error;
+use tokio::{
+    sync::{mpsc, Mutex},
+    task::JoinHandle,
+};
+
+const DEFAULT_BURY_DEPTH: usize = 6;
```
Rajil1213, line 27: We might want to move this to configurable params down the line (for example, if we deploy to Testnet4 where larger reorgs are possible). Could you add a `TODO` here?

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1625 @@
...
+/// It accepts independent connection strings for each of the stream types. Any connection strings
+/// that are left as None when initializing the BtcZmqClient will result in those streams going
+/// unmonitored. In the limit, this means that the default BtcZmqConfig will result in a
+/// BtcZmqClient that does absolutely nothing (NOOP).
+///
+/// You should construct a BtcZmqConfig with [`Default::default`] and modify it with the member
+/// methods on this struct.
+#[derive(Debug, Clone)]
+pub struct BtcZmqConfig {
+    /// depth at which a transaction is considered buried, defaults to 6
```
Rajil1213, line 41:
```suggestion
    /// depth at which a transaction is considered buried, defaults to [`DEFAULT_BURY_DEPTH`].
```

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1625 @@
...
+/// The choice of using an arc here is intentional so that we can directly compare these predicates
+/// (via [`Arc::ptr_eq`]) when managing the active subscription set.
+type TxPredicate = Arc<dyn Fn(&Transaction) -> bool + Sync + Send>;
+
+struct TxSubscriptionDetails {
+    predicate: TxPredicate,
+    outbox: mpsc::UnboundedSender<TxEvent>,
+}
+
+#[cfg_attr(coverage_nightly, coverage(off))]
```
Rajil1213, line 209: It's pretty obvious here but we should try adding comments when we are ignoring code sections from the coverage report.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
Rajil1213: Yes. Or at least, that's the one being used elsewhere in this repo as well as in `strata`.

crates/btc-notify/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+rust.unreachable_pub = "warn"
+rust.unused_crate_dependencies = "deny"
+rust.unused_must_use = "deny"
+rustdoc.all = "warn"
+
+[dependencies]
+bitcoin.workspace = true
+bitcoincore-zmq = { version = "1.5.2", features = ["async"] }
+futures = "0.3.31"
+log = "0.4.25"
```
Rajil1213, line 22: Use the [`tracing`](https://lib.rs/crates/tracing) crate instead. It is already part of this repo (and the one being used in our other repos and also the [recommended](https://blessed.rs/crates) logging crate). You will need to initialize the logger before using it though: `strata_common::logging::init(strata_common::logging::LoggerConfig("test".to_string()))` and then you can set the `RUST_LOG` environment variable to set any log level.

Rajil1213 (changes requested):
Some nits that can mostly be done as part of the next phase of this PR. The API looks good.

crates/btc-notify/Cargo.toml
storopoli: I probably pointed you towards the `log` crate, my bad.
Yes @Rajil1213 is right.

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
storopoli: I also vouch for `thiserror`. We have good organizational knowledge and experience using it.

crates/btc-notify/src/lib.rs
ProofOfKeags: If I don't then my local run of `cargo llvm-cov` complains. That said I didn't try to do it with nextest in the early attempts so maybe it's redundant.

ProofOfKeags: Yeah it doesn't work because

```
error[E0658]: the `#[coverage]` attribute is an experimental feature
   --> crates/btc-notify/src/lib.rs:207:30
    |
207 | #[cfg_attr(coverage_nightly, coverage(off))]
    |                              ^^^^^^^^^^^^^
    |
    = note: see issue #84605 <https://github.com/rust-lang/rust/issues/84605> for more information
    = help: add `#![feature(coverage_attribute)]` to the crate attributes to enable
    = note: this compiler was built on 2024-10-31; consider upgrading it if it is out of date
```

The semantics of this macro are essentially as follows: "if coverage_nightly is set, then enable the 'coverage' attribute". Nextest sets `coverage_nightly` but doesn't enable the coverage attribute automatically.

crates/btc-notify/src/lib.rs
ProofOfKeags: FWIW this is a configurable parameter already, it's just that we don't have a constant that marks a different default. I've added a TODO anyway, but this shouldn't require any changes to be able to adapt to those networks as is.

crates/btc-notify/src/lib.rs
Rajil1213: Ahh I see. TIL. Could you add a comment there as well?

crates/btc-notify/src/lib.rs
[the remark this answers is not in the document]
ProofOfKeags: Looks like `thiserror` isn't a type but just exports derive macros for std::error::Error. Since `btc-notify` doesn't generate any original errors right now, I'm not sure how to incorporate this. All the errors that we return are propagated from dependency calls which will have to be injected into a unified error type or boxed as we do right now.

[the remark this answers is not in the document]
ProofOfKeags: Resolved out-of-band. This will be delayed until we have another error type that needs to be unified.

Rajil1213 (changes requested):
The API looks good now. Only need to organize the code a bit (as per the PR description).

crates/btc-notify/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+] }
+rust.unreachable_pub = "warn"
+rust.unused_crate_dependencies = "deny"
+rust.unused_must_use = "deny"
+rustdoc.all = "warn"
+
+[dependencies]
+bitcoin.workspace = true
+bitcoincore-zmq = { version = "1.5.2", features = ["async"] }
+futures = "0.3.31"
```
storopoli, line 21: This might be worth to move to the root-level `Cargo.toml`. I bet that other crates (especially the `bin/` ones) might use it.

crates/btc-notify/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+tokio = { workspace = true, features = ["macros", "rt", "sync", "test-util"] }
+tracing = "0.1.41"
```
storopoli, line 23: This might be worth to move to the root-level `Cargo.toml`. I bet that other crates (especially the `bin/` ones) might use it.

crates/btc-notify/src/client.rs
```diff
@@ -0,0 +1,692 @@
...
+}
+
+impl Drop for BtcZmqClient {
+    fn drop(&mut self) {
+        self.thread_handle.abort();
+    }
+}
+
+impl BtcZmqClient {
+    /// connect is the main constructor for BtcZmqClient.
```
storopoli, line 59:
```suggestion
    /// Primary constructor for [`BtcZmqClient`].
```

crates/btc-notify/src/client.rs
```diff
@@ -0,0 +1,692 @@
...
+    ///
+    /// It takes a [`BtcZmqConfig`] and uses that information to connect to bitcoind.
```
storopoli, line 61:
```suggestion
    /// It takes a [`BtcZmqConfig`] and uses that information to connect to `bitcoind`.
```

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub(crate) hashtx_connection_string: Option<String>,
+
+    /// connection string used in bitcoin.conf => zmqpubrawblock
+    pub(crate) rawblock_connection_string: Option<String>,
+
+    /// connection string used in bitcoin.conf => zmqpubrawtx
+    pub(crate) rawtx_connection_string: Option<String>,
+
+    /// connection string used in bitcoin.conf => zmqpubsequence
+    pub(crate) sequence_connection_string: Option<String>,
```
storopoli, line 31:
```suggestion
    /// Depth at which a transaction is considered buried, defaults to [`DEFAULT_BURY_DEPTH`].
    pub(crate) bury_depth: usize,

    /// Connection string used in `bitcoin.conf => zmqpubhashblock`.
    pub(crate) hashblock_connection_string: Option<String>,

    /// Connection string used in `bitcoin.conf => zmqpubhashtx`.
    pub(crate) hashtx_connection_string: Option<String>,

    /// Connection string used in `bitcoin.conf => zmqpubrawblock`.
    pub(crate) rawblock_connection_string: Option<String>,

    /// Connection string used in `bitcoin.conf => zmqpubrawtx`.
    pub(crate) rawtx_connection_string: Option<String>,

    /// Connection string used in `bitcoin.conf => zmqpubsequence`.
    pub(crate) sequence_connection_string: Option<String>,
```

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl BtcZmqConfig {
+    /// Updates the BtcZmqConfig with a zmqpubhashblock connection string and returns the updated
```
storopoli, line 35:
```suggestion
    /// Updates the [`BtcZmqConfig`] with a `zmqpubhashblock` connection string and returns the updated
```

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// config.
+    ///
+    /// Useful for a builder pattern with dotchaining.
+    pub fn with_hashblock_connection_string(mut self, s: &str) -> Self {
+        self.hashblock_connection_string = Some(s.to_string());
+        self
+    }
+
+    /// Updates the BtcZmqConfig with a zmqpubhashtx connection string and returns the updated
```
storopoli, line 44: same as the above

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// config.
+    ///
+    /// Useful for a builder pattern with dotchaining.
+    pub fn with_hashtx_connection_string(mut self, s: &str) -> Self {
+        self.hashtx_connection_string = Some(s.to_string());
+        self
+    }
+
+    /// Updates the BtcZmqConfig with a zmqpubrawblock connection string and returns the updated
```
storopoli, line 53: same as the above

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub fn with_rawblock_connection_string(mut self, s: &str) -> Self {
+        self.rawblock_connection_string = Some(s.to_string());
+        self
+    }
+
+    /// Updates the BtcZmqConfig with a zmqpubrawtx connection string and returns the updated
+    /// config.
+    ///
+    /// Useful for a builder pattern with dotchaining.
+    pub fn with_rawtx_connection_string(mut self, s: &str) -> Self {
```
storopoli, line 66: same as the above

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.rawtx_connection_string = Some(s.to_string());
+        self
+    }
+
+    /// Updates the BtcZmqConfig with a zmqpubsequence connection string and returns the updated
```
storopoli, line 71: same as the above

crates/btc-notify/src/config.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// config.
+    ///
+    /// Useful for a builder pattern with dotchaining.
+    pub fn with_sequence_connection_string(mut self, s: &str) -> Self {
+        self.sequence_connection_string = Some(s.to_string());
+        self
+    }
+
+    /// Updates the BtcZmqConfig with a new bury depth and returns the updated config.
```
storopoli, line 80: same as the above

crates/btc-notify/src/constants.rs
```diff
@@ -0,0 +1,5 @@
+//! This module provides the constant values used throughout the crate.
+
+// TODO(proofofkeags), TODO(Rajil1213): Use different default finality depths depending on the
+// network we are on?
+pub(crate) const DEFAULT_BURY_DEPTH: usize = 6;
```
storopoli, line 5: add a dosctring. This is an important `const`.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+//! # btc-notify
+//!
+//! btc-notify is a crate to deliver real-time notifications on the latest transaction and block
```
storopoli, line 3:
```suggestion
//! # `btc-notify`
//!
//! `btc-notify` is a crate to deliver real-time notifications on the latest transaction and block
```

Maybe also add that it uses exclusively the ZMQ stuff?

crates/btc-notify/src/subscription.rs
```diff
@@ -0,0 +1,28 @@
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use tokio::sync::mpsc;
+/// Subscription serves as the primary type that consumers of this API will handle. It is created
```
storopoli, line 7:
```suggestion
use tokio::sync::mpsc;

/// Subscription serves as the primary type that consumers of this API will handle. It is created
```

storopoli (changes requested):
minor nits. Some docstring nits can be also applied to other parts of the codebase that I've found that commenting would be a LOT of effort and you might already got the idea.

crates/btc-notify/Cargo.toml
Rajil1213: This is already there in the workspace toml.

Rajil1213 (commented):
Looks good. Let's try to get this over the finish line by today after the comments by Jose are resolved.

crates/btc-notify/src/lib.rs
ProofOfKeags: I will need to add some RPC stuff for backfilling purposes in a future revision so that won't be accurate.

crates/btc-notify/src/lib.rs
```diff
@@ -0,0 +1,1233 @@
...
+    /// A bury depth of 0 would mean that the block is considered buried the moment it is mined.
+    pub fn with_bury_depth(mut self, n: usize) -> Self {
+        self.bury_depth = n;
+        self
+    }
+}
+
+/// TxStatus is the primary output of this API via the subscription.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TxStatus {
```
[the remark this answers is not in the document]
ProofOfKeags: Revisiting this, the transaction id is computable from the raw transaction data which is always included with the event itself.

Rajil1213 (approved):

storopoli (approved):
ACK 52e7b5ce2c52ef37842a47f226be4d2fcc594bbc

PR #23, opened 2025-01-21
Proof statements

prajwolrg:
## Description

This PR creates a new crate for the proof statements. Once the statements are reviewed, this can be updated to the existing crate. Further benchmarking can be done once the statements are complete and sound.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [ ] I have added tests that prove my changes are effective or that my feature works.

## Related Issues

STR-943


Rajil1213 (changes requested):
As the anchor mechanism is out of scope for Testnet I, we don't need to verify the group signing keys/anchors in the proof. However, we might need it down the line. We could just move all the anchor-related statements to a different function and mark it with `#[expect(unused)]` to satisfy `clippy`.

Could you also add the following section to the `Cargo.toml` file?

```toml
[lints]
rust.missing_debug_implementations = "warn"
rust.missing_docs = "warn"
rust.unreachable_pub = "warn"
rust.unused_must_use = "deny"
rustdoc.all = "warn"
```

These will make sure that we don't have any extra dependencies, all pub items have documentation and all error cases are properly handled.

prajwolrg:
Removed the group signing keys/anchors in the proof altogether, since it was part of a lot of things, and it could not be cleanly separated. Have created [this branch](https://github.com/alpenlabs/strata-bridge/tree/prajwol/bridge-proof-with-anchor-proofs) with the anchor statements included (in case we need it later).

prajwolrg:
> As the anchor mechanism is out of scope for Testnet I, we don't need to verify the group signing keys/anchors in the proof. However, we might need it down the line. We could just move all the anchor-related statements to a different function and mark it with `#[expect(unused)]` to satisfy `clippy`.

Removed the group signing keys/anchors in the proof altogether, since it was part of a lot of things, and it could not be cleanly separated. Have created [this branch](https://github.com/alpenlabs/strata-bridge/tree/prajwol/bridge-proof-with-anchor-proofs) with the anchor statements included (in case we need it later).

crates/bridge-proof/protocol2/src/tx_info.rs
```diff
@@ -0,0 +1,69 @@
...
+// TODO: maybe read this from params or somewhere.
+pub(crate) const ROLLUP_NAME: &str = "alpenstrata";
+
+pub(crate) fn extract_checkpoint(
+    tx: &Transaction,
+    cred_rule: &CredRule,
+) -> Result<BatchCheckpoint, BridgeProofError> {
+    for inp in &tx.input {
+        if let Some(scr) = inp.witness.tapscript() {
+            if let Ok(payload) = parse_envelope_data(&scr.into(), ROLLUP_NAME) {
```
Rajil1213, line 23: `ROLLUP_NAME` should be a param.

Rajil1213 (changes requested):
Some nits and a potential issue with how the assignee assertion works.
I've updated the spec on the Withdrawal Fulfillment to reflect this but please update the proof statements there as well.

crates/bridge-proof/protocol2/src/tx_info.rs
```diff
@@ -0,0 +1,69 @@
...
+pub(crate) fn extract_withdrawal_info(
+    tx: &Transaction,
+) -> Result<(OperatorIdx, XOnlyPk, BitcoinAmount), BridgeProofError> {
+    let operator_id = u32::from_be_bytes(
+        tx.output[1].script_pubkey.as_bytes()[2..6]
+            .try_into()
+            .map_err(|_| {
+                BridgeProofError::TxInfoExtractionError(BridgeRelatedTx::WithdrawalFulfillment)
+            })?,
+    );
```
Rajil1213, line 54: This logic needs to be updated to verify since the withdrawal fulfillment transaction holds the signed operator index. Specifically, the `operator_idx` will be an input and we need to verify its signature against their corresponding public key in the chain state.

Rajil1213 (changes requested):
Forgot to add the comment on the withdrawal fulfillment info extraction.

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    // The deposit's assigned operator, destination address, and amount must match
+    // what was provided in the withdrawal fulfillment transaction.
+    let withdrawal = dispatched_state.cmd().withdraw_outputs().first().unwrap();
+    if operator_idx != dispatched_state.assignee()
+        || address != *withdrawal.dest_addr()
+        || amount != entry.amt()
+    {
+        // TODO: amount might be equal to entry.amt()
+        // TODO: verify if this might instead be equal to withdrawal.amt()
```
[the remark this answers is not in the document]
prajwolrg: Added a TODO for now.

crates/bridge-proof/protocol2/src/tx_info.rs
prajwolrg: Added a TODO for now.

crates/bridge-proof/protocol2/src/tx_info.rs
prajwolrg: Thought about this for a bit. I think we can get away with just including the `operator_idx` in the withdrawal fulfillment transaction. 
The signature instead will be part of the private input in the proof. This signature will be verified against the corresponding public key in the chain state. 

If there is a claim by the operator who actually did the withdrawal fulfillment transaction, they can supply the signature as a private input.
If there is a claim by other operator who did not do the withdrawal fulfillment, they won't be able to supply the valid signature.

The msg to sign can be the withdrawal fulfillment transaction itself, which includes the operator_idx.

Rajil1213: Nice! This looks reasonable.

Less work for me then. 😅

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -130,6 +131,20 @@ pub(crate) fn process_bridge_proof(
         return Err(BridgeProofError::InvalidWithdrawalData);
     }
 
+    // Extract the public key of the operator who did the withdrawal fulfillment
+    let operator_pub_key = input
+        .chain_state
+        .operator_table()
+        .get_operator(operator_idx) // TODO: optimization, maybe use `entry_at_pos` to avoid searching
```
Rajil1213, line 138: In practice, the number of operators is not gonna be too big. So we can defer optimizing this to when this actually becomes a problem.

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -130,6 +131,20 @@ pub(crate) fn process_bridge_proof(
...
+        .unwrap()
+        .signing_pk();
+
+    // NOTE: verifying the signing the withdrawal fulfillment transaction is sufficient
```
Rajil1213, line 142:
```suggestion
    // NOTE: verifying the signature of the withdrawal fulfillment transaction's txid is sufficient
```

Rajil1213 (changes requested):
Some nits.

Please don't forget to update the docs too.

crates/bridge-proof/protocol2/src/tx_info.rs
prajwolrg: Yup, one thing I'm not fully sure here is that the withdrawal transaction is signed by the operator once when the transaction is created itself for the bitcoin network. So the same signature might be used.

But this should be very easy to mitigate. We can sign the (withdrawal fullfillment transaction + some other data).

Rajil1213: When signing bitcoin transactions, it is _not_ the txid that is signed. So it should be fine, I think.

crates/primitives/Cargo.toml
```diff
@@ -14,12 +14,16 @@ rust.unused_must_use = "deny"
...
 lazy_static.workspace = true
 musig2 = { workspace = true, features = ["serde"] }
 rkyv.workspace = true
 secp256k1 = { workspace = true, features = ["global-context", "rand-std"] }
 serde.workspace = true
 sha2.workspace = true
 thiserror.workspace = true
+
+[dev-dependencies]
+strata-test-utils = { git = "https://github.com/alpenlabs/strata" }
```
Rajil1213, line 29: Looks like this dependency brings in `rockbound` which takes a ton of time to compile. And we are only just using `ArbitraryGenerator`. Since this is just a dev dependency, can we just duplicate the `ArbitraryGenerator` and add that to the `strata-bridge-test-utils` crate instead?

prajwolrg: I'm also using `strata-test-utils` for mainnet bitcoin block.
Shall I add a full bitcoin block to this repo as well?

Rajil1213: Yes, there is already a `test-data` directory where you can add these kinds of things.

Rajil1213: PS: I can see that we are importing the entire test-data when we import the `strata-test-utils` (via `include_bytes!`). Shipping binary data embedded in code this way is generally a bad practice. It's probably fine here since this just for testing. But flagging this with @AaronFeickert just in case.

storopoli: Yes, `test-data` is the way to go with those nasty things that are hard to generate. Ideally we would generate them, but out of the scope now.

Rajil1213: Yes, the only problem I have with this specifically is that the test data is included in the compiled code. So any time anybody uses this crate, they get the test-data as well. If it were a runtime thing (via `std::fs::read()`), that's fine.

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -0,0 +1,184 @@
...
+        || address != *withdrawal.dest_addr()
+        // TODO: amount should be equal to entry.amt() - withdrawal_fee
+        // withdrawal_fee will be part of the params
+        || amount != entry.amt()
+    {
+        return Err(BridgeProofError::InvalidWithdrawalData);
+    }
+
+    // 4e. Ensure that the withdrawal was fulfilled before the deadline
+    let withdrawal_fulfillment_height =
```
Rajil1213, line 135: Commenting here for visibility:

"This calculation must be moved before the `header_vs` is mutated so that the math is easier to grok."

crates/primitives/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 #![feature(array_try_from_fn)]
 
 pub mod bitcoin;
+pub mod bitcoin_tx;
```
Rajil1213, line 10: The `BitcoinTx` can just go into the `bitcoin` module. There's already a `BitcoinAddress` type there which was created for mostly the same reason as the `BitcoinTx` was created.

crates/primitives/Cargo.toml
prajwolrg: Moved away from `strata-test-utils`

crates/bridge-proof/protocol2/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+    chain_state: Chainstate,
+
+    /// The [HeaderVerificationState] used to validate the chain of headers.  
+    /// The proof that this HeaderVerificationState is valid must be done extracted from the
+    /// `strata_checkpoint_tx`.
+    header_vs: HeaderVerificationState,
+
+    /// The index of the deposit within the [Chainstate] deposit table.  
+    /// Must match the corresponding information in the withdrawal fulfillment transaction.
+    deposit_idx: usize,
```
Rajil1213, line 48: This should be a `u32`.

crates/bridge-proof/protocol2/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+            withdrawal_fulfillment_tx: input.withdrawal_fulfillment_tx,
+            op_signature: input.op_signature,
+        }
+    }
+}
+
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub(crate) struct BridgeProofOutput {
+    deposit_txid: Buf32,
+    claim_ts: u32,
```
Rajil1213, line 95: I don't think `claim_ts` is required (and I don't see it used anywhere either). We only need there to be a fixed number of blocks after the claim which can be done through the height of the block where the claim tx is mined.

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -0,0 +1,184 @@
...
+    let claim_header = headers[*claim_tx_idx];
+    verify_tx_inclusion(claim_tx, BridgeRelatedTx::Claim, claim_header, true)?;
+
+    // 6c. Check that the claim's recorded withdrawal fulfillment TXID matches the actual TXID of
+    // the withdrawal fulfillment transaction.
+    if withdrawal_fullfillment_txid != compute_txid(withdrawal_fulfillment_tx.transaction()).into()
+    {
+        return Err(InvalidClaimInfo::InvalidWithdrawalCommitment.into());
+    }
+
```
Rajil1213, line 165: We might need one more statement to check that there are enough headers in the provided chain and that the order of transactions is valid: checkpoint -> withdrawal fulfillment -> claim. It's fine it's being done somewhere else that I missed.

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
```diff
@@ -0,0 +1,186 @@
...
+    tx: BitcoinTx,
+    /// The Merkle inclusion proof associated with the transaction’s [`Txid`](bitcoin::Txid) or
+    /// [`Wtxid`](bitcoin::Wtxid).
+    proof: L1TxInclusionProof<T>,
+}
+
+impl<T: TxIdComputable> L1TxWithIdProof<T> {
+    // Ignored for now. This is meant to be called from elsewhere to generate to the format to be
+    // used by the prover
+    #[allow(dead_code)]
```
Rajil1213, line 30:
```suggestion
    #[expect(dead_code)]
```

`expect` will produce an error if this is not in fact `dead_code`. This will help clean things up more easily when this gets used.

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
```diff
@@ -0,0 +1,186 @@
...
+    ///
+    /// This function checks whether the target transaction has witness data. If it does not,
+    /// a proof is built for its `txid` alone. Otherwise, a proof is built for its `wtxid`,
+    /// and the coinbase transaction is used as the “base” transaction with a `txid` proof.
+    ///
+    /// # Panics
+    /// Panics if `idx` is out of bounds for the `txs` array (e.g., `idx as usize >= txs.len()`).
+    // Ignored for now. This is meant to be called from elsewhere to generate to the format to be
+    // used by the prover
+    #[allow(dead_code)]
```
Rajil1213, line 86:
```suggestion
    #[expect(dead_code)]
```

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
```diff
@@ -0,0 +1,186 @@
...
+    #[test]
+    fn test_segwit_tx() {
+        let blocks_bytes = include_bytes!("../../../../test-data/blocks.bin");
+        let blocks: Vec<Block> = bincode::deserialize(blocks_bytes).unwrap();
+        let block = &blocks[31];
+
+        let idx = 2;
+        let tx_bundle = L1TxWithProofBundle::generate(&block.txdata, idx);
+        assert!(tx_bundle.get_witness_tx().is_some());
+        assert!(tx_bundle.verify(block.header));
```
Rajil1213, line 184: To make this test more future-proof, it might be a good idea to not hardcode these block and transaction indices and instead just check the first 100 blocks or something. I don't think it's going to take that long but let me know if it takes much longer and you can ignore this suggestion.

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
```diff
@@ -0,0 +1,186 @@
...
+
+#[cfg(test)]
+mod tests {
+    use bitcoin::Block;
+
+    use super::L1TxWithProofBundle;
+
+    #[test]
+    fn test_segwit_tx() {
+        let blocks_bytes = include_bytes!("../../../../test-data/blocks.bin");
```
Rajil1213, line 177:
```suggestion
        let blocks_bytes = std::fs::read("../../../../test-data/blocks.bin").unwrap();
```
There is no reason this data needs to be in the test binary.

Rajil1213 (changes requested):
Some nits and questions but otherwise, this looks almost done.

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
AaronFeickert: I was not aware of this! TIL.

crates/bridge-proof/protocol2/src/statement.rs
prajwolrg: Nice catch on the ordering.
Regarding the enough number of headers, right now the numbers of headers after claim transaction is passed as the public output. I think it was required that this was the public output. Can't recall now.

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
Rajil1213: This is pretty new -- stabilized in [1.81.0](https://github.com/alpenlabs/strata-bridge/blob/ea2fa37b0982ba9753ac902b55e5b15e89d8faed/Makefile#L46-L54).

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
prajwolrg: Sounds good. Just looking at the block with transactions other than coinbase.

crates/bridge-proof/protocol2/src/tx_inclusion_proof.rs
prajwolrg: This caused some error since this was being used in the test. Updated it to handle that as well.

crates/bridge-proof/protocol2/src/statement.rs
Rajil1213: It was required to increase the cost of mining a private fork because the chain has to be long enough but I don't recall why it needs to be a public output.

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -0,0 +1,212 @@
...
+use crate::{
+    error::{BridgeProofError, BridgeRelatedTx, ChainStateError, InvalidClaimInfo},
+    tx_inclusion_proof::L1TxWithProofBundle,
+    tx_info::{extract_checkpoint, extract_claim_info, extract_withdrawal_info},
+    BridgeProofInputBorsh, BridgeProofOutput,
+};
+
+/// The number of headers after claim transaction that must be provided as private input
+const REQUIRED_NUM_OF_HEADERS_AFTER_CLAIM_TX: usize = 30;
```
Rajil1213, line 15: This number might not actually be 30. But works for now.

crates/bridge-proof/protocol2/src/statement.rs
```diff
@@ -0,0 +1,212 @@
...
+            REQUIRED_NUM_OF_HEADERS_AFTER_CLAIM_TX,
+            headers_after_claim_tx,
+        ));
+    }
+
+    // 8. Construct the proof output.
+    let output = BridgeProofOutput {
+        deposit_txid: entry.output().outpoint().txid.into(),
+        withdrawal_txid: withdrawal_fullfillment_txid.into(),
+    };
```
Rajil1213, line 209: We also need to verify whether the batch checkpoint proof from strata is correct.

Rajil1213 (changes requested):
This requires one more proof statement that checks if the strata batch checkpoint proof is valid as well.

crates/bridge-proof/protocol2/src/statement.rs
prajwolrg: Right now this is verified outside of this function in `process_bridge_proof_outer`, since the logic is zkvm dependent.

Rajil1213: I see. All good then.

Rajil1213 (approved):

PR #27, opened 2025-01-28
STR-859: Stake Chain

storopoli:
## Description

Stake Chain implementation.

TODO (for myself mostly):

- [x] `StakeChain` struct.
- [x] `PreStakeTx` transaction.
- [x] `ConnectorP` with tests.
- [x] `ConnectorS` with tests.
- [x] `StakeTx`
- [x] `StakeChain` and `StakeInputs` decoupling.
- [x] `StakeInputs::to_stake_chain`
- [x] Test a flow that:
  1. creates a `StakeChain` from at least some `StakeInputs`
  1. spends it into the `StakeTx` 0-index.
  1. and advances the `StakeTx` 0- to 1-index.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Does not cover integration with the transaction graph.

## Checklist

- [x] I have added tests that prove my changes are effective or that my feature works.

## Related Issues

STR-859


crates/stake-chain/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+edition = "2021"
+name = "strata-bridge-stake-chain"
+version = "0.1.0"
+
+[lints]
+rust.missing_debug_implementations = "warn"
+rust.rust_2018_idioms = { level = "deny", priority = -1 }
+rust.unreachable_pub = "warn"
+rust.unused_crate_dependencies = "deny"
+rust.unused_must_use = "deny"
```
Rajil1213, line 11: I did not add the `docs` lints in the `tx-graph` crate because it was mostly carried over from the PoC. For this crate, since we're starting from scratch, we should add the docs lints:

```suggestion
rust.unused_must_use = "deny"
rust.missing_docs = "warn"
rustdoc.all = "warn"
```

crates/stake-chain/src/transactions/pre_stake.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        Self {
+            psbt: Psbt::from_unsigned_tx(transaction)
+                .expect("cannot fail since transaction will be always unsigned"),
+            amount: stake_amount,
+        }
+    }
+
+    /// Creates a new [`PreStakeTx`] transaction from a PSBT.
+    pub fn from_psbt(psbt: Psbt, amount: Amount) -> Self {
```
Rajil1213, line 49: This constructor is probably unnecessary since the fields are `pub`. Any consumer that has both the fields can just construct the struct directly.

crates/stake-chain/src/transactions/pre_stake.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// The transaction's outputs.
+    pub fn outputs(&self) -> Result<Vec<TxOut>, StakeChainError> {
+        Ok(self.psbt.clone().extract_tx()?.output)
+    }
+
+    /// The transaction's fee.
+    pub fn fee(&self) -> Result<Amount, StakeChainError> {
+        Ok(self.psbt.fee()?)
```
Rajil1213, line 65: I didn't know this field existed. Nice!

crates/stake-chain/src/transactions/pre_stake.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// Creates a new [`PreStakeTx`] transaction from a PSBT.
+    pub fn from_psbt(psbt: Psbt, amount: Amount) -> Self {
+        Self { psbt, amount }
+    }
+
+    /// The transaction's inputs.
+    pub fn inputs(&self) -> Result<Vec<TxIn>, StakeChainError> {
+        Ok(self.psbt.clone().extract_tx()?.input)
```
Rajil1213, line 55: Since we only need the input, instead of cloning the whole `psbt`, you could just clone the `input` if you used `unsigned_tx`.
```suggestion
        Ok(self.psbt.unsigned_tx.input.clone())
```
Probably don't need this to be a `Result` either. I'm also not sure `extract_tx` would work because it is supposed to return a signed transaction and also expects that on each input the `witness_utxo` field is set.

crates/stake-chain/src/transactions/pre_stake.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// The transaction's outputs.
+    pub fn outputs(&self) -> Result<Vec<TxOut>, StakeChainError> {
+        Ok(self.psbt.clone().extract_tx()?.output)
```
Rajil1213, line 60: Same comment as in the `inputs` method about just cloning the `output`.

crates/stake-chain/src/transactions/pre_stake.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// The transaction's fee rate.
+    ///
+    /// # Note
+    ///
+    /// The fee rate calculation relies on an unchecked division using the total fees and the total
+    /// transaction virtual size. Internally it calls [`FeeRate::from_sat_per_vb_unchecked`].
+    pub fn fee_rate(&self) -> Result<FeeRate, StakeChainError> {
+        let vsize = self.psbt.clone().extract_tx()?.vsize();
```
Rajil1213, line 75: If this method is only supposed to be called on a signed transaction, then this is fine. Otherwise, we can just do this:
```suggestion
        let vsize = self.psbt.unsigned_tx.vsize();
```

Rajil1213 (changes requested):
Good start! Added a few nits (some of which would have surfaced in unit tests too).

crates/stake-chain/src/transactions/pre_stake.rs
storopoli: Done in 3a063c503874e129fe42e640f0aac7b91edfc7f7

crates/stake-chain/src/transactions/pre_stake.rs
storopoli: 3a063c503874e129fe42e640f0aac7b91edfc7f7

crates/stake-chain/src/transactions/pre_stake.rs
storopoli: I totally missed that the `extract_tx` was supposed to be signed.

crates/stake-chain/src/transactions/pre_stake.rs
storopoli: Alright removing it in 3a063c503874e129fe42e640f0aac7b91edfc7f7

crates/stake-chain/src/transactions/pre_stake.rs
storopoli: 3a063c503874e129fe42e640f0aac7b91edfc7f7

crates/tx-graph/src/connectors/connector_p.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// The N-of-N aggregated public key for the operator set.
+    n_of_n_agg_pubkey: XOnlyPublicKey,
+
+    /// The corresponding pre-image for the `k`th stake.
+    ///
+    /// This is a 32-byte array, that should be generated securely from the operator's master
+    /// entropy along with the `k`th stake's index.
+    ///
+    /// A suggestion is to hash them together, i.e. SHA256(master entropy || stake_index), and take
+    /// the 32 bytes of the resulting hash.
```
AaronFeickert, line 49: This seems like a possible footgun. What happens when the operator has multiple stake chains?

AaronFeickert: Would also recommend a more purpose-built KDF, like BLAKE3 or HKDF.

crates/tx-graph/src/connectors/connector_s.rs
```diff
@@ -47,3 +54,410 @@ impl ConnectorS {
...
+    /// The operator's public key.
+    operator_pubkey: XOnlyPublicKey,
+
+    /// The corresponding pre-image for the `k`th stake.
+    ///
+    /// This is a 32-byte array, that should be generated securely from the operator's master
+    /// entropy along with the `k`th stake's index.
+    ///
+    /// A suggestion is to hash them together, i.e. SHA256(master entropy || stake_index), and take
+    /// the 32 bytes of the resulting hash.
```
AaronFeickert, line 106: See related comment.

storopoli: 324ec8d1dfc8e7ac80a10f93239ad9b084d13b4a, see if that solves it

crates/tx-graph/src/connectors/connector_p.rs
storopoli: 324ec8d1dfc8e7ac80a10f93239ad9b084d13b4a, see if that solves it

storopoli: > This seems like a possible footgun. What happens when the operator has multiple stake chains?

The connector was designed to be stateless, i.e. pure implementation, it does not hold state.
The state would be all managed by the `StakeChain`.

I had to implement the missing pieces in these connectors in the `tx-graph` crate.

Eventually we'll move all the connectors to a standalone crate that will be a dependency of both to `tx-graph` and the `stake-chain` crates.

Cc @Rajil1213

AaronFeickert: I get the intent to offload state management, but I think that even the updated wording is not strong enough. It is required that the preimage be securely derived and _never_ reused under any circumstances. I would simply say this and nothing more.

crates/tx-graph/src/connectors/connector_s.rs
AaronFeickert: See related comment!

crates/tx-graph/src/connectors/connector_p.rs
```diff
@@ -0,0 +1,339 @@
...
+    pub fn create_tx_input_key_spend_path(&self, signature: Signature, input: &mut Input) {
+        finalize_input(input, [signature.as_ref()]);
+    }
+
+    /// Finalizes a psbt input where this connector is used with the script spend path.
+    ///
+    /// If the psbt input is already in the final state, then this method overrides witness.
+    pub fn create_tx_input_script_spend_path(&self, input: &mut Input) {
+        finalize_input(input, [&self.stake_preimage]);
+    }
```
storopoli, line 159: @Rajil1213 can you validate these functions for me? I've followed your approach with other connectors but didn't have time to test the PSBT signing yet.

crates/tx-graph/src/connectors/connector_s.rs
```diff
@@ -47,3 +55,405 @@ impl ConnectorS {
...
+    /// signature.
+    pub fn create_tx_input_script_spend_path(&self, signature: Signature, input: &mut Input) {
+        finalize_input(
+            input,
+            vec![
+                &self.stake_preimage.to_vec(),
+                &signature.serialize().to_vec(),
+            ],
+        );
+    }
```
storopoli, line 233: @Rajil1213 can you validate these functions for me? I've followed your approach with other connectors but didn't have time to test the PSBT signing yet.

crates/tx-graph/src/connectors/connector_p.rs
Rajil1213: Agree with Aaron. The suggestion can be removed from the docstring as this crate should not have to care about that. A warning to use a secure preimage should suffice.

crates/tx-graph/src/connectors/connector_s.rs
Rajil1213: A more ergonomic way to handle this is to have a separate enum for the spend paths that encapsulates the required args. Then, you can have just one `finalize` or `create_tx_input` method that matches on the enum. Also, just like the signature, the `preimage` should also be an arg instead of in the connector struct. Is there a reason for this being in the struct?  The connectors must be written to only hold public data (pubkeys, hashes, etc.) instead of the witness. This allows other operators to recreate these transactions provided that the public params are available.

crates/stake-chain/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+//! # Strata Bridge Stake Chain
+//!
+//! In a BitVM-based brideg, each transaction graph requires a stake that gets burnt/slashed/spent
```
Rajil1213, line 3:
```suggestion
//! In a BitVM-based bridge, each transaction graph requires a stake that gets burnt/slashed/spent
```
Should have been caught by `codespell` 🤦

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,52 @@
...
+    ///
+    /// The funding should cover for all the dust amounts in the transaction graph as well as that
+    /// for the stake transaction itself.
+    pub funding: TxOut,
+
+    /// The staking amount.
+    pub amount: Amount,
+
+    /// The `ΔS` interval relative timelock to advance the stake chain.
+    delta: relative::LockTime,
```
Rajil1213, line 48: Why is this the only private field?

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// The stake index, i.e., the index of the stake transaction in the chain.
+    pub index: u32,
+
+    /// The funding UTXO.
+    ///
+    /// # Notes
+    ///
+    /// The funding should cover for all the dust amounts in the transaction graph as well as that
+    /// for the stake transaction itself.
+    pub funding: TxOut,
```
Rajil1213, line 42: The funding UTXO is an arg for the individual state transactions. There is some nuance involved here that is not mentioned in the spec since it assumes 0-BTC outputs. I'll try to spec this out a bit more.

crates/stake-chain/src/transactions/pre_stake.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    /// The stake amount to be locked up.
+    pub amount: Amount,
+}
+
+impl PreStakeTx {
+    /// Creates a new [`PreStakeTx`] transaction from inputs and outputs.
+    ///
+    /// The caller should be responsible for ensuring that the first output should cover for the the
+    /// stake amount and all the dust amounts that are required to cover the transaction graph.
```
Rajil1213, line 28: It is not possible to determine the length of the stake chain when the PreStake is first created. The stake amount must be the same for each Stake transaction otherwise it messes up the cryptoeconomic guarantees. In practice, each stake requires ~50K sats. This amount of "leakage" can have serious implications.

The solution is to have a separate input that funds the dust outputs in the transaction graph _per stake transaction_. This also means that we cannot recreate the stake chain from the PreStake alone. I'll try to sketch out an approach to handle this in the spec.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,147 @@
...
+/// The outputs should be ordered in the following way:
+///
+/// 1. A dust output, [`ConnectorK`] used as an input to the Claim transaction and it is used to
+///    bind the stake to the deposit.
+/// 2. A dust output, [`ConnectorP`] used as an input to the Burn Payouts transaction that makes
+///    sure that, if an operator publishes their next stake transaction before a previous payout has
+///    been received, they will lose the ability to receive the payout. So it is in the operator's
+///    best interest to not advance the stake chain before all previous valid payouts have been
+///    received.
+/// 3. The stake amount, [`ConnectorStake`], which is the first output minus the already taken into
```
Rajil1213, line 29: This comment will need to be updated after the dust-funding mechanism is updated.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,147 @@
...
+///    account dust outputs. This is used to move the stake from the previous [`StakeTx`]
+///    transaction to the current one.
+///
+/// # Implementation Details
+///
+/// The [`StakeTx`] transaction is implemented as a generic type that takes the number of the
+/// previous stake transaction as a generic parameter. This is used to ensure that the stake
+/// transactions are ordered in the correct order.
```
Rajil1213, line 37: How does the generic ensure this? I don't see the generic being used in the struct itself. I do see it used in the constructor but I don't see the advantage of using the generic over just an input arg.

crates/tx-graph/src/connectors/connector_p.rs
```diff
@@ -0,0 +1,339 @@
...
+    n_of_n_agg_pubkey: XOnlyPublicKey,
+
+    /// The corresponding pre-image for the `k`th stake.
+    ///
+    /// This is a 32-byte array, that should be generated securely from the operator's master
+    /// entropy along with the `k`th stake's index.
+    ///
+    /// A suggestion is to use purpose-built KDF, like BLAKE3 or HKDF. Care should be taken given
+    /// that there could be multiple stake chains with different `k`th stakes.
+    stake_preimage: [u8; 32],
```
Rajil1213, line 51: This should not be here. This is the witness data. The data required to construct a connector must be shareable between operators so that each operator can compute the transactions deterministically. This is important for validating transactions before operators offer up their signatures.

crates/tx-graph/src/connectors/connector_p.rs
Rajil1213: As per the previous comment, it is more ergonomic to have an enum for the various paths that a connector can take. This enum should also hold the required witness so that we can have a single function that accepts that enum and the caller can provided whichever path they want to spend the connector to take.

Rajil1213 (changes requested):
This requires some changes to make sure that the connector can be constructed without any private data. This private data must also be used to finalize the connector when spending it.

Another problem is with the approach used to fund the dust outputs in the transaction graph. These dust amounts must not be deducted from the actual stake as that messes up with the cryptoeconomics. I will spec out how this can be done in the spec or in the tx machinery impl doc.

crates/tx-graph/src/connectors/connector_s.rs
storopoli: e24f7398b1f43f1644a8ddb3a131ca46af492f59

crates/stake-chain/src/stake_chain.rs
storopoli: My mistake... 🤦🏻

crates/stake-chain/src/lib.rs
storopoli: Done in e92bf9e08ca66ef6decd1615ad132d3e324e27cc
Thanks!

crates/stake-chain/src/stake_chain.rs
storopoli: Done in 6b5588c7ab3eb989ff3c6b66b620aba4a652bf31

crates/tx-graph/src/connectors/connector_p.rs
storopoli: Gonna resolve for now, let me know if this is not appropriate.

crates/tx-graph/src/connectors/connector_s.rs
storopoli: Gonna resolve for now, let me know if this is not appropriate.

crates/stake-chain/src/transactions/stake.rs
storopoli: Removed in 30611684204271fbe58652073a00b07418474cd5

crates/tx-graph/src/connectors/connector_p.rs
storopoli: Totally right! Done in 86a395159e8acf0b2dade142f8f5196cd804e10d.

crates/tx-graph/src/connectors/connector_p.rs
storopoli: Done in 4bf6aceb2f77685b89ed37dab55440b5fa883282.

crates/tx-graph/src/connectors/connector_s.rs
storopoli: Done in 4bf6aceb2f77685b89ed37dab55440b5fa883282.

crates/stake-chain/src/transactions/pre_stake.rs
storopoli: Done in 26845c48849f89acc4ef0e66380b45e1e048b3f8

crates/stake-chain/src/stake_chain.rs
storopoli: Done in b5255e37e9a628d05c97b3e82a4fa03f27619ebd and abstracted away in a `pub const`.

crates/stake-chain/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+rust.missing_docs = "warn"
+rustdoc.all = "warn"
+
+# TODO: remove the commented out (dev-)dependencies before the PR is merged.
```
Rajil1213, line 15: Adding this comment so we don't forget this.

crates/stake-chain/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+//! Ideally, we would like a single stake to be reused for each deposit. If any operator makes a
+//! faulty claim, this stake is spent by a disprover and renders any future reimbursement payouts
+//! impossible. The purpose of a stake chain is to enable this by creating a series of transactions
+//! that "move" the stake. The tradeoff is that each operator can service any one withdrawal request
+//! per stake chain at any given time i.e., if two withdrawal requests are assigned to the same
+//! operator with one stake chain, they cannot process both those requests concurrently. However, in
+//! the same scenario, if each operator had two stake chains, they could, in fact, process two
+//! separate withdrawal requests concurrently.
+
+#![feature(duration_constructors)] // for constructing `Duration::from_days`
```
Rajil1213, line 22: Is this being used somewhere?

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// 4. `amount`: The staking amount.
+    /// 5. `delta`: The `ΔS` interval relative timelock to advance the stake chain.
+    /// 6. `max_slashing_transactions`: Maximum number of slashing transactions to be created.
+    /// 7. `operator_funds`: The input that needs to be added to cover all the dust outputs of the
+    ///    first [`StakeTx`] transaction. It's value should be equal to
+    ///    [`OPERATOR_FUNDS`](crate::transactions::constants::OPERATOR_FUNDS).
+    /// 8. `connector_k`: The [`ConnectorK`] for the first [`StakeTx`] in the stake chain.
+    /// 9. `connector_p`: The [`ConnectorP`] for the first [`StakeTx`] in the stake chain.
+    /// 10. `connector_s`: The [`ConnectorStake`] for the first [`StakeTx`] in the stake chain.
+    #[allow(clippy::too_many_arguments)]
```
Rajil1213, line 77:
```suggestion
    #[expect(clippy::too_many_arguments)]
```

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// Once a stake chain is created, it can be advanced by revealing a preimage to a locking
+    /// script that is also relative timelocked to a certain `ΔS` interval.
+    ///
+    /// # Arguments
+    ///
+    /// 1. `original_stake`: The prevout for the first stake transaction, i.e., the first output of
+    ///    the [`PreStakeTx`](crate::transactions::PreStakeTx).
+    /// 2. `current_stake`: The input for the current stake transaction, this is the same as
+    ///    `original_stake` but as a [`TxIn`] instead of [`TxOut`].
+    /// 3. `index`: The stake index, i.e., the index of the stake transaction in this stake chain.
```
Rajil1213, line 67:
```suggestion
    /// 3. `index`: The stake index, i.e., the index of the latest stake transaction in this stake chain.
```

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// 5. `delta`: The `ΔS` interval relative timelock to advance the stake chain.
+    /// 6. `max_slashing_transactions`: Maximum number of slashing transactions to be created.
+    /// 7. `operator_funds`: The input that needs to be added to cover all the dust outputs of the
+    ///    first [`StakeTx`] transaction. It's value should be equal to
+    ///    [`OPERATOR_FUNDS`](crate::transactions::constants::OPERATOR_FUNDS).
+    /// 8. `connector_k`: The [`ConnectorK`] for the first [`StakeTx`] in the stake chain.
+    /// 9. `connector_p`: The [`ConnectorP`] for the first [`StakeTx`] in the stake chain.
+    /// 10. `connector_s`: The [`ConnectorStake`] for the first [`StakeTx`] in the stake chain.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
Rajil1213, line 78: This constructor does not actually create a stake chain. It just adds a transaction to the "chain" at the given index. In fact, the transaction might not even be part of the chain. For example, it is possible to create a new `StakeChain` instance that has 5 transactions and add a stake transaction at index 8 (or any value really).

This constructor should ideally accept all funding inputs and all hashes and reconstruct the entire chain to stay true to its name.

crates/stake-chain/src/transactions/constants.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn operator_funds() {
+        assert_eq!(OPERATOR_FUNDS, Amount::from_sat(48_840));
+    }
+}
```
Rajil1213, line 31: This docstring needs to be updated. It might also be better to just add the test as a doctest.

crates/tx-graph/src/connectors/connector_a30.rs
```diff
@@ -12,9 +12,9 @@ use strata_bridge_primitives::{params::connectors::PAYOUT_TIMELOCK, scripts::pre
 /// transaction to slash the stake.
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectorA30 {
-    n_of_n_agg_pubkey: XOnlyPublicKey,
+    pub n_of_n_agg_pubkey: XOnlyPublicKey,
 
-    network: Network,
+    pub network: Network,
```
Rajil1213, line 17: There is a `new()` constructor. This was not made pub to avoid accidentally mutating these values.

crates/tx-graph/src/connectors/connector_a31.rs
```diff
@@ -23,9 +23,9 @@ use crate::partial_verification_scripts::PARTIAL_VERIFIER_SCRIPTS;
 /// Connector from the PostAssert transaction to the Disprove transaction.
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectorA31 {
-    network: Network,
+    pub network: Network,
 
-    wots_public_keys: wots::PublicKeys,
+    pub wots_public_keys: wots::PublicKeys,
```
Rajil1213, line 28: See other comment on the `new()` constructor.

crates/tx-graph/src/connectors/connector_c0.rs
```diff
@@ -12,8 +12,8 @@ use strata_bridge_primitives::{
 /// Connector from the claim transaction used in optimistic payouts or assertions.
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectorC0 {
-    n_of_n_agg_pubkey: XOnlyPublicKey,
-    network: Network,
+    pub n_of_n_agg_pubkey: XOnlyPublicKey,
+    pub network: Network,
 }
```
Rajil1213, line 17: See the other comment on the `new()` constructor.

crates/tx-graph/src/connectors/connector_c1.rs
```diff
@@ -5,8 +5,8 @@ use strata_bridge_primitives::scripts::prelude::*;
 /// Connector output from the Claim transaction that is used for slashing.
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectorC1 {
-    n_of_n_agg_pubkey: XOnlyPublicKey,
-    network: Network,
+    pub n_of_n_agg_pubkey: XOnlyPublicKey,
+    pub network: Network,
```
Rajil1213, line 9: See the other comment on the `new()` constructor.

crates/tx-graph/src/connectors/connector_cpfp.rs
```diff
@@ -13,10 +13,10 @@ use strata_bridge_primitives::scripts::taproot::finalize_input;
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectorCpfp {
     /// The bitcoin network for which to generate output addresses.
-    network: Network,
+    pub network: Network,
 
     /// The public key used to create the child transaction in CPFP.
-    public_key: XOnlyPublicKey,
+    pub public_key: XOnlyPublicKey,
 }
```
Rajil1213, line 20: See the other comment on the `new()` constructor.

crates/tx-graph/src/connectors/witness_data.rs
```diff
@@ -0,0 +1,26 @@
...
+pub enum WitnessData {
+    /// The witness data is a single (aggregated) Schnorr [`Signature`](schnorr::Signature).
+    Signature(schnorr::Signature),
+
+    /// The witness data is a 32-byte hash preimage.
+    Preimage([u8; 32]),
+
+    /// The witness data are a single (aggregated) Schnorr [`Signature`](schnorr::Signature) and a
+    /// 32-byte hash preimage.
+    SignaturePreimage {
```
Rajil1213, line 19: Could just call this `Both` but this is fine too.

crates/tx-graph/src/connectors/witness_data.rs
```diff
@@ -0,0 +1,26 @@
+//! Some connectors can be spent in different ways, depending on the witness data.
+//! This modules contains [`WitnessData`] which represents the different ways that a connector can
+//! be spent given the witness data.
+
+use secp256k1::schnorr;
+
+/// Ways that a connector can be spent given the witness data.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive] // We might want to add more variants in the future.
```
Rajil1213, line 9: We don't really gain much from this. If we add more variants, it _will_ introduce breaking changes in the graph. It is better to have it be exhaustive and treat any changes to the witness data as a breaking change instead of things to proceed with new witness data unbeknownst to the caller.

crates/tx-graph/src/connectors/connector_s.rs
```diff
@@ -1,19 +1,29 @@
...
 #[derive(Debug, Clone, Copy)]
 pub struct ConnectorS {
-    /// The $ N $-of-$ N $ aggregated public key for the operator set.
-    n_of_n_agg_pubkey: XOnlyPublicKey,
+    /// The N-of-N aggregated public key for the operator set.
+    pub n_of_n_agg_pubkey: XOnlyPublicKey,
 
     /// The bitcoin network on which the connector operates.
-    network: Network,
+    pub network: Network,
```
Rajil1213, line 22: There is already a `new()` constructor that does the initialization. So these don't need to be pub.

Rajil1213 (changes requested):
The StakeChain API does not need a bit of a rework. I also realized that it is not sufficient to just have a stake index and a funding transaction + offset. We also need all hashes to be shared so that the entire stake chain can be reconstructed. Not sure if there is a way to reduce the bandwidth usage for sharing these hashes the same way we can reduce the bandwidth usage when sharing the funding inputs.

crates/tx-graph/src/connectors/connector_s.rs
storopoli: Removed the commit that added all this `pub` fields.

crates/tx-graph/src/connectors/witness_data.rs
storopoli: Removed the commit that added all this `pub` fields.

crates/tx-graph/src/connectors/connector_cpfp.rs
storopoli: Removed the commit that added all this `pub` fields.

crates/tx-graph/src/connectors/connector_c1.rs
storopoli: Removed the commit that added all this `pub` fields.

crates/tx-graph/src/connectors/connector_c0.rs
storopoli: Removed the commit that added all this `pub` fields.

crates/tx-graph/src/connectors/connector_a30.rs
storopoli: Removed the commit that added all this `pub` fields.

crates/stake-chain/src/transactions/constants.rs
storopoli: Done in 8272530f9d0110727646e44f4bcde861d8e84f85
great suggestion

crates/stake-chain/src/stake_chain.rs
storopoli: Done in 8272530f9d0110727646e44f4bcde861d8e84f85D

crates/stake-chain/src/stake_chain.rs
storopoli: Done in fc9aab1251b5b5f9f85c9d4db7aae23f69e68f4d

crates/stake-chain/src/lib.rs
storopoli: Removed in 623e8b8b57bf51a45cee14709233f802240b932f

crates/stake-chain/src/transactions/constants.rs
storopoli: Done in 8272530f9d0110727646e44f4bcde861d8e84f85
great suggestion

crates/stake-chain/src/lib.rs
storopoli: Removed in 623e8b8b57bf51a45cee14709233f802240b932f

crates/stake-chain/src/stake_chain.rs
storopoli: Ok done the API scaffolding here: 9919956d478ff89005da7f33fc0aad8fc134bc44

crates/stake-chain/Cargo.toml
storopoli: This is done as well. I need to add the dev-deps once I make the tests in a future commit

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,168 @@
...
+/// 1. The [`OPERATOR_FUNDS`] input that will cover all the dust outputs for the current stake
+///    transaction.
+/// 2. The stake amount from the previous [`StakeTx`] transaction.
+///
+/// # Output order
+///
+/// The outputs must be ordered in the following way:
+///
+/// 1. A dust output, [`ConnectorK`] used as an input to the Claim transaction and it is used to
+///    bind the stake to the deposit.
```
Rajil1213, line 32:
```suggestion
///    bind the stake to the transaction graph for a particular deposit.
```

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,168 @@
...
+/// 2. A dust output, [`ConnectorP`] used as an input to the Burn Payouts transaction that makes
+///    sure that, if an operator publishes their next stake transaction before a previous payout has
+///    been received, they will lose the ability to receive the payout. So it is in the operator's
+///    best interest to not advance the stake chain before all previous valid payouts have been
+///    received.
```
Rajil1213, line 37: If the stake is advanced before a withdrawal has been fully processed, then we burn the stake via the `Slash Stake` transactions. The purpose of the burn payouts is to burn the payout path if an operator starts publishing past claims (that weren't assigned) _after_ their stake has been slashed.

crates/stake-chain/src/transactions/stake.rs
Rajil1213: The comment needs updating still. Nothing is `minus`-ed from the stake.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,168 @@
...
+                value: stake_amount,
+                script_pubkey: connector_s.generate_address().script_pubkey(),
+            },
+            TxOut {
+                value: DUST_AMOUNT,
+                script_pubkey: operator_address.script_pubkey(),
+            },
+        ];
+        let transaction = Transaction {
+            version: transaction::Version(2),
```
Rajil1213, line 113: This should be version 3.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,168 @@
...
+            lock_time: absolute::LockTime::ZERO,
+            input: inputs,
+            output: outputs,
+        };
+
+        // The `ConnectorS` is the third output.
+        let stake_amount = transaction.output[STAKE_VOUT as usize].value;
```
Rajil1213, line 120: What is the purpose of this? The `stake_amount` is already passed in as an argument to this function.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        Self {
+            index,
+            psbt: Psbt::from_unsigned_tx(transaction)
```
Rajil1213, line 124: It is not enough to just construct the psbt. In each of the inputs, the `witness_utxo` field must also be set which contains the `prevout` for this transaction. Without this, the call to `extract_tx` will fail (and I think the fee calculation won't work either).

Rajil1213 (changes requested):
Only reviewed the new changes to this PR (namely, the chain and the stake tx APIs). This  still needs a bit of work but it looks really good. Nice work!

Rajil1213:
A problem with the approach where the stake chain is pre-setup is that in the current design, all of the WOTS public keys need to be shared. Each instance of the `WotsPublicKeys` has a size of 550 Kb. So if we presetup 10,000 stake transactions for 10,000 deposits, we will be sharing 5 Gb of data per operator.

The solution is to just pass in a single `Wots256PublicKey` required for the `ConnectorK` and share that. It has a size of 256 *20 bytes = 5120 bytes.  So for 10,000 stake transactions, it'll be around 50 Mb. Here is the relevant code:

* The ConnectorK constructor: 
https://github.com/alpenlabs/strata-bridge/blob/c27ffc6e14f057dfb1d3b1fcb436a7976c666f7a/crates/tx-graph/src/connectors/connector_k.rs#L11-L39

* The usage:
https://github.com/alpenlabs/strata-bridge/blob/c27ffc6e14f057dfb1d3b1fcb436a7976c666f7a/crates/tx-graph/src/connectors/connector_k.rs#L40-L45

crates/stake-chain/src/transactions/stake.rs
storopoli: Done in 8eebc9f389541fd486a06e14fcc172436c80fdd7

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /// Relative timelock interval to advance the stake chain.
+    ///
+    /// The stake chain can be advanced forward by revealing a preimage to a locking script that is
+    /// also relative timelocked to a certain `ΔS` interval.
+    pub fn delta(&self) -> relative::LockTime {
+        self.delta
+    }
+}
```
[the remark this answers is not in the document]
storopoli: Yes, it is helping me tons with the testing and image when integrating this whole thing; and also using.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// 3. N-of-N aggregated bridge public key.
+    /// 4. `N`-length array of WOTS public keys.
+    /// 5. `N`-length array of stake hashes.
+    /// 6. `N`-length array of operator fund prevouts.
+    /// 7. Original stake prevout.
+    /// 8. `ΔS` relative timelock interval.
+    /// 9. Network.
+    ///
+    /// For an explanation of the parameters, see the documentation for [`StakeInputs`].
+    #[expect(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
Rajil1213: Sure. But this is a pretty minor change.

crates/stake-chain/src/transactions/stake.rs
storopoli: Nice catch removed in 090e775638d0ae3ef8ee3b3d31db8b406fa6ccd0

crates/stake-chain/src/transactions/stake.rs
storopoli: Done in 090e775638d0ae3ef8ee3b3d31db8b406fa6ccd0, thanks!

crates/stake-chain/src/transactions/stake.rs
storopoli: Done in 090e775638d0ae3ef8ee3b3d31db8b406fa6ccd0

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,322 @@
...
+///
+/// # Construction
+///
+/// [`StakeChain`]s can be constructed by first creating a [`StakeInputs`] of length `N` and then
+/// calling [`StakeInputs::<M>::to_stake_chain`](StakeInputs::to_stake_chain), where `M < N`
+/// (compile-time check).
+///
+/// The user can also coerce a [`Vec<StakeTx>`] into a `[StakeChain; N]`, but it does not offer the
+/// same compile-time guarantees as the previous method.
+pub type StakeChain<const N: usize> = [StakeTx; N];
```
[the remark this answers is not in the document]
storopoli: I like this done in 090e775638d0ae3ef8ee3b3d31db8b406fa6ccd0

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,322 @@
+//! The stake chain is a series of transactions that move the stake from a previous stake to a new
+//! stake.
```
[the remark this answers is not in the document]
storopoli: Done in 090e775638d0ae3ef8ee3b3d31db8b406fa6ccd0

crates/stake-chain/src/stake_chain.rs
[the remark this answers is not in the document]
storopoli: No, let me do it then since it is minor.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,610 @@
...
+
+        // Needed for 1P1C TRUC relay
+        let prev_outputs_1p1c = PreviousTransactionOutput {
+            txid: cpfp_txid,
+            vout: cpfp_vout as u32,
+            script_pubkey: cpfp_spk.to_string(),
+            redeem_script: None,
+            witness_script: None,
+            amount: Some(DUST_AMOUNT.to_btc()),
+        };
```
storopoli, line 594: My brain is fried for today...
Can you help me out with this workflow. I always get confused with these `submitpackage`s stuff.
The `StakeTx` does not have a fee, that is by design, because we want the graph to be "pure" and adding/tweaking fees will f*** everything.
Hence the last output, 4th one, is an operator controlled dust output for CPFP.

How is the workflow again to get this to be done by `bitcoind` and `rust-bitcoin`?

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,610 @@
...
+
+    /// Converts a [`StakeInputs`] into a [`StakeChain`].
+    ///
+    /// # Note
+    ///
+    /// The [`StakeChain`] can be of length less than or equal to the [`StakeInputs`].
+    ///
+    /// It is impossible to create a [`StakeChain`] with a length greater than the [`StakeInputs`].
+    /// This is done by compile-time checks.
+    pub fn to_stake_chain<const M: usize>(&self) -> StakeChain<M>
```
Rajil1213, line 188: I see that we're still keeping this around. I don't have an issue with this but it might be better to just move this impl to inside the `StakeChain::new()` method and remove this one.

crates/stake-chain/src/stake_chain.rs
Rajil1213: You can check out the test for the `PegOutGraph`. For example, here's how the `ClaimTx` is settled using CPFP:

https://github.com/alpenlabs/strata-bridge/blob/c27ffc6e14f057dfb1d3b1fcb436a7976c666f7a/crates/tx-graph/src/peg_out_graph.rs#L430-L463

You might need to implement the `CovenantTx` trait for the `StakeTx` (it should be simple enough since it only has accessors).

There are a couple of test-utils created to simplify CPFP-ing transactions in tests but they lack proper documentation. 🫣

Rajil1213: Alternatively, you could just start out by adding these to the bitcoind config:
```
minrelaytxfee=0.0
blockmintxfee=0.0
dustRelayFee=0.0
```

This will allow settling transactions with 0 fees. This is so that you can test the more interesting parts of the Stake Chain first and deal with the boring CPFP stuff later.

storopoli: Oh great!

storopoli: This is great! Man nice suggestion. Simplifies a lot of things!
Done in 911d24291467e32ec054ceb70cce9559971c95ed.
Ok `PreStakeTx -> StakeTx_0` done. I'll start to implement the `StakeTx_0 -> StakeTx_1` tests. But first will take a look on other pending stuff in this PR that you've pointed out.

Thanks!

crates/stake-chain/src/stake_chain.rs
storopoli: Done in bcc22ad1fa7eed52f85cc3b01ad7171debd76c64
Nice catch!

crates/stake-chain/src/stake_chain.rs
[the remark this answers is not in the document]
storopoli: Saved in a stash locally if needed.

crates/stake-chain/src/transactions/stake.rs
storopoli: Yes, `extract_tx` will fail but not because of this. It will fail because the transaction is not signed.

Rajil1213: Yes I believe that is the first check. But it will fail even if it _is_ signed because the `witness_utxo` field is `None`.

storopoli: I'll check I am trying to finish the tests and I want to implement a function that will sign the $StakeTk_{k}$ using the operator key and the preimage of $StakeTk_{k-1}$. This is what is missing now.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -161,4 +171,87 @@ impl StakeTx {
...
+        finalize_input(
+            self.psbt.inputs.get_mut(1).expect("must have second input"),
+            [
+                preimage.to_vec(),
+                signature.serialize().to_vec(),
+                locking_script.to_bytes(),
+                control_block.serialize(),
+            ],
+        );
+    }
```
storopoli, line 256: @Rajil1213, any insights on how to finalize these inputs?
The first input is the operator funds for the dust outputs and the second it the connector S.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,748 @@
...
+    /// The provided [`StakeInputs`] must be of length `N`.
+    pub fn new<const N: usize>(stake_inputs: &StakeInputs<N>) -> Self
+    where
+        [(); N - M]:,
+    {
+        let wots_sk = get_deposit_master_secret_key(
+            &stake_inputs.wots_master_sk,
+            stake_inputs.pre_stake_txid,
+        );
+        let wots_pubkey = wots::Wots256PublicKey::new(&wots_sk);
```
Rajil1213, line 48: The wots public keys are different for each of the stake transactions. This implementation would mean that each stake transaction would have the same locking script for committing to the `withdrawal_fulfillment_txid`.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,748 @@
...
+/// [`StakeChain`].
+///
+/// The original stake is the first stake transaction in the chain, which is used to stake in the
+/// transaction graph for a single deposit and is moved after a successful deposit, i.e., the
+/// operator is not succcesfully challenged and has it's stake slashed.
+/// It is the first output of the [`PreStakeTx`](crate::prelude::PreStakeTx).
+///
+/// The network is the bitcoin network on which the stake chain operates.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StakeInputs<const N: usize> {
```
Rajil1213, line 153: With the addition of the `wots_master_key`, it's worth splitting this struct out to only hold values that can be shared publicly. The staking amount, `n_of_n_agg_pubkey`, `network` and `delta` are parameters dictated by the protocol so don't need to be shared. The `operator_pubkey`, `stake_hashes`, `operator_funds` and `original_stake` are also public data but are only available to the operator that is creating the stake chain. These need to be shared. The `wots_master_key` is a private data that should not be shared at any cost. What should be shared instead is the `wots_public_key` generated with the secret key.

Restructuring the inputs this way will also help in the p2p integration because this stake data is not part of the current implementation in the `strata-p2p` crate and will need to be added (as part of the genesis messages).

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,748 @@
...
+
+    /// Operator fund prevouts to cover dust outputs for the entirety of the `N`-length
+    /// [`StakeChain`].
+    operator_funds: [TxIn; N],
+
+    /// Prevout for the first stake transaction.
+    original_stake: TxIn,
+
+    /// [`PreStakeTx`](crate::transactions::PreStakeTx)'s [`Txid`].
+    pre_stake_txid: Txid,
```
Rajil1213, line 178: Isn't `original_stake` just `pre_stake`? And if so, we can remove the `pre_stake_txid` because it can be computed.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,748 @@
...
+    async fn stake_chain_advancement() {
+        logging::init(LoggerConfig::new("stake_chain_advancement".to_string()));
+
+        // Setup Bitcoin node
+        let mut conf = Conf::default();
+        conf.args.push("-txindex=1");
+        // Let's not deal with CPFP 1P1C TRUC relay annoyances for this test
+        conf.args.push("-minrelaytxfee=0.0");
+        conf.args.push("-blockmintxfee=0.0");
+        conf.args.push("-dustrelayfee=0.0");
```
Rajil1213, line 382: Now that the actual flow is working, we should add CPFP back in and remove these args. It's fine to defer this though and I'll add it during the integration.

crates/stake-chain/src/transactions/stake.rs
Rajil1213: The only thing missing for us to be able to use `extract_tx` at least is to add the `witness_utxo` fields in the psbt inputs. This will also be necessary for fee calculation (for CPFP). This is also fine to defer to later.

crates/tx-graph/src/connectors/connector_k.rs
```diff
@@ -19,33 +19,30 @@ pub struct ConnectorK {
...
+    pub wots_public_key: wots::Wots256PublicKey,
 }
 
 impl ConnectorK {
     /// Constructs a new connector.
     pub fn new(
         n_of_n_agg_pubkey: XOnlyPublicKey,
         network: Network,
-        wots_public_keys: wots::PublicKeys,
+        wots_public_key: wots::Wots256PublicKey,
```
Rajil1213, line 31: This arg should be named `withdrawal_fulfillment_pk` because now we know exactly what the wots pubkey is for.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,748 @@
...
+
+        // Get network
+        let network = btc_client
+            .get_blockchain_info()
+            .expect("must get blockchain info")
+            .chain;
+        let network = network.parse::<Network>().expect("network must be valid");
+
+        // get the xpriv
+        let xpriv = async_client.get_xpriv().await.unwrap().unwrap();
```
Rajil1213, line 395: Can we not use the `btc_client` for this?

Rajil1213 (changes requested):
The API needs a bit of polishing still after the addition of the wots public keys for the Stake Tx.

Other than that, there are a few changes that we can defer to later.

crates/stake-chain/src/stake_chain.rs
storopoli: Not quite because the `original_stake` is not a `TxOut` or a "prevout" it is a `TxIn`, or a "previn".

crates/stake-chain/src/stake_chain.rs
storopoli: Honestly, I'd prefer to punt this 😞

crates/stake-chain/src/stake_chain.rs
storopoli: Sure, I also don't like to do this I'll refactor to use auto-generated `Keypair`s and `Address`es and not deferring to `bitcoind` to sign transactions.
This is the only thing that the `async` client is doing here to get the xpriv.

crates/stake-chain/src/stake_chain.rs
Rajil1213: Yes, but can the `pre_stake_txid` not be extracted from the `TxIn`?

crates/stake-chain/src/stake_chain.rs
Rajil1213: I think that part is fine since this is just for tests. The earlier comment was only regarding using `btc_client` instead of `async_client` since it looks like all that the `async_client` is doing is setting the `XPRIV_RETRIEVABLE` envvar.

crates/stake-chain/src/stake_chain.rs
storopoli: Yes! Damn, I started Monday with the wrong foot. Yes you're right! Will update in a commit after I fix the `async` client.

crates/stake-chain/src/stake_chain.rs
storopoli: Ok 6cb2a8400c020e2b9b225f0fa005c4b44b1682d9 removes the `btcio` and `tokio` deps and adds some helper funs which make the tests nicer.

crates/stake-chain/src/stake_chain.rs
storopoli: Done in d7e2939d5866e9c5a2c5868266c8247f856e5127

crates/tx-graph/src/connectors/connector_k.rs
storopoli: Done in 0bdf48e6fedda36799a4e9486e71b10f794c57e2

crates/stake-chain/src/stake_chain.rs
storopoli: Done in cfd85252463a6fd7c7e5010a5b6ad9e2b95cce65 for the protocol params.

crates/stake-chain/src/stake_chain.rs
storopoli: Ugh makes sense :)

crates/stake-chain/src/stake_chain.rs
storopoli: Done in fc93c57d97fe550330ba42cb085b74a6b59cc8eb for the WOTS PKs.

storopoli:
> The only thing missing for us to be able to use `extract_tx` at least is to add the `witness_utxo` fields in the `psbt` inputs. This will also be necessary for fee calculation (for CPFP). This is also fine to defer to later.

Done in f6d69a30eac2707343698d94b8f27c263c5f77ee

crates/stake-chain/src/params.rs
Rajil1213, line 1: This can be moved to `primitives/params` but it's fine for now.

crates/stake-chain/src/stake_chain.rs
```diff
@@ -0,0 +1,881 @@
...
+
+    /// Operator fund prevouts to cover dust outputs for the entirety of the `N`-length
+    /// [`StakeChain`].
+    operator_funds: [TxIn; N],
+
+    /// Operator funds previous UTXOs.
+    operator_funds_utxo: [TxOut; N],
+
+    /// Prevout for the first stake transaction.
+    original_stake: TxIn,
```
Rajil1213, line 178: Is this the first stake transaction or the `pre_stake`?

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,267 @@
...
+        index: u32,
+        stake_input: TxIn,
+        stake_amount: Amount,
+        operator_funds: TxIn,
+        operator_pubkey: XOnlyPublicKey,
+        previous_utxos: [TxOut; 2],
+        connector_k: ConnectorK,
+        connector_p: ConnectorP,
+        connector_s: ConnectorStake,
+        network: Network,
```
Rajil1213, line 97: Nit: the `previous_utxos` should be split out in the same way `stake_input` and `operator_funds` are split out. This is not just for consistency but also because the two `TxOut`'s in `previous_utxos` are not interchangeable. The documentation and the numbering does help but having the arguments be explicit is better.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,267 @@
...
+                    .minimal_non_dust(),
+                script_pubkey: connector_p.generate_address().script_pubkey(),
+            },
+            TxOut {
+                value: stake_amount,
+                script_pubkey: connector_s.generate_address().script_pubkey(),
+            },
+            TxOut {
+                value: DUST_AMOUNT,
+                script_pubkey: operator_address.script_pubkey(),
```
Rajil1213, line 125: Not sure how I didn't notice this the last time, but you should just use the `cpfp` connector here. It does the same thing as what's being done here but using that connector is a bit ergonomic as it serves as a documentation to the user that this output is for cpfp.

With that, you can remove the `network` and `operator_pubkey` args in this function.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// transaction virtual size. Internally it calls [`FeeRate::from_sat_per_vb_unchecked`].
+    pub fn fee_rate(&self) -> Result<FeeRate, StakeChainError> {
+        let vsize = self.psbt.unsigned_tx.vsize();
+        let fee = self.fee()?;
+        Ok(FeeRate::from_sat_per_vb_unchecked(
+            fee.to_sat() / vsize as u64,
+        ))
+    }
+
+    /// Adds the preimage and signature for the previous [`StakeTx`] transaction is an input to the
```
Rajil1213, line 187:
```suggestion
    /// Adds the preimage and signature for the previous [`StakeTx`] transaction as an input to the
```

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -0,0 +1,267 @@
...
+    ///
+    /// Under the hood, it spents the underlying [`ConnectorStake`] from the previous [`StakeTx`].
+    pub fn finalize_connector_s(
+        &mut self,
+        operator_funds: TxOut,
+        preimage: &[u8; 32],
+        keypair: &Keypair,
+        n_of_n_agg_pubkey: XOnlyPublicKey,
+        delta: relative::LockTime,
+        network: Network,
```
Rajil1213, line 202: This function should just take the preimage and the signature. Using the `keypair` means that the secret service needs to call this function which should not be the case. The secret service accepts the message hash (raw bytes) and return the signature. That signature can be used to finalize the stake connector. The preimage is returned by the secret service, so that part is fine.

crates/tx-graph/src/connectors/connector_k.rs
```diff
@@ -19,33 +19,30 @@ pub struct ConnectorK {
     /// The bitcoin network for the addresses generated by the connector.
     pub network: Network,
 
-    /// The WOTS public keys used for bitcommitment scripts.
-    pub wots_public_keys: wots::PublicKeys,
+    /// The WOTS public key used for bitcommitment scripts.
+    pub wots_public_key: wots::Wots256PublicKey,
```
Rajil1213, line 23: This field should be renamed as well.

Rajil1213 (changes requested):
Some final nits and polishing and I think this is good to go.

crates/stake-chain/src/params.rs
storopoli: Done in 29ff49cbedf2d98cc269102ca94abf17bf6e9034
Nice suggestion.

crates/stake-chain/src/stake_chain.rs
storopoli: Good catch corrected in 8ea81aa736d46a00485e260e13242ecd39bcde68

crates/tx-graph/src/connectors/connector_k.rs
storopoli: I don't get it. Why it should be renamed? It is the public key no, and `wots_public_key` reflects that.

crates/stake-chain/src/transactions/stake.rs
storopoli: Nice catch. Done in 20df86e2038f18c1f38368a0be1108a8f26eff44

crates/tx-graph/src/connectors/connector_k.rs
Rajil1213: The fact that it's the `wots_public_key` is reflected in the type. The struct is more specific, and it's better for it to reflect the purpose of this key.

Rajil1213 (approved):
Okay, I think we're all good now. Nice work!

PR #28, opened 2025-01-31
feat: secret service

Zk2u:
## Description

This PR implements a secret service system consisting of 3 library crates, server, proto and client, and 1 binary that acts as the server implementor. The client can be used by the other bridge binaries.

The "server" is responsible for keeping secrets used in strata bridge. The "clients" are the sequencer/operator binaries for their respective duties. They communicate using a QUIC connection with a binary-based rkyv wire protocol. Security is handled through mutually authenticated TLS 1.3 as part of QUIC. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Checklist

- [x] I have added tests that prove my changes are effective or that my feature works.

## Related Issues

STR-934

### DIFF
--- a/bin/secret-service/src/tls.rs
+++ b/bin/secret-service/src/tls.rs
@@ -34,7 +34,7 @@ pub(crate) async fn load_tls(conf: TlsConfig) -> ServerConfig {
         (vec![cert], key.into())
     } else {
         error!("TLS configuration is missing certificate and key paths");
-        std::process::exit(10);
+        std::process::exit(1);
     };
 
     let tls_builder = if let Some(ca_path) = conf.ca {
@@ -54,7 +54,7 @@ pub(crate) async fn load_tls(conf: TlsConfig) -> ServerConfig {
         ServerConfig::builder().with_no_client_auth()
     } else {
         error!("TLS configuration is missing CA certificate path");
-        std::process::exit(11);
+        std::process::exit(1);
     };
 
     tls_builder


Rajil1213 (changes requested):
At the moment, it feels like this is still very much in the scaffolding stage rather than in the concrete implementation. The only concrete implementation I see is in the client-server parts of the secret service rather than what it's supposed to be doing. 

I can also see that the types have been laid out such that the server also does the musig2 session book-keeping but the actual implementation for this book-keeping/signing/aggregating has not been written out yet. Another thing missing is where all this data is stored. If this service is going to handle the musig2 signing session as well, it will need this information to be persistent and recoverable.  

Another thing I don't see is a wallet-like functionality, specifically an API to retrieve unspent UTXOs.

I would recommend working on the concrete parts of the implementation:

- [ ] musig2 signing/aggregating
- [ ] wots signing/aggregating
- [ ] p2p signing
- [ ] pre-image generation
- [ ] wallet integration

...before moving on to the binary/client implementation.

The reason I bring this up is that the way it currently stands right now, at least some of the interfaces and designs might change and bitcoin-core being what it is, I can easily see these taking a full sprint which is far from ideal.

Rajil1213:
> Check the secret-service-server crate for the musig2 in-memory session management. The implementor (secret-service) is responsible for handling persistence.

Is the implementor also in scope for this PR?

> PS: whaddya mean by bitcoin core? None of the secret service touches or depends on core.

Right but when you write tests for all the trait implementors, you will need to eventually broadcast a signed transaction. And if debugging those can be a pain.

PS: I also see that you've brought the entire `musig2` crate. If this is just for serialization, you should just create newtype wrappers. But let me know if there is some other reason.

crates/musig2/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+name = "musig2"
+version = "0.1.0"
+edition = "2021"
+authors = ["conduition <conduition@proton.me>"]
+description = "Flexible Rust implementation of the MuSig2 multisignature protocol, compatible with Bitcoin."
+readme = "README.md"
+license = "Unlicense"
+repository = "https://github.com/conduition/musig2"
+keywords = ["musig", "schnorr", "bitcoin", "multisignature", "musig2"]
+include = ["/src", "!/src/test_vectors", "*.md"]
```
storopoli, line 11: What the hell? 😂
just `cargo add musig2`...

Cargo.toml
```diff
@@ -79,14 +85,18 @@ esplora-client = { git = "https://github.com/BitVM/rust-esplora-client", default
...
+futures = "0.3.31"
 hex = { version = "0.4", features = ["serde"] }
 jsonrpsee = "0.24.7"
 jsonrpsee-types = "0.24.7"
+kanal = "0.1.0-pre8"
 lazy_static = "1.5.0"
 musig2 = { version = "0.1.0", features = [
   "serde",
   "rand",
 ] } # can't be updated without updating bitcoin
```
storopoli, line 97: I mean it's here...

crates/musig2/Cargo.toml
storopoli: Oh I see because of the serialization for `rkyv`.
Just do the same to what we did on `strata` for the `musig2` stuff for Borsh.
See https://github.com/alpenlabs/strata/blob/a0d413d711bb3af5ce312544f45600806f3f18bb/crates/primitives/src/bridge.rs#L225-L263

crates/musig2/src/rkyv_wrappers.rs
```diff
@@ -0,0 +1,190 @@
+use rkyv::{Archive, Deserialize, Serialize};
+use secp256k1::ffi::CPtr;
```
storopoli, line 2: Just do the same to what we did on `strata` for the `musig2` stuff for Borsh.
See https://github.com/alpenlabs/strata/blob/a0d413d711bb3af5ce312544f45600806f3f18bb/crates/primitives/src/bridge.rs#L225-L263

crates/secret-service-client/src/lib.rs
```diff
@@ -0,0 +1,479 @@
+use std::{
+    future::Future,
+    io,
+    net::{Ipv4Addr, SocketAddr},
+    sync::Arc,
+    time::Duration,
+};
```
storopoli, line 7: This needs to not be a huge `lib.rs`.

crates/secret-service-server/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+
+use std::{
+    future::Future,
+    io,
+    marker::Sync,
+    net::SocketAddr,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
```
storopoli, line 12: Must not be a huge single `lib.rs`

crates/secret-service/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "secret-service"
+version = "0.1.0"
+edition = "2021"
```
storopoli, line 4: This should be moved to `bin/` directory maybe.
Cc @Rajil1213.

storopoli (changes requested):
The direction is good, but it needs ton of polishing:

1. Get rid of the `musig2` code and do tuple wrapper structs in the same fashion as how we did for `musig2` stuff for Borsh in `strata`, see Just do the same to what we did on strata for the musig2 stuff for Borsh.
See https://github.com/alpenlabs/strata/blob/a0d413d711bb3af5ce312544f45600806f3f18bb/crates/primitives/src/bridge.rs#L225-L263 for examples.
2. Needs to be better organized, the crates are huge `lib.rs` files and the binary should be moved to a `bin/` subdir somewhere.
3. Needs also @Rajil1213's clippy lints secret sauce config for the `Cargo.toml`s.

crates/musig2/Cargo.toml
Rajil1213: What is the exact type that necessitates forking `musig2` for the sake of serialization? It will be tedious but unless you also want to serialize private types from the `musig2` crate, you should be able to create a new wrapper type that implements the serialization no matter how complex.

crates/secret-service/src/main.rs
```diff
@@ -0,0 +1,361 @@
...
+
+impl Musig2Signer<Server, ServerFirstRound> for Ms2Signer {
+    fn new_session(
+        &self,
+        public_keys: Vec<PublicKey>,
+    ) -> impl Future<Output = ServerFirstRound> + Send {
+        async move {
+            let nonce_seed = thread_rng().gen::<[u8; 32]>();
+            let first_round = FirstRound::new(
+                KeyAggContext::new(public_keys),
```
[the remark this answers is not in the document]
Rajil1213: Sure. `rust-bitcoin`'s `Psbt` does allow signing of segwit transactions but it does not support musig2 (specifically key aggregation). So this aggregation needs to be done manually (in this case, with the `musig2` crate). In the second round of musig2, all the individual signatures are aggregated. Now, there are three ways to spend a taproot:

* key-path spend where the taproot address is just an address created using an internal key with no scripts
* script-path spend where the internal key for the taproot is an unspendable key but tweaked with the merkle root of the available scripts, so it can be spent by revealing the script and an inclusion proof of the script in the merkle root
* key-path but that also commits to a merkle root -- in this case, the taproot address "contains" scripts but the internal key is not an unspendable one

During key aggregation, if we are taking the key-path or the key-path (with some merkle root), the `KeyAggContext` needs to be tweaked before aggregating the signatures. There are two methods `.with_unspendable_taproot_tweak` and `.with_taproot_tweak` that you need to invoke. This requires that you pass in some arg that tells this function whether you need to tweak the context. There is a `TaprootWitness` enum somewhere in the codebase to indicate this.

Cargo.toml
```diff
@@ -62,38 +66,35 @@ zkaleido-sp1-adapter = { git = "https://github.com/alpenlabs/zkaleido", tag = "v
 anyhow = "1.0.95"
 arbitrary = { version = "1.4.1", features = ["derive"] }
 ark-bn254 = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
-ark-crypto-primitives = { git = "https://github.com/arkworks-rs/crypto-primitives/" }
 ark-ec = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
 ark-ff = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
 ark-groth16 = { git = "https://github.com/arkworks-rs/groth16" }
-ark-r1cs-std = { git = "https://github.com/arkworks-rs/r1cs-std/" }
-ark-relations = { git = "https://github.com/arkworks-rs/snark/" }
-ark-std = { git = "https://github.com/arkworks-rs/std/" }
```
storopoli, line 71: What's these? Maybe bad rebase?

crates/secret-service-proto/src/v1/traits.rs
```diff
@@ -0,0 +1,164 @@
...
+    type P2PSigner: P2PSigner<O>;
+    type Musig2Signer: Musig2Signer<O, FirstRound>;
+    type WotsSigner: WotsSigner<O>;
+    type StakeChain: StakeChainPreimages<O>;
+
+    fn operator_signer(&self) -> Self::OperatorSigner;
+    fn p2p_signer(&self) -> Self::P2PSigner;
+    fn musig2_signer(&self) -> Self::Musig2Signer;
+    fn wots_signer(&self) -> Self::WotsSigner;
+    fn stake_chain(&self) -> Self::StakeChain;
```
storopoli, line 43: I got the idea but this is **_not_** the Stake Chain.
This is the preimages for a Stake transaction in the chain.
We need better naming.

storopoli: Maybe 
```suggestion
    fn stake_preimages(&self) -> Self::StakePreimages;
```

crates/secret-service-proto/src/v1/wire.rs
```diff
@@ -0,0 +1,244 @@
...
+    OpaqueServerError,
+
+    OperatorSignPsbt {
+        sig: [u8; 64],
+    },
+    OperatorPubkey {
+        pubkey: [u8; 33],
+    },
+
+    SignP2P {
```
storopoli, line 24: I prefer explicitness
```suggestion
    P2PSignature {
```

Also why this is the only one that the context came before the object, e.g. `P2PPubkey` below and others.

crates/secret-service-proto/src/v1/wire.rs
```diff
@@ -0,0 +1,244 @@
...
+        Option<RoundContributionError>,
+    ),
+    Musig2SecondRoundFinalize(Musig2SessionResult),
+
+    WotsGet160Key {
+        key: [u8; 20 * 160],
+    },
+    WotsGet256Key {
+        key: [u8; 20 * 256],
+    },
```
storopoli, line 76: Why these are `Get`?
```suggestion
    Wots160Key {
        key: [u8; 20 * 160],
    },
    Wots256Key {
        key: [u8; 20 * 256],
    },
```
Maybe `s/Key/{Pub,Sec}Key`

crates/secret-service-proto/src/v1/wire.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    StakeChainGetPreimage {
```
storopoli, line 78:
```suggestion
    StakePreimage {
```
Same logic above

crates/secret-service-server/src/ms2sm.rs
```diff
@@ -0,0 +1,252 @@
+use std::{mem::MaybeUninit, sync::Arc};
```
storopoli, line 1: What is this palindrome `ms2sm`? Can't we have a more intuitive naming or module-level docs?

crates/secret-service/src/disk/musig2.rs
```diff
@@ -0,0 +1,247 @@
...
+                        .expect("must be able to tweak the key agg context")
+                }
+                _ => {}
+            }
+
+            let nonce_seed = {
+                let info = {
+                    let mut buf = [0; 36];
+                    buf[0..32].copy_from_slice(&input_txid.as_raw_hash().to_byte_array());
+                    buf[32..36].copy_from_slice(&input_vout.to_le_bytes());
```
storopoli, line 95: Whenever using these hardcoded lengths I prefer to move then to `const` like we did in the `p2p` crate, [example](https://github.com/alpenlabs/strata-p2p/blob/293e16bd8ad5d9e291d2db8d32f92bafcda2f918/crates/types/src/stake_data.rs#L50-L79):

```rs
    /// Converts a [`StakeData`] instance into a flattened byte array.
    ///
    /// The byte array is structured as follows:
    ///
    /// - 5,120 (20 * 256) bytes for the withdrawal fulfillment public key.
    /// - 32 bytes for the hash.
    /// - 36 (32 + 4) bytes for the operator funds
    ///
    /// Total is 5,188 bytes.
    ///
    /// # Implementation Details
    ///
    /// The vout in the operator funds are encoded in little-endian bytes.
    /// [`OutPoint`].
    pub fn to_flattened_bytes(
        &self,
    ) -> [u8; WOTS_SINGLE * 256 + HASH_SIZE + TXID_SIZE + VOUT_SIZE] {
        let mut bytes = [0u8; WOTS_SINGLE * 256 + HASH_SIZE + TXID_SIZE + VOUT_SIZE];
        bytes[0..WOTS_SINGLE * 256]
            .copy_from_slice(&self.withdrawal_fulfillment_pk.to_flattened_bytes());
        bytes[WOTS_SINGLE * 256..WOTS_SINGLE * 256 + HASH_SIZE]
            .copy_from_slice(&self.hash.to_byte_array());
        bytes[WOTS_SINGLE * 256 + HASH_SIZE..WOTS_SINGLE * 256 + HASH_SIZE + TXID_SIZE]
            .copy_from_slice(&self.operator_funds.txid.to_byte_array());
        bytes[WOTS_SINGLE * 256 + HASH_SIZE + TXID_SIZE
            ..WOTS_SINGLE * 256 + HASH_SIZE + TXID_SIZE + VOUT_SIZE]
            .copy_from_slice(&self.operator_funds.vout.to_le_bytes());
        bytes
    }
}
```

storopoli (dismissed):
Looks good!
Some overview comments from a quick review

Rajil1213 (changes requested):
Since this PR is in the testing/documentation phase right now, could you add the following section to the `Cargo.toml` file in each of the crates?

```rust
[lints]
rust.missing_debug_implementations = "warn"
rust.unreachable_pub = "warn"
rust.unused_must_use = "deny"
rust.unsafe_op_in_unsafe_fn = "warn"
rust.missing_docs = "warn"
rustdoc.all = "warn"
```

These are mostly for stricter automatic checks for documentation. There is also a lint for unsafe code since I saw some unsafe blocks in this PR (which will be warn by default in the next rust edition anyway).

Rajil1213:
Looks like the `Cargo.lock` was regenerated at some point in this PR which updated all the unpinned dependencies including `ark-std` and `strata`. I have pinned these down to specific commits and the build seems to be working.

Note that we'll need to update the `strata` deps to its release version at some point during the integration.

CC: @ProofOfKeags @prajwolrg

Rajil1213:
> is this just waiting for an approval then?

The docs are still failing. So that needs fixing. Will review it after the CI is green. But it should be good to go and we should be able to punt any changes down the line.

crates/secret-service-proto/src/v1/wire.rs
```diff
@@ -36,16 +36,10 @@ pub enum ServerMessage {
...
-        sig: [u8; 64],
-    },
-
-    /// Response for [`P2PSigner::pubkey`](super::traits::P2PSigner::pubkey).
-    P2PPubkey {
-        /// Serialized Schnorr [`XOnlyPublicKey`] for P2P signatures.
-        pubkey: [u8; 32],
+    /// Response for [`P2PSigner::secret_key`](super::traits::P2PSigner::secret_key).
+    P2PSecretKey {
+        /// Serialized [`SecretKey`](musig::secp256k1::SecretKey)
```
storopoli: This should be a `bitcoin` not a `musig`

crates/secret-service/src/tests.rs
```diff
@@ -0,0 +1,142 @@
...
+        pki_types::{CertificateDer, PrivatePkcs8KeyDer, ServerName, UnixTime},
+        ClientConfig, ServerConfig,
+    },
+};
+
+use crate::disk::Service;
+
+#[tokio::test]
+async fn e2e() {
+    let server_addr: SocketAddr = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 20000).into();
```
storopoli, line 25:
```suggestion
    let server_addr: SocketAddr = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 20_000).into();
```
Nice port choice

crates/secret-service/src/tests.rs
```diff
@@ -0,0 +1,142 @@
...
+        })
+        .collect::<Vec<_>>();
+    for handle in handles {
+        handle.await.unwrap();
+    }
+
+    // p2p signer
+    let p2p_signer = client.p2p_signer();
+    p2p_signer.secret_key().await.expect("good response");
+}
```
storopoli, line 86: Can we also test a stake preimage, a wots 160 and 256 secret keys, and a secret nonce?

storopoli (changes requested):
few nits.

Ideally we would also mock a MuSig2 first round and second round,
check the [tests vectors in BIP-327](https://github.com/bitcoin/bips/tree/master/bip-0327/vectors)

If you have time for that that would be great!

Rajil1213 (changes requested):
The musig2 stuff might need a few changes. Other than that, the secret service crate should also be broken down. It contains a `library` part that handles the actual logic and a binary part `main.rs` that runs the binary. This should be split out so that the library part of it is what remains inside the `crates` directory while the `main.rs` should be moved to separate binary in the `bin` directory. This library part should also be fully tested (given its nature, I think this is doable). I also think the `disk` module is a bit of a misnomer at this point.

storopoli (dismissed):
ACK 8116631158a19132c561f1e46391746c076ba620

I like the `MakeEven` and it consumes the sk which is a good security feature.

Tests are failing though...

Rajil1213 (changes requested):
The `secret_service` crate needs to be moved to `bin`. I don't know how I feel about the name `seeded_impl` (something like `generator` that lives inside the crates directory might have been better?) but at least it's more accurate now.

Otherwise, this looks mergeable.

storopoli (approved):
ACK 066d984b30c90d24c199ca10148ae1af12829b17

Agree with Rajil's suggestions

Rajil1213:
Also wanted to note that I really like the test coverage for this PR. I was fully expecting the project coverage to go down a bit given the size of the PR but it's actually gone up a percent which is impressive!

Cargo.toml
```diff
@@ -42,17 +46,17 @@ strata-bridge-test-utils = { path = "crates/test-utils" }
...
+strata-common = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-config = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-crypto = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-l1tx = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-primitives = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-proofimpl-btc-blockspace = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-rpc-api = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-rpc-types = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-state = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
```
storopoli, line 59: Why do we need that?

Cargo.toml
```diff
@@ -68,7 +72,7 @@ ark-ff = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-lo
 ark-groth16 = { git = "https://github.com/arkworks-rs/groth16" }
 ark-r1cs-std = { git = "https://github.com/arkworks-rs/r1cs-std/" }
 ark-relations = { git = "https://github.com/arkworks-rs/snark/" }
-ark-std = { git = "https://github.com/arkworks-rs/std/" }
+ark-std = { git = "https://github.com/arkworks-rs/std/", rev = "db4367e68ff60da31ac759831e38f60171f4e03d" }
```
storopoli, line 75: and this?

Cargo.toml
```diff
@@ -137,4 +145,4 @@ ark-serialize = { git = "https://github.com/chainwayxyz/algebra/", branch = "new
 ark-bn254 = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
 ark-crypto-primitives = { git = "https://github.com/arkworks-rs/crypto-primitives/" }
 ark-groth16 = { git = "https://github.com/arkworks-rs/groth16" }
-ark-std = { git = "https://github.com/arkworks-rs/std/" }
+ark-std = { git = "https://github.com/arkworks-rs/std/", rev = "db4367e68ff60da31ac759831e38f60171f4e03d" }
```
storopoli, line 148: and this?

storopoli (changes requested):
Why we are adding `rev` stuff to the git repos?
Just control it in the `Cargo.lock`.
Maybe you should checkout `main`'s `Cargo.lock` and then run a `cargo build` to add the new repos to `main`'s `Cargo.lock`:


```sh
git checkout main -- Cargo.lock
cargo check --workspace
```

Cargo.toml
Rajil1213: This is to avoid having to fix merge conflicts in the lockfile manually. `strata` isn's stable enough for us to just use whatever is in the `main` branch. This pins the dependency to some known working version. Without this, we'd have to solve all the rebase/merge conflicts in the Cargo.lock file manually because a simple `cargo build` (without a lock file) would bring the latest commit from `Strata` which we don't want right now.

Cargo.toml
Rajil1213: The latest version of `ark-std` also has issues and it's even more cumbersome to resolve than the one in `strata` since this is the version that `chainwayxyz` uses in the other `ark-*` stuff.

Cargo.toml
Rajil1213: See the other comment.

bin/secret-service/src/seeded_impl/wots.rs
```diff
@@ -0,0 +1,81 @@
...
+                )
+                .unwrap()
+                .private_key
+                .secret_bytes(),
+        }
+    }
+}
+
+impl WotsSigner<Server> for SeededWotsSigner {
+    async fn get_160_key(&self, txid: Txid, vout: u32, index: u32) -> [u8; 20 * 160] {
```
Rajil1213, line 58: This is a prime candidate for using const generics. Just have `get_key<const SIZE: usize>`. Similarly, the ikm stuff can also use generics since the constructors for both `ikm_160` and `ikm_256`  only differ in the size.

It's also fine to punt this off to later.

bin/secret-service/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use std::{net::SocketAddr, path::PathBuf};
+
+use serde::Deserialize;
+
+/// Configuration for the Secret Service.
+///
+/// It is parsed from a TOML file.
+#[derive(Debug, Deserialize)]
+pub struct TomlConfig {
```
Rajil1213, line 11: Should just be called `Config`. `toml` is just one way to serialize this which may change in the future. Please update the comment as well.

Rajil1213 (changes requested):
Reviewed just the `secret-service` bin. Let's try to get this merged today.

bin/secret-service/src/seeded_impl/wots.rs
Rajil1213: No, this isn't strictly necessary. This suggestion is only for keeping the code DRY.

That said, you don't need to change the wire protocol. You could just have a generic function that is called by  `get_160_key` and `get_256_key` methods.

bin/secret-service/src/config.rs
```diff
@@ -0,0 +1,45 @@
+//! Configuration for the Secret Service.
+
+use std::{net::SocketAddr, path::PathBuf};
+
+use bitcoin::Network;
+use serde::Deserialize;
+
+/// Configuration for the Secret Service.
+///
+/// It is parsed from a TOML file.
```
Rajil1213, line 10: This comment can be improved. It is the caller (`main.rs`) that decides what file format to use, not the library.

bin/secret-service/src/seeded_impl/wots.rs
```diff
@@ -0,0 +1,79 @@
...
+                .secret_bytes(),
+            ikm_256: base
+                .derive_priv(
+                    SECP256K1,
+                    &[
+                        // TODO: move to constants.
+                        ChildNumber::from_hardened_idx(79).unwrap(),
+                        ChildNumber::from_hardened_idx(256).unwrap(),
+                        ChildNumber::from_hardened_idx(0).unwrap(),
+                    ],
```
Rajil1213, line 46: Needs to be replaced with the consts.

crates/primitives/src/secp.rs
```diff
@@ -0,0 +1,30 @@
...
+
+impl Deref for EvenSecretKey {
+    type Target = SecretKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl AsRef<SecretKey> for EvenSecretKey {
```
Rajil1213, line 17: Is this necessary since we already have `Deref`?

Cargo.toml
storopoli: Ok!

bin/secret-service/src/tls.rs
```diff
@@ -0,0 +1,73 @@
...
+        let client_cert_verifier = WebPkiClientVerifier::builder(root_store.into())
+            .build()
+            .expect("valid client verifier");
+        ServerConfig::builder().with_client_cert_verifier(client_cert_verifier)
+    } else if *DEV_MODE {
+        warn!("⚠️ no CA certificate provided, disabling client authentication");
+        ServerConfig::builder().with_no_client_auth()
+    } else {
+        error!("TLS configuration is missing CA certificate path");
+        std::process::exit(11);
```
Rajil1213, line 57: Isn't exit code 11 a segfault?

bin/secret-service/src/tls.rs
```diff
@@ -0,0 +1,73 @@
...
+        (cert_chain, key)
+    } else if *DEV_MODE {
+        warn!("⚠️ using self-signed certificate");
+        let cert = rcgen::generate_simple_self_signed(vec!["localhost".into()]).unwrap();
+        let key = PrivatePkcs8KeyDer::from(cert.key_pair.serialize_der());
+        let cert = cert.cert.into();
+        (vec![cert], key.into())
+    } else {
+        error!("TLS configuration is missing certificate and key paths");
+        std::process::exit(10);
```
Rajil1213, line 37: What's exit code 10?

Rajil1213 (changes requested):
Looks good.

Just noticed the exit codes on `load_tls`. So left a few comments there.

Rajil1213 (approved):
All good now. 🥳

PR #34, opened 2025-02-11
Updated proof statements and refactors

prajwolrg:
## Description

- Verifies the signature using `wallet_pk` instead of `signing_pk`
- Move `BitcoinTx` from `strata-bridge-primitives` to `strata-bridge-proof-primitives`. This is done because `strata-bridge-primitives` has dependency on `bitvm` which has dependency on `tokio` which cannot be compiled inside ZkVm.
- Adds mock support for SP1

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [X] Refactor
- [X] New or updated tests
- [ ] Dependency Update

## Checklist

- [X] I have performed a self-review of my code.
- [X] I have commented my code where necessary.
- [X] I have updated the documentation if needed.
- [X] My changes do not introduce new warnings.
- [X] I have added tests that prove my changes are effective or that my feature works.
- [X] New and existing tests pass with my changes.


Cargo.toml
```diff
@@ -34,6 +34,7 @@ strata-bridge-db = { path = "crates/db" }
 strata-bridge-guest-builder = { path = "bridge-guest-builder" }
 strata-bridge-primitives = { path = "crates/primitives" }
 strata-bridge-proof-protocol = { path = "crates/bridge-proof/protocol" }
+strata-bridge-proof-protocol2 = { path = "crates/bridge-proof/protocol2" }
```
storopoli, line 37: Is this temporary?

prajwolrg: `protocol2` will replace `protocol` eventually.

storopoli: Then maybe add a TODO to remove `protocol` or rename `protocol2` -> `protocol`

crates/agent/src/operator.rs
```diff
@@ -2018,30 +2018,50 @@ where
...
+                            (height - l1_start_height) as usize,
+                        )
+                    })
+            });
 
-            if withdrawal_fulfillment.is_none() {
-                // check and get withdrawal fulfillment txid with proof
-                if let Some(tx) = block
+            // Only set `withdrawal_fulfillment` if it's currently `None` and we find a matching tx
+            withdrawal_fulfillment = withdrawal_fulfillment.or_else(|| {
```
Rajil1213, line 2046: Nice refactor!

Rajil1213 (approved):
Looks good. I see that you've refactored the `agent` code as well. These are good but most of these will be refactored more heavily in the final implementation too.

Now, we only need to replace `protocol` with `protocol2`. Note that the `strata-bridge-protocol` crate is still used in the `verifier.rs` file. Once it's removed from there too, the build should failed because of the unused dependency in the `Cargo.toml` file for the `strata-bridge-agent` crate.

prajwolrg:
> Looks good. I see that you've refactored the `agent` code as well. These are good but most of these will be refactored more heavily in the final implementation too.
> 
> Now, we only need to replace `protocol` with `protocol2`. Note that the `strata-bridge-protocol` crate is still used in the `verifier.rs` file. Once it's removed from there too, the build should fail because of the unused dependency in the `Cargo.toml` file for the `strata-bridge-agent` crate.

Yes, will work on completely replacing `protocol2`.

Rajil1213 (approved):

PR #32, opened 2025-02-07
export Subscription directly from subscription module

ProofOfKeags:
### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [x] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.

PR #39, opened 2025-02-12
DutyTracker MegaPR

ProofOfKeags:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


crates/btc-notify/src/lib.rs
```diff
@@ -12,4 +12,4 @@ mod config;
 mod constants;
 mod event;
 mod state_machine;
-mod subscription;
+pub mod subscription;
```
storopoli, line 15: This was already merged in #32.
Maybe needs rebasing.

crates/duty-tracker/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+rust.unused_crate_dependencies = "deny"
+rust.unused_must_use = "deny"
+rustdoc.all = "warn"
+
+[dependencies]
+btc-notify.workspace = true
+bitcoin.workspace = true
+futures = "0.3.31"
+strata-btcio.workspace = true
+tokio = "1.43.0"
```
storopoli, line 23: You can use the workspace one for `futures` and `tokio`.

crates/duty-tracker/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+};
+
+use bitcoin::{
+    opcodes::all::OP_RETURN, Amount, Network, Script, ScriptBuf, Transaction, XOnlyPublicKey,
+};
+use btc_notify::client::BtcZmqClient;
+use futures::{poll, FutureExt, Stream, StreamExt};
+use tokio::task::JoinHandle;
+
+const DUST_VALUE: Amount = Amount::from_sat(330);
```
storopoli, line 16: Ugh, I am also using that `const` in https://github.com/alpenlabs/strata-bridge/blob/437f2db78b630898c17e8cbbf305ebe7c06e3678/crates/stake-chain/src/transactions/constants.rs?plain=1#L29-L30

Maybe we should move this to a `primitives` crate? What you think?

crates/duty-tracker/src/lib.rs
```diff
@@ -0,0 +1,131 @@
+mod tx_driver;
```
storopoli, line 1: Bunch of `const` here, maybe move to a `src/constants.rs` file.

storopoli (commented):
Excited to see this coming to life.
Few comments

Cargo.toml
```diff
@@ -19,7 +19,7 @@ members = [
   "bin/assert-splitter",
 
   # test utilities
-  "crates/test-utils",
+  "crates/test-utils", "crates/duty-tracker",
```
Rajil1213, line 22: Nit: should be grouped with the rest of the crates (instead of under the test utils).

crates/duty-tracker/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    if let Some(Ok(bitcoin::script::Instruction::PushBytes(bytes))) = instructions.next() {
+        Some(bytes.as_bytes())
+    } else {
+        None
+    }
+}
+
+fn magic_tagged_data(script: &Script) -> Option<&[u8]> {
+    const MAGIC_BYTES: &[u8; 6] = b"strata";
```
Rajil1213, line 34: Just a heads up -- the plan is to move this to a params file that can be passed to the `strata-bridge` binary.

crates/duty-tracker/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    op_return_data(script).and_then(|data| {
+        if data.starts_with(MAGIC_BYTES) {
+            Some(&data[MAGIC_BYTES.len()..])
+        } else {
+            None
+        }
+    })
+}
+
+const EL_ADDR_SIZE: usize = 20;
```
Rajil1213, line 44: This will also be part of the params file (since we want to support multiple execution environments whose address sizes might be different).

crates/duty-tracker/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+fn is_strata_checkpoint_transaction(tx: &Transaction) -> bool {
+    todo!()
+}
+
+// Duties, like transactions are ultimately events. These events are generated in reaction to other
+// events.
+
+pub enum OperatorDuty {
```
Rajil1213, line 64: It's fine to have it here but does this change actually belong in this PR (as opposed to the state-machine PR you're working on)?

crates/duty-tracker/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    CollectSignatures,  // Originates when detecting a deposit request
```
Rajil1213, line 65: This is actually multiple steps:

1. Publish WOTS public keys (for the graph)
2. Publish and collect covenant nonces (for the graph)
3. Publish and collect covenant signatures (for the graph)
4. Publish and collect nonces (for the deposit)
5. Publish and collect signatures (for the deposit)

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -0,0 +1,126 @@
...
+                            btc_notify::client::TxStatus::Unknown => {
+                                // Transaction has been evicted, resubmit and see what happens
+                                match rpc_client.send_raw_transaction(&event.rawtx).await {
+                                    Ok(_txid) => { /* NOOP, we good fam */ }
+                                    Err(_err) => {
+                                        // TODO(proofofkeags): in this case we need to analyze the
+                                        // reported error. There are a few scenarios that can play
+                                        // out here.
+                                        //
+                                        // 1. It failed because one or more of the inputs is double
```
Rajil1213, line 80: Note: only need to report this as a warning (usually means another path was taken in the transaction graph). There's nothing we can do here.

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -0,0 +1,126 @@
...
+                                        // spent.
+                                        // 2. It failed because the fee didn't exceed the purge
```
Rajil1213, line 82: Note: This is recoverable via fee-bumping. Might be non-recoverable if there aren't sufficient funds even after a finite number of retries.

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -0,0 +1,126 @@
...
+                                        // rate.
+                                        // 3. If failed because the transaction has already
```
Rajil1213, line 84: Note: I've not observed this error (at least when running unit tests in `corepc-node`. It does error when you do the `testmempoolaccept` rpc call).

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -0,0 +1,126 @@
...
+                                        // re-entered the mempool automatically upon reorg.
+                                    }
+                                }
+                            }
+                            btc_notify::client::TxStatus::Mined { blockhash } => {
+                                // Since our responsibility ends at block inclusion we will send an
+                                // answer on the response channel now. It is the API caller's
+                                // responsibility for handling reorgs after inclusion.
```
Rajil1213, line 92: This might complicate things. Ideally, the terminal state needs to be `Buried` and only reorgs longer than the bury depth should be handled by the caller because that's a bit extreme.

Rajil1213 (changes requested):
Looks like this PR covers some of the duty-tracker state-machine stuff as well. I've mostly left some implementation notes. The only real request for change is in the terminal state for the tx-driver events.

crates/duty-tracker/src/lib.rs
storopoli: What about for the Stake Chain and other genesis related "duties" that are not tied to a deposit?

ProofOfKeags: Not part of the contract state machine. The state machine is a per-contract (deposit) state machine. These will need to be housed in a larger routing structure.

crates/duty-tracker/src/predicates.rs
```diff
@@ -0,0 +1,81 @@
...
+    op_return_data(script).and_then(|data| {
+        if data.starts_with(MAGIC_BYTES) {
+            Some(&data[MAGIC_BYTES.len()..])
+        } else {
+            None
+        }
+    })
+}
+
+const EL_ADDR_SIZE: usize = 20;
```
storopoli, line 38: This can be moved to the `crates/primitives/src/params/strata.rs`

crates/duty-tracker/src/predicates.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub(crate) fn is_deposit_request(tx: &Transaction) -> bool {
+    const MERKLE_PROOF_SIZE: usize = 32;
```
storopoli, line 41: This can be moved to the `crates/primitives/src/params/strata.rs`

storopoli (changes requested):

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -0,0 +1,319 @@
...
+                            deposit_amount: BRIDGE_DENOMINATION,
+                            operator_pubkey: key.x_only_public_key().0,
+                        };
+                        PegOutGraph::generate(
+                            input,
+                            &self.build_context,
+                            deposit_tx.compute_txid(),
+                            *idx,
+                            StakeChainParams::default(),
+                            todo!(), // Where am I supposed to get these from @Rajil1213?
```
Rajil1213, line 168: The previous claim txids are `txid`'s of the Claim Tx for the previous contracts. This is needed for slashing the operator's stake if they advance the stake chain while some previous claim is in progress.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -0,0 +1,319 @@
...
+                            todo!(), /* I think the only way to get this parameter is to move
```
Rajil1213, line 169: This one we can move to the `input`.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -0,0 +1,319 @@
...
+                let peg_out_graphs = self
+                    .build_context
+                    .pubkey_table()
+                    .0
+                    .iter()
+                    .map(|(idx, key)| {
+                        let input = PegOutGraphInput {
+                            stake_outpoint: todo!(),                  // @Rajil1213
+                            withdrawal_fulfillment_outpoint: todo!(), // @Rajil1213
+                            stake_hash: todo!(),                      // @Rajil1213
```
Rajil1213, line 157: All these will be part of the p2p message being relayed during the deposit setup. Relevant PR: https://github.com/alpenlabs/strata-p2p/pull/37

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -0,0 +1,153 @@
...
+        PersistErr
+    }
+}
+
+/// System for persisting the relevant data for [`crate::contract_state_machine::ContractSM`]
+#[derive(Debug)]
+pub struct ContractPersister {
+    // TODO(proofofkeags): figure out how to avoid monomorphizing to Sqlite. We'd like the
+    // persister to be generalized over SQL implementations.
+    pool: Pool<Sqlite>,
```
Rajil1213, line 33: General idea is to have a trait and to implement that trait over the `SqliteDb` type. See the `db` crate.

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+impl ContractPersister {
+    /// Initializes the [`ContractPersister`]
+    pub async fn new(pool: Pool<Sqlite>) -> Result<Self, PersistErr> {
+        let _: SqliteQueryResult = sqlx::query(
```
Rajil1213, line 38: This should just be a migration.

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -0,0 +1,153 @@
...
+        )
+        .execute(&pool)
+        .await
+        .map_err(|_| PersistErr)?;
+        Ok(ContractPersister { pool })
+    }
+
+    /// Initializes a new contract with the given [`ContractCfg`] and [`MachineState`].
+    pub async fn init(&self, cfg: &ContractCfg, state: &MachineState) -> Result<(), PersistErr> {
+        let _: SqliteQueryResult = sqlx::query(
```
Rajil1213, line 59: Using the `query` function instead of the `query!` macro means that we don't get any compile-time checks for these queries. You can check the `db` crate for how to use that macro.

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -0,0 +1,153 @@
...
+            INSERT INTO contracts (deposit_txid, deposit_tx, operator_set, perspective, state) VALUES (?, ?, ?, ?, ?)
+            "#
+        )
+        .bind(cfg.deposit_tx.compute_txid().to_string())
+        .bind(bincode::serialize(&cfg.deposit_tx)?)
+        .bind(bincode::serialize(&cfg.operator_set)?)
+        .bind(cfg.perspective)
+        .bind(bincode::serialize(&cfg.peg_out_graphs)?)
+        .bind(bincode::serialize(&state)?)
+        .execute(&self.pool)
```
Rajil1213, line 70: It is preferable to have the serialization logic for each of the fields to be decoupled from the database logic (reads/writes). This can be done by introducing wrapper types and models. This reduces the surface area for bugs arising out of inconsistent serialization/deserialization logic (for example, `Txid` can be serialized directly as a string or consensus serialized which has the same endianness as in bitcoin).

That said, this is something I can help out with since I've done this before. So, it should not be something that you should worry about atm.

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    /// Updates the [`MachineState`] for a contract.
+    pub async fn commit(
+        &self,
+        deposit_txid: &Txid,
+        state: &MachineState,
+    ) -> Result<(), PersistErr> {
+        let _: SqliteQueryResult = sqlx::query(
+            r#"
+            UPDATE contracts SET state = ? WHERE deposit_txid = ?;
```
Rajil1213, line 84: This should be done over a transaction (we can punt this to later).

crates/duty-tracker/src/predicates.rs
```diff
@@ -0,0 +1,102 @@
...
+    Some(DepositInfo::new(
+        OutPoint::new(tx.compute_txid(), 0),
+        el_addr.to_vec(),
+        deposit_request_output.value,
+        TapNodeHash::from_slice(take_back_leaf_hash).unwrap(),
+        deposit_request_output.script_pubkey.clone(),
+    ))
+}
+
+pub(crate) fn is_txid(txid: Txid) -> TxPredicate {
```
Rajil1213, line 70:
```suggestion
pub(crate) fn matches_txid(txid: Txid) -> TxPredicate {
```
?

Rajil1213 (changes requested):
Did a first pass of this PR.

Most comments are related to the persister. It should be moved to the `db` crate. It should also follow the patterns used there but that isn't the priority at the moment.

I've also addressed the TODOs assigned to me.

crates/duty-tracker/src/contract_manager.rs
ProofOfKeags: For all previous contracts or just the ones that aren't terminal?

crates/duty-tracker/src/contract_manager.rs
ProofOfKeags: We can but that doesn't help us with respect to how to get it. Unless you're saying that there is another function that can generate the input and include those

Rajil1213: Yeah, I meant that it can be constructed in the same way that the rest of the fields in the input are constructed. There is almost a 1:1 correspondence between this input type and the DepositSetup type in the p2p crate.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -40,8 +51,7 @@ impl ContractManager {
                     })
                     .collect::<BTreeMap<Txid, ContractSM>>(),
                 Err(_) => {
-                    debug_assert!(false, "Failed to load contracts");
-                    BTreeMap::new()
+                    todo!() // TODO(proofofkeags): probably wanna crash here?
```
storopoli, line 54: Fine to crash for now.
But should try to back off for a `const` configurable time and then try again for `const` configurable number of times; and then CRASH!

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -58,14 +68,21 @@ impl ContractManager {
             let mut block_sub = zmq_client.subscribe_blocks().await;
             loop {
                 tokio::select! {
-                    Some(block) = block_sub.next() => if let Err(e) = ctx.process_block(block).await {
-                        todo!() // TODO(proofofkeags): error handling
+                    Some(block) = block_sub.next() => {
+                        let blockhash = block.block_hash();
+                        if let Err(e) = ctx.process_block(block).await {
+                            tracing::error!("failed to process block {}: {}", blockhash, e);
```
storopoli, line 74: full import.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -58,14 +68,21 @@ impl ContractManager {
...
+                            break;
+                        }
                     },
                     Some(event) = p2p_handle.next() => match event {
-                        Ok(Event::ReceivedMessage(msg)) => if let Err(e) = ctx.process_p2p_message(msg).await {
-                            todo!() // TODO(proofofkeags): error handling
+                        Ok(Event::ReceivedMessage(msg)) => if let Err(e) = ctx.process_p2p_message(msg.clone()).await {
+                            tracing::error!("failed to process p2p msg {:?}: {}", msg, e);
```
storopoli, line 80: full import

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -58,14 +68,21 @@ impl ContractManager {
             let mut block_sub = zmq_client.subscribe_blocks().await;
             loop {
                 tokio::select! {
```
storopoli, line 70: full import

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -58,14 +68,21 @@ impl ContractManager {
...
                     Some(event) = p2p_handle.next() => match event {
-                        Ok(Event::ReceivedMessage(msg)) => if let Err(e) = ctx.process_p2p_message(msg).await {
-                            todo!() // TODO(proofofkeags): error handling
+                        Ok(Event::ReceivedMessage(msg)) => if let Err(e) = ctx.process_p2p_message(msg.clone()).await {
+                            tracing::error!("failed to process p2p msg {:?}: {}", msg, e);
+                            break;
                         },
-                        Err(e) => todo!(),
+                        Err(e) => {
+                            tracing::error!("{}", e);
```
storopoli, line 84: full import
Also we prefer
```suggestion
                            error!(%e "other unknown error");
```

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -155,36 +214,37 @@ impl ContractManagerCtx {
...
     }
 
     async fn process_p2p_message(&mut self, msg: GossipsubMsg) -> Result<(), ContractManagerErr> {
         match msg.unsigned {
-            UnsignedGossipsubMsg::StakeChainExchange {
-                stake_chain_id,
-                info,
-            } => todo!(),
+            UnsignedGossipsubMsg::StakeChainExchange { .. } => todo!(),
             UnsignedGossipsubMsg::DepositSetup { scope, wots_pks } => {
```
storopoli, line 241: This will change in https://github.com/alpenlabs/strata-p2p/pull/37 to https://github.com/alpenlabs/strata-p2p/blob/82ccb672589d2a2ea8e7dbb42e0a8bfe893c8b33/crates/wire/src/p2p/v1/typed.rs#L231-L255:

```rs
    /// New deposit request appeared, and operators
    /// exchanging setup data.
    ///
    /// This is primarily used for the WOTS PKs.
    DepositSetup {
        /// [`Scope`] of the deposit data.
        scope: Scope,

        /// [`sha256::Hash`] hash of the deposit data.
        hash: sha256::Hash,

        /// Funding transaction ID.
        ///
        /// Used to cover the dust outputs in the transaction graph connectors.
        funding_txid: Txid,

        /// Funding transaction output index.
        ///
        /// Used to cover the dust outputs in the transaction graph connectors.
        funding_vout: u32,

        /// [`WotsPublicKeys`] of the deposit data.
        wots_pks: WotsPublicKeys,
    },
```

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -155,36 +214,37 @@ impl ContractManagerCtx {
...
         Ok(())
     }
 
     async fn process_p2p_message(&mut self, msg: GossipsubMsg) -> Result<(), ContractManagerErr> {
         match msg.unsigned {
-            UnsignedGossipsubMsg::StakeChainExchange {
-                stake_chain_id,
-                info,
-            } => todo!(),
+            UnsignedGossipsubMsg::StakeChainExchange { .. } => todo!(),
```
storopoli, line 240: And this will change in https://github.com/alpenlabs/strata-p2p/pull/37 to https://github.com/alpenlabs/strata-p2p/blob/82ccb672589d2a2ea8e7dbb42e0a8bfe893c8b33/crates/wire/src/p2p/v1/typed.rs#L219-L230:

```rs
    /// Operators exchange stake chain info.
    StakeChainExchange {
        /// 32-byte hash of some unique to stake chain data.
        stake_chain_id: StakeChainId,

        /// [`Txid`] of the pre-stake transaction.
        pre_stake_txid: Txid,

        /// vout of the pre-stake transaction.
        pre_stake_vout: u32,
    },
```

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -81,12 +82,19 @@ pub enum ContractState {
         /// been converted to a deposit.
         abort_deadline: BitcoinBlockHeight,
 
-        /// An accumulator for all of the wots keys we have received from our peers.
+        /// This is a collection of the wots keys we have received from our peers.
```
storopoli, line 85: Accumulator would be nice, but maybe V2 😄

storopoli (changes requested):

crates/duty-tracker/src/contract_manager.rs
ProofOfKeags: Trying to tighten this. Is it the set of all active claim txids for contracts who have the same assigned fulfiller?

Rajil1213: No, these txids are used for constructing the `Slash Stake` transactions. These transactions are created on a per-graph (deposit) basis. The number of these transactions depends on the number of past non-terminal claims (and a  parameter that dictates the maximum number of such transactions to create). Each previous claim txid corresponds to a `Slash Stake` transaction.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -0,0 +1,906 @@
...
+            ContractEvent::DepositConfirmation(tx, deposit_idx) => {
+                self.process_deposit_confirmation(tx, deposit_idx)
+            }
+            ContractEvent::PegOutGraphConfirmation(tx, height) => {
+                self.process_peg_out_graph_tx_confirmation(height, &tx)
+            }
+            ContractEvent::Block(height) => self.notify_new_block(height),
+            ContractEvent::ClaimFailure => self.process_claim_verification_failure(),
+            ContractEvent::AssertionFailure => self.process_assertion_verification_failure(),
+            ContractEvent::Assignment(deposit_entry) => self.process_assignment(&deposit_entry),
```
storopoli, line 396: This is everything! The blueprint. Nice that you kept the ordering proper here.

crates/duty-tracker/src/contract_manager.rs
ProofOfKeags: Shouldn't the length of this vec be 0 then always?

Rajil1213: How so? It will only be zero if every prior contract has reached their terminal stage.

ProofOfKeags: Ok, so this is the vec of all claims that are ongoing for *all* operators, not just the current one.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -171,16 +171,16 @@ impl ContractManagerCtx {
                     })
                     .collect::<Result<BTreeMap<OperatorIdx, PegOutGraphSummary>, TxGraphError>>()?;
                 let (sm, duty) = ContractSM::new(
-                    self.build_context.own_index(),
+                    todo!(), // TODO(proofofkeags): We need to get our own operator pubkey.
```
storopoli, line 174: This can be done by the Secret Service in https://github.com/alpenlabs/strata-bridge/blob/33bb8af6425369d23fa9ebf9e2acd1bdc1e0e35e/crates/secret-service-client/src/musig2.rs#L68-L77

@azz can you confirm that this is the operator pubkey that is used to identify him in the `OperatorPubkeyTable`?

ProofOfKeags: Uhh it's unlikely. We don't do MuSig for signing P2P messages, do we?

storopoli: Then you can do this: https://github.com/alpenlabs/strata-bridge/blob/33bb8af6425369d23fa9ebf9e2acd1bdc1e0e35e/crates/secret-service-client/src/p2p.rs#L31-L40

and create a new `strata-p2p-impl::Configuration`: https://github.com/alpenlabs/strata-bridge/blob/33bb8af6425369d23fa9ebf9e2acd1bdc1e0e35e/crates/p2p-impl/src/config.rs#L9-L34

and call bootstrap: https://github.com/alpenlabs/strata-bridge/blob/33bb8af6425369d23fa9ebf9e2acd1bdc1e0e35e/crates/p2p-impl/src/bootstrap.rs#L14

storopoli: I'll add an easy constructor to the `strata-p2p-impl::Configuration` that takes a `SecretKey` and also a method that returns an `XOnlyPublicKey`.

storopoli (approved):
ACK 40503745ad0582306a015a493d4367b94590ec52

PR #45, opened 2025-02-26
chore: update nightly

storopoli:
## Description

Despite https://github.com/rust-lang/rust/issues/133199#issuecomment-2630615573 saying that:

> The "fix" is to add `#![feature(generic_const_exprs)]` to the top of every crate that depends on your crate, and to the top of every doctest.

it does not work.

I tried the approach that I add `#![feature(generic_const_exprs)]` to every crate and doctest in the `strata-bridge` then keep removing one-by-one until I cannot.

I thought that it would be easier to find the minimal satisfiable set of `#![feature(generic_const_exprs)]`.

Yet, somehow this does not work and I triple-checked that I added in everything that has a nested `Cargo.toml`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

Leaving this open if someone has a better idea or insight.

## Checklist


Rajil1213:
Looks like the issue is because the generics in the bitvm repo are not local. The most straightforward solution is to bring all the contents of the `bitvm::signatures::wots_api` to `strata-bridge-primitives`.

Here's the source:

https://github.com/alpenlabs/BitVM/blob/f7fd93e9c9187100e7c9b7b68e07f9f7c568a843/bitvm/src/signatures/wots_api.rs

storopoli:
Closing this for now. We can always revisit later.

PR #48, opened 2025-02-27
add height support to transaction events

ProofOfKeags:
## Description

It has been in the backlog to have transaction events include block heights. It turns out that bip34 offers a consensus enforced way of getting this in the block data itself. It is brittle in the context of testnet4 and regtest right now. For that reason this implementation will set this value to zero if it cannot be recovered which confers a responsibility on the consumer to check for that for the time being.

Once rust-bitcoin 0.33.0 lands, this won't be necessary any longer since that is the first release where this bug will be fixed. For now we dodge the scenarios where this value isn't properly defined by enforcing that the test code doesn't use block heights between 0 and 16.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,7 @@ ark-std = { git = "https://github.com/arkworks-rs/std/" }
 async-trait = "0.1.81"
 base64 = "0.22.1"
 bincode = "1.3.3"
+# TODO(proofofkeags): when updating `bitcoin` to >=0.33.0, go through btc-notify and update tests to no longer constrain heights.
 bitcoin = { version = "0.32.5", features = ["rand-std", "serde"] }
 bitcoin-bosd = "0.4.0"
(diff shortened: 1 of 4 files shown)


Rajil1213 (changes requested):
Looks good.

In the block tracker PR, I assumed that we'd get the height from the `blockhash` via the `getblock` RPC (with verbosity = 1). It helps to have the height be there directly too.

As an extra measure, maybe add a comment next to the `bitcoin` dependency in the workspace `Cargo.toml` to remember to remove the special casing once we update to  `>=0.33`?

storopoli (dismissed):
ACK d13a00c94ef90e9b4e6b0d95bb399abd07a69934

👍🏻 for Rajil's comment on the `Cargo.toml` rust-bitcoin new version.

Rajil1213 (approved):

PR #52, opened 2025-03-03
STR-996 RPC traits for monitoring the bridge

krsnapaudel:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

[STR-1141](https://alpenlabs.atlassian.net/browse/STR-1141)

[STR-1141]: https://alpenlabs.atlassian.net/browse/STR-1141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[serde(rename = "Complete")]
+    Complete,
+}
+
+/// Represents deposit transaction details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcDepositInfo {
+    pub deposit_request_txid: Txid,
+    pub deposit_txid: Option<Txid>, // Some deposits may not have completed yet
```
Rajil1213, line 33: It is possible to construct the deposit transaction with just the information about the deposit request. So, the deposit txid will always be available. This means that we don't need this to be an `Option`. The completion status can be tracked via the `status` field.

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub status: TransactionStatus,
+}
+
+/// Represents withdrawal transaction details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcWithdrawalInfo {
+    // Some withdrawals may still be pending
+    pub fulfillment_txid: Option<Txid>,
+    pub status: TransactionStatus,
```
Rajil1213, line 43: Slightly different than the above but whether a withdrawal is complete or not does not need to be tracked twice here either. For example, what does it mean for the `fulfillement_txid` to be `None` and `status` to be `Complete` or the other way around?

When possible, we should try to make [invalid states unrepresentable](https://youtu.be/z-0-bbc80JM). You should look into tagged enums in serde and refactor this type to something like the following:

```rust
pub struct RpcWithdrawalInfo {
     Unfulfillled,

      Challenged {
              fulfilllment_txid: Txid,
              claim_txid: Txid,
              ...
        }
 }```

File: crates/rpc/bridge/src/types.rs [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+    Challenge,
+    Assert,
+}
+
+/// Represents reimbursement claim details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcClaimInfo {
+    pub claim_txid: Txid,
+    pub challenge_step: ChallengeStep, // Fixed typo from "challege_step"
```
Rajil1213, line 60: Why is this comment here?

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,63 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum OperatorStatus {
+    Online,
+    Offline,
+}
+
+/// Transaction status variants with custom serialization
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum TransactionStatus {
```
Rajil1213, line 14: Related to the other comments but if this type is supposed to represent the status of an individual transaction, it should be refactored to hold the Txid when the status is in `InProgress` state, and the mined block height when the status is in `Complete` state. Also why is the `Challenge` state in this enum? If it is necessary, perhaps it is better to renamed this enum to `WithdrawalStatus` instead of `TransactionStatus`.

Rajil1213 (changes requested):
All of the methods in the `MonitoringApi` should be possible to implement without much fuss.

However, some of the ways in which the RPC types are laid out could use some polish.

Also adding @ProofOfKeags to the reviewers so that he can assess whether its possible with the testnet-i version of the State machine persistence to implement these RPCs.

crates/rpc/bridge/src/types.rs
krsnapaudel: Ah this is from a copy-paste of AI suggestion.

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,89 @@
...
+#[serde(rename_all = "PascalCase")]
+pub enum RpcOperatorStatus {
+    Online,
+    Offline,
+}
+
+/// Represents a valid deposit status
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "status")]
+pub enum RpcDepositStatus {
```
Rajil1213, line 15: It does not make sense for each of the variants to hold the same two values: `deposit_request_txid` and `deposit_txid`. These should be separate fields and there should be another separate field that just has the status.

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,89 @@
...
+#[serde(rename_all = "PascalCase")]
+pub enum ChallengeStep {
+    Claim,
+    Challenge,
+    Assert,
+}
+
+/// Represents reimbursement claim details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
Rajil1213, line 86: Why is this `camelCase` while the rest are in `PascalCase`?

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Deposit exists, but minting hasn't happened yet.
+    #[serde(rename = "In progress")]
+    InProgress {
+        deposit_request_txid: Txid,
+        deposit_txid: Option<Txid>,
+    },
+
+    /// Deposit exists, but was never completed (can be reclaimed).
+    #[serde(rename = "Failed")]
+    Failed { deposit_request_txid: Txid },
```
Rajil1213, line 25: It might be useful to have this status hold the reason for the failure as well (this could just be a String).

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,89 @@
...
+    Complete {
+        deposit_request_txid: Txid,
+        deposit_txid: Txid,
+    },
+}
+
+/// Shared status and relevant info for withdrawals and claims
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "status")]
+pub enum RpcWithdrawalStatus {
```
Rajil1213, line 38: The process of withdrawal actually begins with a front-payment to the user by the assigned operator. Do we not want to track that in the bridge monitoring dashboard? If not, the `claim_txid` should be refactored out to its own field as it is common to each of the variants. If we do want to track that, we should have another status such as `Paid` that holds the fulfillment txid.

crates/rpc/bridge/Cargo.toml
```diff
@@ -11,7 +11,10 @@ rust.unused_crate_dependencies = "deny"
 rust.unused_must_use = "deny"
```
Rajil1213, line 11:
```suggestion
rust.unused_must_use = "deny"
rust.missing_docs = "warn"
rustdoc.all = "warn"

```

Rajil1213: These are for enforcing stricter documentation requirements.

crates/rpc/bridge/Cargo.toml
```diff
@@ -11,7 +11,10 @@ rust.unused_crate_dependencies = "deny"
...
 
 [dependencies]
+bitcoin.workspace = true
 jsonrpsee = { workspace = true, features = ["server", "macros", "tokio"] }
+serde.workspace = true
+strata-bridge-primitives.workspace = true
```
Rajil1213, line 17:
```suggestion
strata-bridge-primitives.workspace = true

bitcoin.workspace = true
jsonrpsee = { workspace = true, features = ["server", "macros", "tokio"] }
serde.workspace = true
```

Rajil1213 (changes requested):
The types need a bit more polishing still.

crates/rpc/bridge/src/types.rs
krsnapaudel: The deposit status is based on this from PRD: 

Status: In progress, Failed, or Complete. The table should only show deposits that are In progress or have fewer than 6 confirmations after reaching the Failed or Complete status.
    **In progress:**
        - The deposit has a Deposit request TXID, no Deposit TXID, and no Mint TXID, and the deposit cannot be reclaimed by the user yet.
        - The deposit has a Deposit request TXID, a Deposit TXID, but no Mint TXID.
    
    **Failed:** The deposit has a Deposit request TXID, no Deposit TXID, and no Mint TXID, and the deposit can be reclaimed by the user.

    **Complete:** The deposit has a Deposit request TXID, a Deposit TXID, and a Mint TXID.

crates/rpc/bridge/src/types.rs
krsnapaudel: This is a question for @john-light. I will check with him.

crates/rpc/bridge/src/types.rs
Rajil1213: I see. In that case, the `deposit_txid` must be removed from the `InProgress` variant. There is no way for the bridge to extract the Mint TXID by looking at the strata chain state. In fact, the mint txid is not even part of the chain state and afaik, there is no RPC for retrieving that information. So, as far as the bridge is concerned, the two points under the `InProgress` status in the PRD are covered by the `InProgress` and the `Completed` variants respectively.

crates/rpc/bridge/src/types.rs
Rajil1213: Both `PascalCase` and `camelCase` are JavaScript conventions as far as I know. Elsewhere in the codebase, we have simply been using `snake_case` which is what Rust uses for field names.

Rajil1213 (changes requested):

crates/rpc/bridge/src/types.rs
krsnapaudel: It seems `snake_case` does not need renaming. It's mostly a question of whether to rename enums to lowercase or not.

crates/rpc/bridge/src/types.rs
john-light: > There is no way for the bridge to extract the Mint TXID by looking at the strata chain state. In fact, the mint txid is not even part of the chain state and afaik, there is no RPC for retrieving that information.

Can't the Mint TXID be retrieved from the EVM chain?

krsnapaudel: From @sapinb: Mint TXID -> Impossible ... we dont have a mint txn. We are co-opting the ethereum beaconchain withdrawal mechanism for our bridge ins ... So we can potentially get the ethereum block where it was minted, but there is no txn to get unfortunately.

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,75 @@
...
+    },
+
+    /// Operator was slashed, claim is no longer valid.
+    Cancelled { claim_txid: Txid },
+
+    /// Claim has been successfully reimbursed.
+    Complete {
+        claim_txid: Txid,
+        payout_txid: Txid,
+        fulfillment_txid: Txid,
```
Rajil1213, line 61: Fulfillment should actually be the first thing. The order in which these things happen is:

1. Front-payment (fulfillment)
2. Claim
3. Challenge (or Payout)
4. Assert
5. Complete (or Disprove)

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,75 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcDepositInfo {
+    pub status: RpcDepositStatus,
+}
+
+/// Represents withdrawal transaction details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcWithdrawalInfo {
+    pub status: RpcWithdrawalStatus,
+}
```
Rajil1213, line 75: Are we expecting there to be other fields in these structs? If not, these the `*Info` structs seem needless.

Rajil1213 (changes requested):
Left a comment to clarify what fulfillment is.

Still waiting on light for what to do about Mint Txid but otherwise, this is almost good to go.

crates/rpc/bridge/src/types.rs
krsnapaudel: This is actually outdated. Fulfillment is now in `RpcWithdrawalStatus`. The others are in `RpcReimbursementStatus`.

crates/rpc/bridge/Cargo.toml
```diff
@@ -9,9 +9,15 @@ rust.rust_2018_idioms = { level = "deny", priority = -1 }
...
 rust.unused_crate_dependencies = "deny"
 rust.unused_must_use = "deny"
+rust.missing_docs = "warn"
+rustdoc.all = "warn"
 
 [dependencies]
+strata-bridge-primitives.workspace = true
+
+bitcoin.workspace = true
 jsonrpsee = { workspace = true, features = ["server", "macros", "tokio"] }
```
storopoli, line 19:
```suggestion
jsonrpsee = { workspace = true, features = ["server", "macros"] }
```

We don't need `tokio` for the traits, probably in the `dev-dependencies` but it should not go here.

crates/rpc/bridge/src/lib.rs
```diff
@@ -15,3 +21,39 @@ pub trait StrataBridgeControlApi {
     #[method(name = "uptime")]
     async fn get_uptime(&self) -> RpcResult<u64>;
 }
+
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "stratabridge"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "stratabridge"))]
+pub trait StrataBridgeMonitoringApi {
+    /// Get all bridge operator IDs
+    #[method(name = "bridgeOperators")]
+    async fn get_bridge_operators(&self) -> RpcResult<Vec<OperatorIdx>>;
```
storopoli, line 30: This returns a 
https://github.com/alpenlabs/strata-bridge/blob/24a0e87391aaf0ff240b718cb4a7d23ea0b607e9/crates/primitives/src/types.rs#L10

How useful is that?
Isn't it better to have?
```suggestion
    async fn get_bridge_operators(&self) -> RpcResult<PublicKeyTable>;
```
due to the fact that this gives much more information? https://github.com/alpenlabs/strata-bridge/blob/24a0e87391aaf0ff240b718cb4a7d23ea0b607e9/crates/primitives/src/types.rs#L13-L19

crates/rpc/bridge/src/lib.rs
```diff
@@ -15,3 +21,39 @@ pub trait StrataBridgeControlApi {
...
+
+    /// Query operator status (Online/Offline)
+    #[method(name = "operatorStatus")]
+    async fn get_operator_status(&self, operator_idx: OperatorIdx) -> RpcResult<RpcOperatorStatus>;
```
storopoli, line 34: This might need a new P2P command.
I don't know how to circumvent this.
Which is fine, we should have that.
Cc @Rajil1213

crates/rpc/bridge/src/lib.rs
```diff
@@ -15,3 +21,39 @@ pub trait StrataBridgeControlApi {
...
+    /// Get deposit details using the deposit request outpoint
+    #[method(name = "depositInfo")]
+    async fn get_deposit_info(
+        &self,
+        deposit_request_outpoint: OutPoint,
+    ) -> RpcResult<RpcDepositInfo>;
+
+    /// Get bridge duties assigned to an operator
+    #[method(name = "bridgeDuties")]
+    async fn get_bridge_duties(&self, operator_id: u32) -> RpcResult<Vec<BridgeDuty>>;
```
storopoli, line 45: This should be split into two
```suggestion
    /// Get bridge duties assigned to an operator by his [`XOnlyPublicKey`].
    #[method(name = "bridgeDuties")]
    async fn get_bridge_duties_by_operator_id(&self, operator_pk: XOnlyPublicKey) -> RpcResult<Vec<BridgeDuty>>;
    
    /// Get bridge duties assigned to an operator by [`OperatorIdx`].
    #[method(name = "bridgeDutiesById")]
    async fn get_bridge_duties_by_operator_id(&self, operator_id: u32) -> RpcResult<Vec<BridgeDuty>>;
```

storopoli (changes requested):

crates/rpc/bridge/src/lib.rs
krsnapaudel: Yes. I asked @Rajil1213 if public key is needed for any of the API calls. Apparently not. I am fine either way.

storopoli (changes requested):

storopoli (dismissed):
ACK 907e803720bf899e2eb70b25058949d43eaa8e24

Rajil1213:
This PR still needs an input from @sapinb regarding what to do about the `withdrawal_fulfillment_txid` in the `RpcWithdrawalInfo`.

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,84 @@
...
+/// Represents deposit transaction details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcDepositInfo {
+    pub status: RpcDepositStatus,
+}
+
+/// Represents withdrawal transaction details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcWithdrawalInfo {
+    pub withdrawal_request_txid: Txid,
```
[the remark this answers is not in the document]
sapinb: My understanding was that these rpcs were to display the deposits and withdrawals (historical and current) for the chain. I'm curious how the outpoints (for both deposit and withdrawal infos) will be known in the first place to query these endpoints.
I think the deposit table index could be a more natural input, easier to use as its just an incrementing number. (although bridge dbs likely do not have them indexed by the deposit index).

Also `withdrawal_request_txid` is the txid from the evm chain, not bitcoin chain. So this should be represented as some kind of Buf32 or Hex string.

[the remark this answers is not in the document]
storopoli: Can we drop `withdrawal_request_txid`?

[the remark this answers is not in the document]
Rajil1213: > I think the deposit table index could be a more natural input, easier to use as its just an incrementing number. (although bridge dbs likely do not have them indexed by the deposit index).

It's called stake index in the bridge but it's the same idea and should correspond directly to the deposit id.

> Can we drop withdrawal_request_txid?

@john-light any thoughts on this?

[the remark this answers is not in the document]
john-light: > Not sure if the withdrawal_fulfillment_txid is something that can be extracted from the EVM. I think this is possible but that would require storing it in the chain state as well.

My (perhaps naiive) expectation would be that the backend can observe withdrawal requests on Strata/Alpen and also observe the bitcoin chain for a bitcoin transaction that matches the same amount/ destination address coming from the assigned operator, and pull out that TXID and show it in the frontend here.

> Can we drop withdrawal_request_txid?

Why?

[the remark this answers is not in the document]
john-light: > Because the bridge only communicates with Strata through the L1.

I don't see the relevance of that. The backend of this monitoring dashboard need not be limited to only observing L1.

> And we don't have this since as @sapinb said
>> Also `withdrawal_request_txid` is the txid from the evm chain, not bitcoin chain. So this should be represented as some kind of Buf32 or Hex string.

We most certainly do have it. It is a transaction that happens in the EVM. We should extract it from there.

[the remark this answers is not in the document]
Rajil1213: > We most certainly do have it. It is a transaction that happens in the EVM. We can extract it from there, right?

Yes, but I don't think the bridge node should be the one "talking" to the evm just for the sake of monitoring. The monitoring backend can do this on its own (not exactly sure how but should be doable), and then query the bridge node for any recognized withdrawal duties and correlate from there.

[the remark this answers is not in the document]
krsnapaudel: > My understanding was that these rpcs were to display the deposits and withdrawals (historical and current) for the chain. I'm curious how the outpoints (for both deposit and withdrawal infos) will be known in the first place to query these endpoints. I think the deposit table index could be a more natural input, easier to use as its just an incrementing number. (although bridge dbs likely do not have them indexed by the deposit index).
> 
> Also `withdrawal_request_txid` is the txid from the evm chain, not bitcoin chain. So this should be represented as some kind of Buf32 or Hex string.

Based on the miro diagram on bridge monitoring, `get_bridge_duties` should provide `withdrawal_outpoint`. I realized that, at least according to the miro diagram, it should be `withdrawal_outpoint`. So the outpoint is coming from a previous RPC call to bridge.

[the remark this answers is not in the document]
krsnapaudel: > Correct, this is the interface for the bridge node. Not for the bridge dashboard. This is what the dashboard and anyone else wants from bridge operator's nodes. @krsnapaudel can we refactor and move this specific functionality to the monitoring dashboard backend?

I have removed WRT. However, not sure how we will get it on the monitoring backend.

krsnapaudel:
Based on discussion with @sapinb, we don't need to query bridge duties by operator. Since all operators have duties for others.

crates/rpc/bridge/src/types.rs
```diff
@@ -0,0 +1,87 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcDepositInfo {
+    pub status: RpcDepositStatus,
+}
+
+/// Represents withdrawal transaction details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RpcWithdrawalInfo {
+    pub status: RpcWithdrawalStatus,
+    pub assigned_operator_idx: OperatorIdx,
```
krsnapaudel, line 77: I am not sure how useful this is. This information is already in the `BridgeDuty` fetched using `get_bridge_duties`.

crates/rpc/bridge/src/types.rs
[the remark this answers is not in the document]
Rajil1213: Marking this as resolved for now to unblock other things. At the moment, there is more work that needs to be done elsewhere than on the bridge to make this possible.

Rajil1213 (dismissed):

storopoli (changes requested):
@krsnapaudel can you rebase after https://github.com/alpenlabs/strata-bridge/pull/58 hits `main`, it should be any moment in the next hour.
That will fix all the CI failures

storopoli (changes requested):
Looks good to go, but one minor final nit:
can you move everything from `crates/rpc/bridge` to `crates/rpc`? It does not make sense to have another thing right now as the `rpc`.

storopoli (approved):
ACK 741de90e637edbaaeb5a0ba07751ca20648c6aab

PR #53, opened 2025-03-05
Move WOTS pubkeys to `PegOutGraphInput`

Rajil1213:
## Description

This PR:

* makes `wots::PublicKeys` a part of the `PegOutGraphInput` struct so that it corresponds more directly to `DepositSetup` message in `strata-p2p`.
* fixes a bug in the `agent` crate where the bitcoin chain segment only counted blocks after the last verified bitcoin height  for the given strata checkpoint.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/bridge-proof/protocol/src/statement.rs
+++ b/crates/bridge-proof/protocol/src/statement.rs
@@ -15,7 +15,7 @@ use crate::{
 /// input
 ///
 /// TODO: update this once this is fixed
-const REQUIRED_NUM_OF_HEADERS_AFTER_WITHDRAWAL_FULFILLMENT_TX: usize = 30;
+pub const REQUIRED_NUM_OF_HEADERS_AFTER_WITHDRAWAL_FULFILLMENT_TX: usize = 30;
 
 /// The fixed withdrawal fee for Bitcoin transactions.
 ///
--- a/crates/primitives/src/params/connectors.rs
+++ b/crates/primitives/src/params/connectors.rs
@@ -104,17 +104,11 @@ const _: [(); 0] = [(); (NUM_PKS_A256 + NUM_PKS_A160 - TOTAL_VALUES)];
 
 pub const BLOCK_TIME: Duration = Duration::from_secs(30);
 
-/// The number of blocks expected after the block that contains the withdrawal fulfillment
-/// transaction for a valid proof.
-pub const EXPECTED_BLOCK_COUNT: u32 = 50; // blocks
-
 pub const PAYOUT_OPTIMISTIC_TIMELOCK: u32 = 500;
 
 pub const PRE_ASSERT_TIMELOCK: u32 = PAYOUT_OPTIMISTIC_TIMELOCK + 100; // 100 is slack
 
 // compile-time checks
 const _: () = assert!(PRE_ASSERT_TIMELOCK > PAYOUT_OPTIMISTIC_TIMELOCK);
-
-const _: u32 = PAYOUT_OPTIMISTIC_TIMELOCK - (EXPECTED_BLOCK_COUNT + 100); // 100 is slack
 
 pub const PAYOUT_TIMELOCK: u32 = 288; // 2 day's worth of blocks in mainnet
--- a/crates/primitives/src/params/connectors.rs
+++ b/crates/primitives/src/params/connectors.rs
@@ -104,6 +104,8 @@ const _: [(); 0] = [(); (NUM_PKS_A256 + NUM_PKS_A160 - TOTAL_VALUES)];
 
 pub const BLOCK_TIME: Duration = Duration::from_secs(30);
 
+/// The number of blocks expected after the block that contains the withdrawal fulfillment
+/// transaction for a valid proof.
 pub const EXPECTED_BLOCK_COUNT: u32 = 50; // blocks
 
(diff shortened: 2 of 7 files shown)


crates/primitives/src/params/connectors.rs
```diff
@@ -104,7 +104,7 @@ const _: [(); 0] = [(); (NUM_PKS_A256 + NUM_PKS_A160 - TOTAL_VALUES)];
 
 pub const BLOCK_TIME: Duration = Duration::from_secs(30);
 
-pub const EXPECTED_BLOCK_COUNT: u32 = 100; // blocks
+pub const EXPECTED_BLOCK_COUNT: u32 = 50; // blocks
```
storopoli, line 107: What is this?
Can we document this constant?

crates/tx-graph/src/peg_out_graph.rs
```diff
@@ -33,10 +33,7 @@ use crate::{
 ///
 /// This data is shared between various operators and verifiers and is used to construct the peg out
 /// graph deterministically.
-///
-/// This assumes that the WOTS public keys for the claim/assertions/disprove have already been
-/// shared.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
```
storopoli, line 36: We don't need to serialize these for db purposes?
Maybe we need to do the tuple struct heuristic here if these are from the `bitvm` crate.

storopoli (changes requested):

crates/primitives/src/params/connectors.rs
Rajil1213: I've just removed this and I'm now using the value from the proofs crate to keep things consistent.

crates/tx-graph/src/peg_out_graph.rs
Rajil1213: I've refactored this type even more and now the `PegOutGraphInput` type is almost identical to the `DepositSetup` type in the p2p crate. My understanding is that the p2p crate already stores these in its own database. Is that not the case?

storopoli: Yes, but we're planning to have no database or maybe some message database in TN2 when the P2P is only to "fly authenticated bytes" around.

storopoli: This is one of the reasons we are moving away from hard types in the P2P: code duplication.

crates/bridge-proof/protocol/src/statement.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
 /// input
 ///
 /// TODO: update this once this is fixed
```
storopoli, line 17: What is "this"? Is this still a valid `TODO`?

crates/tx-graph/src/peg_out_graph.rs
```diff
@@ -48,15 +44,14 @@ pub struct PegOutGraphInput {
...
-    /// This is kept as an input instead of a constant to allow for flexibility in the future.
-    pub deposit_amount: Amount,
+    /// The WOTS public keys used to verify commitments to the withdrawal fulfillment txid and the
+    /// Groth16 proof.
+    pub wots_public_keys: wots::PublicKeys,
 
     /// The public key of the operator.
+    ///
+    /// This key is used for CPFP outputs and for receiving reimbursements.
+    // TODO: Make this a [`descriptor`](bitcoin_bosd::Descriptor).
```
storopoli, line 54: We need to open a new ticket under T2 Refactors, can you do it please?

storopoli (approved):
ACK a51ebec300023945ecade2f1e01eeece699f063f

crates/tx-graph/src/peg_out_graph.rs
Rajil1213: I agree. We should definitely do that down the line when we remove the concrete db impl from the p2p crate. I'm not sure we need to do that right now. This will mostly involve replacing all the wots types in the current `strata-bridge` repo with those in the `strata-p2p` repo and implement all the necessary methods on them. `ser/de` can be derived for the rest of the values.

crates/bridge-proof/protocol/src/statement.rs
Rajil1213: Yes. This TODO is about updating the value of the constant once it is decided. Right now, it is at 30 but we might change that before the deployment.

crates/tx-graph/src/peg_out_graph.rs
Rajil1213: Sure.

Rajil1213: Done: https://alpenlabs.atlassian.net/browse/STR-1150

storopoli:
We can merge if you think this is ready. Feel free to hit the green merge button

crates/tx-graph/src/peg_out_graph.rs
Rajil1213: Created one for this too: https://alpenlabs.atlassian.net/browse/STR-1151

Rajil1213:
@storopoli  feel free to merge if you are okay with the replies.

PR #57, opened 2025-03-10
add secret service dockerfile

Zk2u:
## Description

Adds a dockerfile for the secret service

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,35 @@
+# Ignore git repository files
+.git
+.gitignore
+
+# Ignore GitHub workflow files
+.github/
+
+# Ignore Rust target directory (build artifacts)
+target/
+
+# Ignore Node.js and other dependency directories if present
+node_modules/
+bower_components/
+
+# Ignore logs
+logs/
+*.log
+
+# Ignore temporary files
+*.tmp
+*.temp
+*.swp
+*.swo
+
+# Ignore utilities that are not necessary for the build
+# utils/html-solver/
+
+# Ignore any local configuration or environment files
+*.env
+*.local
+
+# Ignore editor or IDE specific files
+*.idea/
+*.vscode/
+*.DS_Store
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,7 +14,7 @@ bower_components/
 
 # Ignore logs
 logs/
-*.log
+*.log*
 
 # Ignore temporary files
 *.tmp
@@ -33,3 +33,8 @@ logs/
 *.idea/
 *.vscode/
 *.DS_Store
+
+test-data/
+assets/
+*.bak
+seed
--- a/bin/secret-service/Dockerfile
+++ b/bin/secret-service/Dockerfile
@@ -1,0 +1,49 @@
+# warning: run from repo root, NOT inside secret-service dir
+FROM rust:latest AS builder
+
+WORKDIR /app
+
+# Set environment variables for optimized release builds
+ENV CARGO_INCREMENTAL=0 \
+    CARGO_TERM_COLOR=always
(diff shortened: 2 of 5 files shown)


bin/secret-service/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+# Set environment variables for optimized release builds
+ENV CARGO_INCREMENTAL=0 \
+    CARGO_TERM_COLOR=always
+# TARGET=x86_64-unknown-linux-musl
+
+# Install system dependencies
+RUN apt-get update && apt-get -y upgrade && apt-get install -y \
+    pkg-config \
+    build-essential
```
storopoli, line 14: Don't you need to do the
```
apt-get clean && \
rm -rf /var/lib/apt/lists/*
```
here as well?

storopoli (dismissed):
ACK ee39105f13324af397435047eb2bc68e5a764548

Looks good to me.

But wait for @purusang.

.dockerignore
```diff
@@ -0,0 +1,35 @@
...
+# utils/html-solver/
+
+# Ignore any local configuration or environment files
+*.env
+*.local
+
+# Ignore editor or IDE specific files
+*.idea/
+*.vscode/
+*.DS_Store
```
Rajil1213, line 35: Some other files we want to ignore:

```gitignore
test-data/
assets/
*.log*
*.bak

**Review by Rajil1213 — CHANGES_REQUESTED** [review]
In terms of file organization, we should create a separate `build` or `docker` directory to house all the different Dockerfiles similar to how we do it in strata. So the file structure would be:

`docker/secret-service/Dockerfile`

We should also create a `docker-compose` file which will only have the secret service for now but we'll add the bridge node to it later that connects to the secret service via its docker assigned network interface.

File: compose.yml [hunk 3]
```diff
@@ -0,0 +1,14 @@
+services:
+  secret-service:
+    image: secret-service:latest
+    build:
+      context: .
+      dockerfile: ./docker/secret-service/Dockerfile
+    environment:
+      - SECRET_SERVICE_DEV=1
+    ports:
+      - 9898:69/udp
```
Rajil1213, line 10: This port shouldn't be exposed to the host.

docker/secret-service/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN apt-get update && \
+    apt-get install -y \
+    curl && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Copy the built binaries from the builder stage
+COPY --from=builder /app/secret-service /usr/local/bin/secret-service
+
+EXPOSE 3000
```
Rajil1213, line 47: Probably should expose the same port as the one in `config.toml` just for the sake of consistency.

docker/vol/secret-service/config.toml
```diff
@@ -0,0 +1,10 @@
+[tls]
+# key = "/app/tls/key.der"
+# cert = "/app/tls/cert.der"
+# ca = "/app/tls/bridge.ca.der"
+
+[transport]
+addr = "127.0.0.1:69"
+
+seed = "/app/seed/seed"
+network = "signet"
```
Rajil1213, line 10: In practice, the compose file will only be used when running the system locally since there'll be a separate k8s file that we'll use in prod. So, this should be `regtest` (unless there's some other reason to not do this).

Rajil1213 (changes requested):
Thanks for the quick turnaround. Left a few nits but this looks good.

Rajil1213 (approved):

PR #59, opened 2025-03-11
alpen-bridge: add p2p integration

storopoli:
## Description

Adds the P2P integration to the bridge node.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

It needs the `SecretKey` using the P2P derivation path from the Secret Service.

## Checklist

## Related Issues

The same as the target branch, if I say it Jira will close the issue 😂


bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -34,8 +37,21 @@ fn init_secret_service_client(_config: &Config) -> SecretServiceClient {
...
+async fn init_p2p_msg_handler(config: &Config, sk: SecretKey) -> anyhow::Result<MessageHandler> {
+    let config = strata_bridge_p2p_service::Configuration::new_with_secret_key(
+        sk,
+        config.p2p.idle_connection_timeout,
+        config.p2p.listening_addr.clone(),
+        config.p2p.allowlist.clone(),
+        config.p2p.connect_to.clone(),
+        config.p2p.signers_allowlist.clone(),
+        config.p2p.num_threads,
+    );
```
Rajil1213, line 52: Does this work?

```suggestion
    let P2PConfig { idle_connection_timeout, listening_addr, allowlist, connect_to, signers_allowlist, num_threads } = config; 
   
    let config = strata_bridge_p2p_service::Configuration::new_with_secret_key(
        sk,
        idle_connection_timeout,
        listening_addr,
        allowlist,
        connect_to,
        signers_allowlist,
        num_threads,
    );
```

Rajil1213 (commented):
A minor nit but otherwise, looks good to go.

bin/alpen-bridge/src/mode/operator.rs
storopoli: Much better thanks!
Done in ffb71d14495a913f1cb1629fe0b073c2c0868a82

Rajil1213: I did a 1:1 mapping of all non-secret config fields from the `strata_bridge_p2p_service::Configuration` to `alpen_bridge::config::P2PConfig` but since we can derive the `PeerID`'s from the p2p pubkey, do you think we can remove the `allowlist` from the `P2PConfig` and instead pass in the p2p pubkeys for all operators and derive the peer IDs from there?

storopoli: You need this https://docs.rs/libp2p-identity/0.2.10/libp2p_identity/secp256k1/struct.PublicKey.html then you can into it to this https://docs.rs/libp2p-identity/0.2.10/libp2p_identity/struct.PublicKey.html and then you can into `PeerId`.

storopoli: `allowlist` is the list that we allow to connect to.
`signers_allowlist` is the list of authenticated messages that we've accept.
From `strata-p2p`:
```rs
    /// List of [`PeerId`]s that the node is allowed to connect to.
    pub allowlist: Vec<PeerId>,

    /// Initial list of nodes to connect to at startup.
    pub connect_to: Vec<Multiaddr>,

    /// List of signers' P2P public keys, whose messages the node is allowed to accept.
    pub signers_allowlist: Vec<P2POperatorPubKey>,
```

storopoli: We should populate both.

Rajil1213: Assuming that we have three bridge nodes each with their own p2p keys, how do we populate these fields? The p2p keys can be converted to bytes and those will go into the `signers_allowlist`. What would go into the `allowlist`?

Rajil1213: Wait, I think I get it -- the `signers_allowlist` is for incoming messages and `allowlist` is for outgoing ones?

storopoli: Exactly.

storopoli: Suppose you have 3 nodes, A, B, and C.

```rs
// A
allowlist_A: Vec<PeerId> = vec![B, C];
signers_allowlist_A: Vec<P2POperatorPubKey> = vec![B, C];
connect_to_A: Vec<Multiaddr> = vec![B, C];
```

storopoli: Otherwise they won't connect to each other and we will need to send a connect to peer command with the P2PHandle

Rajil1213: Okay then, I think we can derive both the `allowlist` and `signers_allowlist` from just the p2p keys from the params.

storopoli: Rebased and derived `allowlist` and `signers_allowlist`.

bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -14,7 +24,9 @@ pub(crate) async fn bootstrap(params: Params, config: Config) -> anyhow::Result<
 
     let s2_client = init_secret_service_client(&config);
 
-    let message_handler = init_p2p_msg_handler(&config);
+    // TODO(@Zk2u!): give the `init_p2p_message_handler` the P2P secret key `sk`.
+    let sk = todo!();
```
Rajil1213, line 28: Adding this `todo!` macro will probably fail clippy.

bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -34,8 +46,52 @@ fn init_secret_service_client(_config: &Config) -> SecretServiceClient {
...
+        .collect();
+    let signers_allowlist: Vec<P2POperatorPubKey> =
+        other_operators.into_iter().map(Into::into).collect();
+
+    let P2PConfig {
+        idle_connection_timeout,
+        listening_addr,
+        connect_to,
+        num_threads,
+    } = config.p2p.clone();
```
Rajil1213, line 82: Do we really need clone here?

Rajil1213 (changes requested):

bin/alpen-bridge/src/mode/operator.rs
Rajil1213: I'd recommend adding a `get_key` function that accepts the `s2_client` and then retrieves all the relevant keys (musig2, p2p, operator funds). The body of this function could have an `unimplemented!()`.

bin/alpen-bridge/src/mode/operator.rs
storopoli: Either we clone this here or we clone above: https://github.com/alpenlabs/strata-bridge/blob/f84e8be2f8d886d91689c478ea5684a1f554dd75/bin/alpen-bridge/src/mode/operator.rs#L86-L95
`P2PConfig` is not `Clone` and neither is `Multiaddr`: https://github.com/alpenlabs/strata-bridge/blob/f84e8be2f8d886d91689c478ea5684a1f554dd75/bin/alpen-bridge/src/config.rs#L76-L93

bin/alpen-bridge/src/mode/operator.rs
storopoli: Done

Rajil1213 (approved):

PR #61, opened 2025-03-12
alpen-bridge: RPC server

storopoli:
## Description

Implements the RPC server in `alpen-bridge`.
Also closes https://github.com/alpenlabs/strata-bridge/security/dependabot/10.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This refactors the RPC traits to use existing or new types in the `crates/primitives/src/duties.rs`.
This makes more sense since the RPC will query the bridge database and the database will be populated by the duty tracker & Co.

Added a bunch of traits in `crates/db/src/tracker.rs` that we need to implement to have the RPC working.

## Checklist

## Related Issues

STR-995


crates/db/src/inmemory/public.rs
```diff
@@ -269,6 +269,16 @@ impl PublicDb for PublicDbInMemory {
...
+    async fn get_deposits(&self) -> DbResult<Vec<Txid>> {
+        Ok(self
+            .claim_txid_to_operator_index_and_deposit_txid
+            .read()
+            .await
+            .keys()
+            .cloned()
+            .collect())
+    }
+
```
storopoli, line 281: Don't know if this is right, the table name was very confusing.

crates/db/src/persistent/sqlite.rs
```diff
@@ -485,6 +491,23 @@ impl PublicDb for SqliteDb {
...
+            )
+            .fetch_all(&self.pool)
+            .await
+            .map_err(StorageError::from)?
+            .into_iter()
+            .map(|row| *row.deposit_txid)
+            .collect())
+        })
+        .await
+    }
```
storopoli, line 509: Don't know if this is right, the table name was very confusing.

Rajil1213 (commented):
This is a great start especially considering that a lot of pieces required to build this out completely is still in development. I'm going to work on the db refactors first so that it unblocks stuff for the RPC. 

As for the decision to use the types in the primitives, that's fine. There might be some types that don't exactly fit there and those can go into `rpc/types.rs` (which I see that you've already done!).

crates/db/src/persistent/sqlite.rs
Rajil1213: This comment was resolved but the right table to use is the `deposits` table.

storopoli:
Okay (from mobile I'm at the dentist). I'll work in the new `strata-rpc` "ping" command. I think this is ready to merge since all the ROC stuff depends on new database trait implementations.

Rajil1213 (approved):

PR #62, opened 2025-03-12
Operator wallet

Zk2u:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ btc-notify = { path = "crates/btc-notify" }
 prover-test-utils = { path = "crates/bridge-proof/test-utils" }
 secret-service-client = { path = "crates/secret-service-client" }
 secret-service-proto = { path = "crates/secret-service-proto" }
+secret-service-server = { path = "crates/secret-service-server" }
 strata-bridge-agent = { path = "crates/agent" }
 strata-bridge-connectors = { path = "crates/connectors" }
 strata-bridge-db = { path = "crates/db" }
@@ -61,8 +62,12 @@ strata-bridge-rpc = { path = "crates/rpc" }
 strata-bridge-stake-chain = { path = "crates/stake-chain" }
 strata-bridge-test-utils = { path = "crates/test-utils" }
 strata-bridge-tx-graph = { path = "crates/tx-graph" }
+
+# deps from original strata repo
+strata-bridge-tx-builder = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
 strata-btcio = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
 strata-common = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
+strata-config = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
 strata-crypto = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
 strata-l1tx = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
 strata-primitives = { git = "https://github.com/alpenlabs/strata.git", rev = "c5a46a9e80ee12c25868a5276dd13b56fe0c4e7c" }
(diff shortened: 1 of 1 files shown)


crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+    descriptor,
+    error::CreateTxError,
+    KeychainKind, LocalOutput, Wallet,
+};
+use sync::{Backend, SyncError};
+use tracing::{debug, info};
+
+/// Config for [OperatorWallet]
+#[derive(Debug)]
+pub struct OperatorWalletConfig {
```
Rajil1213, line 15: This needs a constructor since all the values are private. Might be prudent to check that the `cpfp_value` and `s_value`'s are different in the constructor.

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// (excluding CPFP outputs). Needs signing by the general wallet.
+    pub fn refill_claim_funding_utxos(&mut self, fee_rate: FeeRate) -> Result<Psbt, CreateTxError> {
+        let cpfp_utxos = self
+            .cpfp_utxos()
+            .into_iter()
+            .map(|lo| lo.outpoint)
+            .collect();
+        let mut tx_builder = self.general_wallet.build_tx();
+        // DON'T spend any of the cpfp outputs
+        tx_builder.unspendable(cpfp_utxos);
```
Rajil1213, line 122: This should not spend any of the previous funding utxos either (if still present in the wallet but reserved). A hacky approach is to filter out all UTXOs that have either 33K sats or 330 sats.

Rajil1213 (dismissed):
This looks great overall. Left a nit and a comment regarding refilling the funding utxos.

I also see that this PR is built on top of the other PR that you have open regarding the wallet separation. So, I haven't reviewed those part here.

Cargo.toml
```diff
@@ -28,7 +28,7 @@ members = [
   "bin/alpen-bridge",
 
   # test utilities
-  "crates/test-utils",
+  "crates/test-utils", "crates/operator-wallet",
```
storopoli, line 31: nit, should be arranged above

Cargo.toml
```diff
@@ -88,15 +83,10 @@ zkaleido-sp1-adapter = { git = "https://github.com/alpenlabs/zkaleido", tag = "v
...
 arbitrary = { version = "1.4.1", features = ["derive"] }
 ark-bn254 = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
-ark-crypto-primitives = { git = "https://github.com/arkworks-rs/crypto-primitives/" }
 ark-ec = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
 ark-ff = { git = "https://github.com/chainwayxyz/algebra/", branch = "new-ate-loop" }
 ark-groth16 = { git = "https://github.com/arkworks-rs/groth16" }
-ark-r1cs-std = { git = "https://github.com/arkworks-rs/r1cs-std/" }
-ark-relations = { git = "https://github.com/arkworks-rs/snark/" }
-ark-std = { git = "https://github.com/arkworks-rs/std/", rev = "db4367e68ff60da31ac759831e38f60171f4e03d" }
 async-trait = "0.1.81"
```
storopoli, line 89: Why are we removing these in the "Operator Wallet" PR?
This does not make sense

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+    error::CreateTxError,
+    KeychainKind, LocalOutput, Wallet,
+};
+use sync::{Backend, SyncError};
+use tracing::{debug, info};
+
+/// Config for [OperatorWallet]
+#[derive(Debug)]
+pub struct OperatorWalletConfig {
+    /// Value of the funding UTXO for stakes. Not `s`.
```
storopoli, line 16: What is `s`? You mean the connector?

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+    stake_funding_utxo_value: Amount,
+    /// Number of UTXOs of amount claim_funding_utxo_value sat to keep on hand at minimum
+    stake_funding_utxo_pool_size: usize,
+    /// Value of CPFP UTXOs to identify them
+    cpfp_value: Amount,
+    /// Value of `s`, the stake amount, to identify the UTXO
```
storopoli, line 22: connector?

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+    s_value: Amount,
+}
+
+impl OperatorWalletConfig {
+    /// Creates a new [OperatorWalletConfig]. Panics if cpfp_value == s_value
```
storopoli, line 27:
```suggestion
    /// Creates a new [`OperatorWalletConfig`].
    ///
    /// # Panics
    ///
    /// Panics if `cpfp_value == s_value`
```

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        Self {
+            stake_funding_utxo_value,
+            stake_funding_utxo_pool_size,
+            cpfp_value,
+            s_value,
+        }
+    }
+}
+
+const NETWORK: Network = Network::Signet;
```
storopoli, line 47: This should use something configurable, either a `Params` or a `const` in `bridge_primitives`.

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// The [OperatorWallet] is responsible for
```
storopoli, line 49: Finish this...

```suggestion
/// The [`OperatorWallet`] is responsible for
```

crates/operator-wallet/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+    general_wallet: Wallet,
+    stakechain_wallet: Wallet,
+    config: OperatorWalletConfig,
+    stakechain_addr_script_buf: ScriptBuf,
+    general_addr_script_buf: ScriptBuf,
+    sync_backend: Backend,
+}
+
+impl OperatorWallet {
+    /// Creates a new [OperatorWallet]
```
storopoli, line 61:
```suggestion
    /// Creates a new [`OperatorWallet`].
```

storopoli (changes requested):

crates/operator-wallet/src/lib.rs
Rajil1213: Maybe should say "stake" explicitly.

Cargo.toml
Rajil1213: Yeah these should not be removed. We'll remove these once we start using the main branch of bitvm.

crates/operator-wallet/src/lib.rs
Rajil1213: Should be read from a file as a param. Ideally, we'd get this value dynamically by calling the bitcoin node rpc so that there's never any discrepancy. For now, using the value from a params file should be fine. It's probably not there right now so feel free to add it to Params struct in `bin/alpen-bridge/src/params.rs` at the top level.

Rajil1213 (approved):

PR #64, opened 2025-03-13
feat!: DRT with X-Only PK metadata

storopoli:
## Description

Deposit Request Transactions (DRT) now need X-Only PK of recovery address instead of take back leaf hash.
This fixes the DRT metadata OP_RETURN construction/parsing.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

The new `OP_RETURN` now is:

1. magic bytes
2. Second Element:
  a. DRT: 32-byte X-only PK for the recovery P2TR address
  b. DT: 4-byte big-endian `u32` stake index
4. 20-byte EL address

This PR needs polishing. It is not ready, but the main ideas are here.

## Checklist

## Related Issues

STR-1776

### DIFF
--- a/crates/duty-tracker/src/predicates.rs
+++ b/crates/duty-tracker/src/predicates.rs
@@ -1,11 +1,12 @@
 //! This module supplies helpers for categorizing transactions and extracting payloads from them
 //! where relevant.
+
 use std::sync::Arc;
 
 use alpen_bridge_params::{prelude::PegOutGraphParams, sidesystem::SideSystemParams};
 use bitcoin::{
-    hashes::Hash, opcodes::all::OP_RETURN, taproot::TAPROOT_CONTROL_NODE_SIZE, OutPoint, Script,
-    TapNodeHash, Transaction, Txid,
+    key::constants::SCHNORR_PUBLIC_KEY_SIZE, opcodes::all::OP_RETURN, OutPoint, Script,
+    Transaction, Txid, XOnlyPublicKey,
 };
 use btc_notify::client::TxPredicate;
 use strata_bridge_primitives::deposit::DepositInfo;
@@ -50,15 +51,18 @@ pub(crate) fn deposit_request_info(
     // TODO(proofofkeags): validate that the script_pubkey pays to the right operator set
 
     let ee_address_size = sidesystem_params.ee_addr_size;
-    let (take_back_leaf_hash, el_addr) =
+    let (recovery_x_only_pk, el_addr) =
         magic_tagged_data(MAGIC_BYTES, &tx.output.get(1)?.script_pubkey).and_then(|meta| {
-            if meta.len() != TAPROOT_CONTROL_NODE_SIZE + ee_address_size {
+            if meta.len() != SCHNORR_PUBLIC_KEY_SIZE + ee_address_size {
                 return None;
             }
-            let take_back_leaf_hash = meta.get(..TAPROOT_CONTROL_NODE_SIZE)?;
+            let recovery_x_only_pk = meta.get(..SCHNORR_PUBLIC_KEY_SIZE)?;
+            // TODO: handle error variant and get rid of expect.
+            let recovery_x_only_pk = XOnlyPublicKey::from_slice(recovery_x_only_pk)
+                .expect("Failed to parse XOnlyPublicKey");
             let el_addr =
-                meta.get(TAPROOT_CONTROL_NODE_SIZE..TAPROOT_CONTROL_NODE_SIZE + ee_address_size)?;
-            Some((take_back_leaf_hash, el_addr))
+                meta.get(SCHNORR_PUBLIC_KEY_SIZE..SCHNORR_PUBLIC_KEY_SIZE + ee_address_size)?;
+            Some((recovery_x_only_pk, el_addr))
         })?;
 
     Some(DepositInfo::new(
@@ -66,7 +70,7 @@ pub(crate) fn deposit_request_info(
         stake_index,
         el_addr.to_vec(),
         deposit_request_output.value,
-        TapNodeHash::from_slice(take_back_leaf_hash).unwrap(),
+        recovery_x_only_pk,
         deposit_request_output.script_pubkey.clone(),
     ))
 }
--- a/crates/primitives/src/deposit.rs
+++ b/crates/primitives/src/deposit.rs
@@ -3,11 +3,12 @@
 //! Contains types, traits and implementations related to creating various transactions used in the
 //! bridge-in dataflow.
 
+use alpen_bridge_params::{prelude::PegOutGraphParams, sidesystem::SideSystemParams};
 use bitcoin::{
     key::TapTweak,
     secp256k1::SECP256K1,
-    taproot::{self, ControlBlock},
-    Address, Amount, OutPoint, Psbt, ScriptBuf, TapNodeHash, Transaction, TxOut,
+    taproot::{self, ControlBlock, LeafVersion},
+    Amount, OutPoint, Psbt, ScriptBuf, TapNodeHash, Transaction, TxOut, XOnlyPublicKey,
 };
 use serde::{Deserialize, Serialize};
 
@@ -39,9 +40,9 @@ pub struct DepositInfo {
     /// The stake index must be encoded in 4-byte big-endian.
     stake_index: u32,
 
-    /// The execution layer address to mint the equivalent tokens to.
+    /// The execution environment address to mint the equivalent tokens to.
     /// As of now, this is just the 20-byte EVM address.
-    el_address: Vec<u8>,
+    ee_address: Vec<u8>,
 
     /// The amount in bitcoins that the user is sending.
     ///
@@ -49,9 +50,9 @@ pub struct DepositInfo {
     /// confirmed on bitcoin. The excess amount is used as miner fees for the Deposit Transaction.
     total_amount: Amount,
 
-    /// The hash of the take back leaf in the Deposit Request Transaction (DRT) as provided by the
+    /// The [`XOnlyPublicKey`] in the Deposit Request Transaction (DRT) as provided by the
     /// user in their `OP_RETURN` output.
-    take_back_leaf_hash: TapNodeHash,
+    x_only_public_key: XOnlyPublicKey,
 
     /// The original script_pubkey in the Deposit Request Transaction (DRT) output used to sanity
     /// check computation internally i.e., whether the known information (n/n script spend path,
@@ -60,6 +61,7 @@ pub struct DepositInfo {
     original_script_pubkey: ScriptBuf,
 }
 
+// FIXME: remove the trait.
 impl TxKind for DepositInfo {
     fn construct_signing_data<C: BuildContext>(
         &self,
@@ -73,6 +75,7 @@ impl TxKind for DepositInfo {
             build_context,
             deposit_amt,
             tag.expect("deposit tx must have a tag in the metadata"),
+            20,
         )?;
 
         let mut psbt = Psbt::from_unsigned_tx(unsigned_tx)?;
@@ -96,15 +99,15 @@ impl DepositInfo {
         stake_index: u32,
         el_address: Vec<u8>,
         total_amount: Amount,
-        take_back_leaf_hash: TapNodeHash,
+        x_only_public_key: XOnlyPublicKey,
         original_script_pubkey: ScriptBuf,
     ) -> Self {
         Self {
             deposit_request_outpoint,
             stake_index,
-            el_address,
+            ee_address: el_address,
             total_amount,
-            take_back_leaf_hash,
+            x_only_public_key,
             original_script_pubkey,
         }
     }
@@ -121,7 +124,7 @@ impl DepositInfo {
 
     /// Get the address in EL to mint tokens to.
     pub fn el_address(&self) -> &[u8] {
-        &self.el_address
+        &self.ee_address
     }
 
     /// Get the outpoint of the Deposit Request Transaction (DRT) that is to spent in the Deposit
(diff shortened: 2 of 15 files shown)


crates/duty-tracker/src/predicates.rs
```diff
@@ -43,30 +43,34 @@ pub(crate) fn deposit_request_info(
     pegout_graph_params: &PegOutGraphParams,
     stake_index: u32,
 ) -> Option<DepositInfo> {
+    const X_ONLY_PUBKEY_SIZE: usize = 32;
```
Rajil1213, line 46: There is a `bitcoin::key::constants::SCHNORR_PUBLIC_KEY_SIZE`.

crates/duty-tracker/src/predicates.rs
```diff
@@ -43,30 +43,34 @@ pub(crate) fn deposit_request_info(
...
     if deposit_request_output.value <= pegout_graph_params.deposit_amount {
         return None;
     }
     // TODO(proofofkeags): validate that the script_pubkey pays to the right operator set
 
     let ee_address_size = sidesystem_params.ee_addr_size;
-    let (take_back_leaf_hash, el_addr) =
+    let (recovery_x_only_pk, el_addr) =
         magic_tagged_data(MAGIC_BYTES, &tx.output.get(1)?.script_pubkey).and_then(|meta| {
             if meta.len() != TAPROOT_CONTROL_NODE_SIZE + ee_address_size {
```
Rajil1213, line 57: Since we are no longer including the tapleaf hash, we should update this condition too.
```suggestion
            if meta.len() != SCHNORR_PUBLIC_KEY_SIZE + ee_address_size {
```

crates/primitives/src/deposit.rs
```diff
@@ -142,12 +142,17 @@ impl DepositInfo {
...
         let spend_script = n_of_n_script(&build_context.aggregated_pubkey()).compile();
-        let spend_script_hash =
-            TapNodeHash::from_script(&spend_script, taproot::LeafVersion::TapScript);
+        let spend_script_hash = TapNodeHash::from_script(&spend_script, LeafVersion::TapScript);
 
-        let takeback_script_hash = self.take_back_leaf_hash();
+        // Compute the merkle root using the x-only public key in the OP_RETURN
+        let recovery_xonly_pubkey = self.x_only_public_key();
+        // TODO: how to get the PegOutGraphParams.recover_delay (from Params) here?
+        let refund_delay = 1_008; // Placeholder
```
Rajil1213, line 150: I think it's time to scrap the `TxKind` trait directly and have a method on the `DepositInfo` type that accepts the `buidl_context` and the `PegOutGraphParams`. We'll need to move the `tag` to the `PegOutGraphParams` as well. For the present, we can keep the `TxKind` trait around while the duty tracker is still in development.

crates/primitives/src/deposit.rs
```diff
@@ -170,7 +175,7 @@ impl DepositInfo {
         let control_block = ControlBlock {
             leaf_version: taproot::LeafVersion::TapScript,
             internal_key: *UNSPENDABLE_INTERNAL_KEY,
-            merkle_branch: vec![*takeback_script_hash].try_into().map_err(|_| {
+            merkle_branch: vec![takeback_script_hash].try_into().map_err(|_| {
```
Rajil1213, line 178: Since we're the ones constructing the take back script, it should be fine to just use `expect` here.

crates/primitives/src/scripts/general.rs
```diff
@@ -20,6 +25,21 @@ pub fn n_of_n_script(aggregated_pubkey: &XOnlyPublicKey) -> Script {
     }
 }
 
+/// Creates a "take back" script that is used to validate a deposit request transaction (DRT) given
+/// a user's public key.
+///
+/// The `refund_delay` should be provided by the the consensus-critical parameters that dictate the
```
Rajil1213, line 31:
```suggestion
/// The `refund_delay` should be provided by the consensus-critical parameters that dictate the
```

crates/primitives/src/scripts/metadata.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// This is a 4-byte big-endian encoded unsigned 32-bit integer.
+    StakeIndex([u8; 4]),
+
+    /// 32-bit X-only public key.
+    // TODO: make this a BOSD Descriptor.
+    XOnlyPublicKey([u8; 32]),
+
+    /// 20-byte Execution layer address.
+    ExecutionLayerAddress([u8; 20]),
+}
```
Rajil1213, line 23: I usually treat enum variants as representing a `OneOf` type meaning that the consumer only uses one of the variants at any given time. That is not the use case for this enum since one or more of these variants are used when constructing the `OP_RETURN` data in the `DT` or `DRT`.

I'd refactor to something like the following:

```rust
pub struct AuxillaryData<'tag> {
     tag: &'tag [u8],
     
     metadata: DepositMetadata;
}

pub enum DepositMetadata {
    DepositRequestTx {
        takeback_pubkey: XOnlyPublicKey,
        ee_address: Vec<u8>, // can be any length but restricted by ee_addr_size param
   },
   DepositTx {
        ee_addr_size: Vec<u8>,
        stake_index: u32,
   }
}
```

crates/primitives/src/test_utils.rs
```diff
@@ -50,38 +53,33 @@ pub(crate) fn generate_pubkey_table(table: &[PublicKey]) -> PublickeyTable {
     PublickeyTable::from(pubkey_table)
 }
 
+pub(crate) fn generate_xonly_pubkey() -> XOnlyPublicKey {
+    let mut rng = thread_rng();
+    let mut sk = SecretKey::new(&mut rng);
+    // negate sk if odd
+    if sk.x_only_public_key(SECP256K1).1 == Parity::Odd {
```
Rajil1213, line 60: There should be an `EvenSecretKey` type, I think.

crates/primitives/src/test_utils.rs
```diff
@@ -50,38 +53,33 @@ pub(crate) fn generate_pubkey_table(table: &[PublicKey]) -> PublickeyTable {
...
-        .add_leaf(1, n_of_n_spend_script.compile())
-        .unwrap()
-        .add_leaf(1, op_return_script)
-        .unwrap();
-
-    let spend_info = taproot_builder
-        .finalize(SECP256K1, *UNSPENDABLE_INTERNAL_KEY)
-        .unwrap();
+    let recovery_xonly_pubkey = generate_xonly_pubkey();
+    let refund_delay = 1_008; // Placeholder
```
Rajil1213, line 70: This is fine for testing. But you could use the params too.

Rajil1213 (changes requested):
Left a few nits but this mostly looks good.

crates/primitives/src/scripts/metadata.rs
Rajil1213: I also think they're calling it the execution _environment_ instead of _layer_.

crates/primitives/src/deposit.rs
storopoli: I prefer to use the `map_err` than `expect`

crates/primitives/src/test_utils.rs
storopoli: To be fair I much prefer to hardcode for now. Will move to a test `const`.
This is ludicrous:
image

crates/duty-tracker/src/predicates.rs
storopoli: Done in 42dbcc2dfec1013af435fdbd9b2936af2fed010d

Thanks!

crates/duty-tracker/src/predicates.rs
storopoli: Done in 42dbcc2dfec1013af435fdbd9b2936af2fed010d

Thanks!

crates/primitives/src/scripts/general.rs
storopoli: Done in 42dbcc2dfec1013af435fdbd9b2936af2fed010d

Thanks!

crates/primitives/src/scripts/metadata.rs
storopoli: Done in 42dbcc2dfec1013af435fdbd9b2936af2fed010d

Thanks!

crates/primitives/src/test_utils.rs
storopoli: Agreed, I think @Zk2u has something like this somewhere. But out of the scope of the PR.

crates/duty-tracker/src/predicates.rs
Rajil1213: This constant can be removed now.

crates/duty-tracker/src/predicates.rs
```diff
@@ -43,30 +44,34 @@ pub(crate) fn deposit_request_info(
...
-            let take_back_leaf_hash = meta.get(..TAPROOT_CONTROL_NODE_SIZE)?;
-            let el_addr =
-                meta.get(TAPROOT_CONTROL_NODE_SIZE..TAPROOT_CONTROL_NODE_SIZE + ee_address_size)?;
-            Some((take_back_leaf_hash, el_addr))
+            let recovery_x_only_pk = meta.get(..X_ONLY_PUBKEY_SIZE)?;
+            // TODO: handle error variant and get rid of expect.
+            let recovery_x_only_pk = XOnlyPublicKey::from_slice(recovery_x_only_pk)
+                .expect("Failed to parse XOnlyPublicKey");
+            let el_addr = meta.get(X_ONLY_PUBKEY_SIZE..X_ONLY_PUBKEY_SIZE + ee_address_size)?;
+            Some((recovery_x_only_pk, el_addr))
```
Rajil1213, line 66: The `X_ONLY_PUBKEY_SIZE` const should be replaced by the `SCHNORR_PUBLIC_KEY_SIZE`.

crates/primitives/src/deposit.rs
Rajil1213: Incidentally, this is failing a test.

crates/primitives/src/test_utils.rs
Rajil1213: It's already there, just need to use it here:

```suggestion
    let even_sk: EvenSecretKey = sk.into();
```

Rajil1213 (changes requested):

crates/primitives/src/scripts/metadata.rs
```diff
@@ -0,0 +1,81 @@
+//! Primitives for Bridge metadata.
+
+use std::mem;
+
+use bitcoin::{key::constants::SCHNORR_PUBLIC_KEY_SIZE, XOnlyPublicKey};
+
+/// Execution Environment address size in bytes.
+const EE_ADDRESS_SIZE: usize = 20;
```
Rajil1213, line 8: This should be a param, not a const.

crates/primitives/src/scripts/metadata.rs
```diff
@@ -0,0 +1,81 @@
...
+    },
+}
+
+impl<'tag> AuxiliaryData<'tag> {
+    /// Extracts the metadata as bytes.
+    pub fn to_vec(&'tag self) -> Vec<u8> {
+        let mut bytes = Vec::with_capacity(
+            self.tag.len() // FIXME: make this known at compile time once the tag length is defined.
+                + match &self.metadata {
+                    DepositMetadata::DepositRequestTx { .. } => SCHNORR_PUBLIC_KEY_SIZE + EE_ADDRESS_SIZE,
```
Rajil1213, line 55: You should get the size from the param.
```suggestion
                    DepositMetadata::DepositRequestTx { ee_address, .. } => SCHNORR_PUBLIC_KEY_SIZE + ee_address.len(),
```

crates/primitives/src/scripts/metadata.rs
```diff
@@ -0,0 +1,81 @@
...
+                    DepositMetadata::DepositTx { .. } => U32_SIZE + EE_ADDRESS_SIZE,
```
Rajil1213, line 56: Same here.
```suggestion
                    DepositMetadata::DepositTx { ee_address, .. } => U32_SIZE + ee_address.len(),
```

crates/primitives/src/scripts/metadata.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use std::mem;
+
+use bitcoin::{key::constants::SCHNORR_PUBLIC_KEY_SIZE, XOnlyPublicKey};
+
+/// Execution Environment address size in bytes.
+const EE_ADDRESS_SIZE: usize = 20;
+
+/// u32 size in bytes.
+const U32_SIZE: usize = mem::size_of::<u32>();
```
Rajil1213, line 11: We should call this the `STAKE_INDEX_SIZE` to make the usage explicit.

Rajil1213 (changes requested):
Okay, some final nits and this should be ready to go.

crates/primitives/src/scripts/metadata.rs
storopoli: We can't it's now known at compile time. So we don't do `Vec::with_capacity()` and we do `Vec::new()`.

storopoli: Done in 7493aa70b37f98525e3e78e6697daca0ddf4659b

crates/primitives/src/scripts/metadata.rs
storopoli: Done in 7493aa70b37f98525e3e78e6697daca0ddf4659b

crates/primitives/src/scripts/metadata.rs
storopoli: Done in 7493aa70b37f98525e3e78e6697daca0ddf4659b

Rajil1213 (approved):
Looks good!

PR #65, opened 2025-03-13
integrate secret service

Zk2u:
## Description

fills in the todos for the alpen-bridge crate

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Run db migrations
         run: make migrate
 
-      - run: cargo hack check --locked
+      - run: cargo hack check
         env:
           DATABASE_URL: sqlite://./operator.db
           SKIP_GUEST_BUILD: 1
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Run db migrations
         run: make migrate
 
-      - run: cargo clippy --workspace --lib --examples --tests --benches --all-features --all-targets --locked
+      - run: cargo clippy --workspace --lib --examples --tests --benches --all-features --all-targets
         env:
           RUSTFLAGS: -D warnings
           DATABASE_URL: sqlite://./operator.db


bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -53,12 +61,55 @@ pub(crate) async fn bootstrap(params: Params, config: Config) -> anyhow::Result<
...
+    if path.extension().is_some_and(|x| x == "der") {
+        Ok(vec![CertificateDer::from(cert_chain)])
+    } else {
+        rustls_pemfile::certs(&mut &*cert_chain).collect()
+    }
 }
 
-async fn get_p2p_key(_secret_service: &SecretServiceClient) -> anyhow::Result<SecretKey> {
-    unimplemented!("@Zk2u!");
+async fn get_p2p_key(s2_client: &SecretServiceClient) -> anyhow::Result<SecretKey> {
```
Rajil1213, line 107: This can be inlined. This function was created only to appease clippy and so that we can divide the work more clearly.

bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -53,12 +61,55 @@ pub(crate) async fn bootstrap(params: Params, config: Config) -> anyhow::Result<
...
+        .with_root_certificates(root_store)
+        .with_client_auth_cert(certs, key)
+        .expect("good client config");
+
+    let s2_config = secret_service_client::Config {
+        server_addr: config.server_addr.parse().expect("invalid server address"),
+        server_hostname: config.server_hostname.clone(),
+        local_addr: None,
+        tls_config: tls_client_config,
+        timeout: Duration::from_secs(5),
```
Rajil1213, line 90: This should be made configurable. Could you add it to the `SecretServiceConfig`.

Rajil1213 (changes requested):
Some nits but this looks good to go. In fact, I think we are ready to dockerize the bridge node binary once this is done. It will crash when the duty tracker is initialized since that's still unimplemented but it will help us test out whether the bridge node can talk to the secret-service properly before the crash.

Rajil1213 (approved):
All the relevant changes for this PR look good but for some reason `cargo` needs to update the lockfile which it can't because of the `--locked` flag. I've pushed a commit that removes that flag in the CI workflows. If that works, then this should be mergeable.

PR #69, opened 2025-03-25
bridge dockerization

Zk2u:
## Description

This PR adds dockerisation support for `alpen-bridge` and reworks the dockerization of `secret-service` by using prebuilt base and runtime images, making fast recompiles easy.

To try:

```sh
./docker/gen_s2_tls.sh
docker build -f docker/base.Dockerfile . -t bridge-base:latest
docker build -f docker/rt.Dockerfile . -t bridge-rt:latest
docker compose up --build
```

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR also enables keep alives in the `secret-service-client` so the connection between bridge and s2 doesn't time out. It also corrects the logic in the server when an error happens during incoming stream acceptance so it breaks from the loop and closes the connection down instead of continuing to loop.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,24 @@ docker-down: ## docker compose down
 	cd $(DOCKER_DIR) && docker compose down && \
 	rm -rf $(DOCKER_DIR)/$(DOCKER_DATADIR) 2>/dev/null
 
+##@ Docker
+
+.PHONY: build-base
+build-base: ## Builds the base image used to build the binaries
+	docker build -f docker/base.Dockerfile . -t bridge-base:latest
+
+.PHONY: build-rt
+build-rt: ## Builds the runtime image used as the final container
+	docker build -f docker/rt.Dockerfile . -t bridge-rt:latest
+
+.PHONY: build-compose
+build-compose: ## Builds all images in the compose.yml
+	docker compose build
+
+.PHONY: gen-s2-tls
+gen-s2-tls: ## (Re)generates the TLS CAs, certs and keys for S2 and the bridge to connect
+	./docker/gen_s2_tls.sh
+
 ##@ Code Quality
 
 .PHONY: fmt-check-ws
--- a/docker/gen_s2_tls.sh
+++ b/docker/gen_s2_tls.sh
@@ -1,4 +1,4 @@
-# run from the root of the repo!!!
+# run via `make gen-s2-tls` not directly!
 
 S2_TLS_DIR=docker/vol/secret-service/tls
 BRIDGE_TLS_DIR=docker/vol/alpen-bridge/tls


docker/vol/alpen-bridge/config.toml
```diff
@@ -0,0 +1,29 @@
+finality_depth = 6
+datadir = "/app/data"
+rpc_addr = "0.0.0.0:5678"
+is_faulty = false
```
storopoli, line 4: What is this option?
Are we simulating faulty operators?!

docker/vol/alpen-bridge/params.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[tx_graph]
+deposit_amount = 100_000_000
+operator_fee = 1000000
+challenge_cost = 10000000
+refund_delay = 1008
+
+[stake_chain]
+stake_amount      = 100000000
+burn_amount       = 10000000
```
storopoli, line 16: Can we also do thousands `_` separators here as well? I'm really bad at counting zeros.

docker/vol/secret-service/config.toml
```diff
@@ -2,9 +2,9 @@ seed = "/app/seed"
...
-# key = "/app/tls/key.der"
-# cert = "/app/tls/cert.der"
-# ca = "/app/tls/bridge.ca.der"
+key = "/app/tls/key.der"
+cert = "/app/tls/cert.der"
+ca = "/app/tls/bridge.ca.der"
 
 [transport]
-addr = "127.0.0.1:3000"
+addr = "0.0.0.0:69"
```
storopoli, line 10: Why are we using "well-known" ports here?
See: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

storopoli (changes requested):
I'm very confused. This PR has no description.
I am assuming that this is the defacto bridge docker.
If it is, then we need to address these nits.

docker/vol/secret-service/config.toml
storopoli: Yes but some users might not have privileges to run that port range.
Do 6969 if you want funny number but that does not have issues.
See the wikipedia entry for well-known ports.

bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -49,6 +49,9 @@ pub(crate) async fn bootstrap(params: Params, config: Config) -> anyhow::Result<
         .secret_key()
         .await
         .map_err(|e| anyhow!("error while asking for p2p key: {e:?}"))?;
+
+    info!("Retrieved P2P secret key from S2: {sk:?}");
```
storopoli, line 53: This is still logging a secret key.
I don't think we should do this.
Maybe you can log the PK associated with this secret key?

crates/secret-service-client/src/lib.rs
```diff
@@ -93,14 +93,16 @@ impl SecretServiceClient {
         )
         .map_err(OneOf::new)?;
 
+        let mut transport_config = TransportConfig::default();
+        transport_config.keep_alive_interval(Some(Duration::from_secs(25)));
```
storopoli, line 97: move this to a constant with a FIXME to move to configurable configs/params

storopoli (changes requested):
few nits

storopoli (dismissed):
ACK 3f861585bbddd93cdb75ff240c3e92f5e854ae38

Rajil1213 (dismissed):
Haven't actually run the node yet but only reviewed the code.

storopoli (approved):
ACK db322f503050b1ef0e63a7374a02fecd960e5f1e

PR #99, opened 2025-04-07
adds error information to TransitionErr

ProofOfKeags:
self explanatory


Rajil1213 (dismissed):
This gets the job done for the time being but we should still consider converting `TransitionErr` to an enum with proper error variants. I'll fix the codespell issue before merging this in.

Rajil1213 (approved):

PR #102, opened 2025-04-14
STR-1247: fully check the DRT in predicates

storopoli:
## Description

Reconstructs the DRT from the `OP_RETURN` data in order to fully check that the DRT is compliant with the Strata Bridge specs.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

I added some additional arguments to the `deposit_request_info` in `predicates` in order to reconstruct properly the DRT.
Also, I added some quality of life improvements to the `OperatorTable` to make it easier to get the aggregated pubkey and the BTC keys.

## Checklist

## Related Issues

STR-1247

### DIFF
--- a/crates/duty-tracker/src/predicates.rs
+++ b/crates/duty-tracker/src/predicates.rs
@@ -5,12 +5,18 @@ use std::sync::Arc;
 
 use alpen_bridge_params::prelude::PegOutGraphParams;
 use bitcoin::{
-    consensus, key::constants::SCHNORR_PUBLIC_KEY_SIZE, opcodes::all::OP_RETURN, Network, OutPoint,
-    Script, Transaction, Txid, XOnlyPublicKey,
+    consensus, key::constants::SCHNORR_PUBLIC_KEY_SIZE, opcodes::all::OP_RETURN,
+    secp256k1::SECP256K1, taproot::TaprootBuilder, Address, Network, OutPoint, Script, Transaction,
+    Txid, XOnlyPublicKey,
 };
 use bitcoin_bosd::Descriptor;
 use btc_notify::client::TxPredicate;
-use strata_bridge_primitives::{deposit::DepositInfo, types::OperatorIdx};
+use strata_bridge_primitives::{
+    constants::UNSPENDABLE_INTERNAL_KEY,
+    deposit::DepositInfo,
+    scripts::prelude::{n_of_n_script, n_of_n_with_timelock},
+    types::OperatorIdx,
+};
 use strata_l1tx::{envelope::parser::parse_envelope_payloads, filter::TxFilterConfig};
 use strata_primitives::params::RollupParams;
 use strata_state::batch::{Checkpoint, SignedCheckpoint};
@@ -44,13 +50,14 @@ pub(crate) fn deposit_request_info(
     tx: &Transaction,
     sidesystem_params: &RollupParams,
     pegout_graph_params: &PegOutGraphParams,
+    bridge_aggregated_pk: XOnlyPublicKey,
+    network: Network,
     stake_index: u32,
 ) -> Option<DepositInfo> {
     let deposit_request_output = tx.output.first()?;
     if deposit_request_output.value <= pegout_graph_params.deposit_amount {
         return None;
     }
-    // TODO(proofofkeags): validate that the script_pubkey pays to the right operator set
 
     let ee_address_size = sidesystem_params.address_length as usize;
     let tag = pegout_graph_params.tag.as_bytes();
@@ -68,6 +75,29 @@ pub(crate) fn deposit_request_info(
                 meta.get(SCHNORR_PUBLIC_KEY_SIZE..SCHNORR_PUBLIC_KEY_SIZE + ee_address_size)?;
             Some((recovery_x_only_pk, el_addr))
         })?;
+
+    // Regenerate the P2TR address from the OP_RETURN data.
+    let taproot_builder = TaprootBuilder::new()
+        .add_leaf(1, n_of_n_script(&bridge_aggregated_pk).compile())
+        .expect("failed to add bridge multisig script to tree")
+        .add_leaf(
+            1,
+            n_of_n_with_timelock(&recovery_x_only_pk, pegout_graph_params.refund_delay as u32)
+                .compile(),
+        )
+        .expect("failed to add timelock script");
+
+    let taproot_info = taproot_builder
+        .finalize(SECP256K1, *UNSPENDABLE_INTERNAL_KEY)
+        .unwrap();
+    let merkle_root = taproot_info.merkle_root();
+
+    let p2tr_address = Address::p2tr(SECP256K1, *UNSPENDABLE_INTERNAL_KEY, merkle_root, network);
+
+    // The DRT must have the first output address the bridge aggregated P2TR address.
+    if p2tr_address.script_pubkey() != deposit_request_output.script_pubkey {
+        return None;
+    }
 
     Some(DepositInfo::new(
         OutPoint::new(tx.compute_txid(), 0),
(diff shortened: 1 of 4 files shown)


crates/duty-tracker/src/predicates.rs
```diff
@@ -69,6 +76,29 @@ pub(crate) fn deposit_request_info(
...
+        .unwrap();
+    let merkle_root = taproot_info.merkle_root();
+
+    let p2tr_address = Address::p2tr(SECP256K1, *UNSPENDABLE_INTERNAL_KEY, merkle_root, network);
+
+    // The DRT must have the first output address the bridge aggregated P2TR address.
+    if p2tr_address.script_pubkey() != deposit_request_output.script_pubkey {
+        return None;
+    }
+
```
Rajil1213, line 101: This is fine but it duplicates the logic inside here: 
https://github.com/alpenlabs/strata-bridge/blob/3d2b9fe08ffb8cc679275b377aecefdad16bce08/crates/primitives/src/deposit.rs#L176

You could just make that function pub and then invoke it here.

crates/primitives/src/operator_table.rs
```diff
@@ -109,6 +114,12 @@ impl OperatorTable {
         PublickeyTable(self.idx_key.iter().map(|(k, v)| (*k, v.1)).collect())
     }
 
+    pub fn aggregated_btc_key(&self) -> secp256k1::PublicKey {
+        let pks: Vec<secp256k1::PublicKey> = self.btc_keys().into_iter().collect();
+
+        KeyAggContext::new(pks).unwrap().aggregated_pubkey()
+    }
+
```
Rajil1213, line 122: This is available in `operator_table.tx_build_context`. You could just invoke that from this new method for the sake of convenience.

Rajil1213 (changes requested):
Looks good. The only comments are related to code duplication in different parts of the codebase.

crates/primitives/src/operator_table.rs
storopoli: Aren't we getting rid of the `TxBuild` trait?

crates/duty-tracker/src/predicates.rs
storopoli: I need to have a `build_context: &impl BuildContext` as the second argument.
How do I get that from:

```rs
    tx: &Transaction,
    sidesystem_params: &RollupParams,
    pegout_graph_params: &PegOutGraphParams,
    bridge_aggregated_pk: XOnlyPublicKey,
    network: Network,
    stake_index: u32,
```

crates/primitives/src/operator_table.rs
Rajil1213: No, that's the `TxKind` trait.

crates/duty-tracker/src/predicates.rs
Rajil1213: It's part of the operator table. So, `self.cfg.operator_table.tx_build_context`.

storopoli:
@Rajil1213 check the two new commits.

crates/duty-tracker/src/predicates.rs
storopoli: Done in b54e51eacbf74a7b43a819cef3dacd5287ae104b
and made the function `pub` in 91a2d3bbf92f404ad89768d8ba93f141b50baa31

crates/duty-tracker/src/predicates.rs
```diff
@@ -76,37 +72,26 @@ pub(crate) fn deposit_request_info(
...
         deposit_request_output.script_pubkey.clone(),
-    ))
+    );
+
+    // Regenerate the P2TR address from the OP_RETURN data, for now the spend info does all the
+    // necessary validations.
+    deposit_info
+        .compute_spend_infos(build_context)
+        .map_err(|e| {
+            error!("failed to compute spend info: {:?}", e);
```
Rajil1213, line 89: This should just be a `warn`. This only means that a transaction was not a proper DRT. It's not really an error.

crates/primitives/src/deposit.rs
```diff
@@ -147,12 +147,19 @@ impl DepositInfo {
...
+    /// 1. The script pubkey in the DRT matches the one provided in the [`DepositInfo`].
+    /// 1. The x-only public key in the DRT matches the one provided in the [`DepositInfo`].
+    /// 1. The script pubkey in the DRT is a valid taproot script.
+    pub fn compute_spend_infos(
         &self,
         build_context: &impl BuildContext,
         refund_delay: u16,
     ) -> BridgeTxBuilderResult<TaprootWitness> {
-        // The Deposit Request (DT) spends the n-of-n multisig leaf
+        // The Deposit Request (DRT) spends the n-of-n multisig leaf
```
Rajil1213: This should actually be:

```suggestion
        // The Deposit Transaction (DT) spends the n-of-n multisig leaf
```

crates/primitives/src/deposit.rs
```diff
@@ -147,12 +147,19 @@ impl DepositInfo {
...
     }
 
-    fn compute_spend_infos(
+    /// Computes the spend info for the Deposit Request Transaction (DRT).
+    ///
+    /// This also performs the following validations:
+    ///
+    /// 1. The script pubkey in the DRT matches the one provided in the [`DepositInfo`].
+    /// 1. The x-only public key in the DRT matches the one provided in the [`DepositInfo`].
+    /// 1. The script pubkey in the DRT is a valid taproot script.
```
Rajil1213, line 156: This treats the `DepositInfo` and the `DRT` as two separate things but they really are two representations of the same underlying thing. I'd update the doc to something like the following:

```suggestion
    /// 1. The taproot address computed from the x-only public key in the DRT and the bridge multisig address is the same as the output address in the DRT.
    /// 1. The control block computed from the DRT commits to the multisig script that the Deposit Transaction (DT) spends.
```

Rajil1213 (changes requested):
Some final nits.

crates/primitives/src/deposit.rs
storopoli: Yeah I realized that and already did in 77927ea23ba543457a5dd3b835ee536c1061d929 with some crazy rebasing issues...

crates/duty-tracker/src/predicates.rs
storopoli: Makes sense: done in 881b393c501c7dde52f2ef488279c6d119633691

crates/primitives/src/deposit.rs
```diff
@@ -147,12 +147,16 @@ impl DepositInfo {
         })
     }
 
-    fn compute_spend_infos(
+    /// 1. The taproot address computed from the x-only public key in the DRT and the bridge
```
Rajil1213, line 150: This needs a heading:


```suggestion
    /// Computes the witness data for spending the Deposit Request Transaction (DRT) and in the process also validates the following:

    /// 1. The taproot address computed from the x-only public key in the DRT and the bridge
```

crates/duty-tracker/src/predicates.rs
```diff
@@ -69,14 +72,26 @@ pub(crate) fn deposit_request_info(
...
         deposit_request_output.script_pubkey.clone(),
-    ))
+    );
+
+    // Regenerate the P2TR address from the OP_RETURN data, for now the spend info does all the
+    // necessary validations.
+    deposit_info
+        .compute_spend_infos(build_context, pegout_graph_params.refund_delay)
+        .map_err(|e| {
+            warn!("failed to compute spend info: {:?}", e);
```
Rajil1213, line 89: Should log the txid while we'e at it.

```suggestion
            warn!(err=%e, txid=%tx.compute_txid(), "failed to compute spend info, malformed DRT");
```

Rajil1213 (changes requested):
Looks like some things got borked while applying the suggestions from the previous round. These should be the final nits now.

Rajil1213 (approved):
Rebased this branch against main to fix the `sccache` issue.

PR #106, opened 2025-04-17
Update Deposit Tx Metadata to include take back key and amount

Rajil1213:
## Description

This PR updates the metadata output in the Deposit Transaction to hold the take back key of the user (as per the one in DRT) and the input amount. This data can be used to compute the prevout required to compute the transaction hash of the Deposit Transaction.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/primitives/src/deposit.rs
+++ b/crates/primitives/src/deposit.rs
@@ -246,6 +246,8 @@ impl DepositInfo {
             metadata: DepositMetadata::DepositTx {
                 stake_index: self.stake_index(),
                 ee_address: self.ee_address.to_vec(),
+                takeback_pubkey: self.x_only_public_key,
+                input_amount: self.total_amount,
             },
         };
         let metadata_script = metadata_script(metadata);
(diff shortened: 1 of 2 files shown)


crates/primitives/src/deposit.rs
```diff
@@ -486,4 +489,99 @@ mod tests {
...
+
+                    data = &data[metadata.len()..];
+                }
+
+                true
+            }));
+
+            true
+        }));
+    }
```
bewakes, line 586: This function is very long to follow along what's going on from top to bottom. Could you split this to helper functions?
Also for the last assertion part, seems like it would be simpler, if possible, to convert the expected metadata to OP_RETURN bytes and just compare with the instructions bytes. But this won't check if parsing/deserializing works or not.

bewakes (changes requested):

crates/primitives/src/deposit.rs
Rajil1213: > This function is very long to follow along what's going on from top to bottom. Could you split this to helper functions?

Agreed. Refactored it slightly here: https://github.com/alpenlabs/strata-bridge/commit/5d0efe90453e4c1b91553afeb8fb0469c5c2fdf4

Let me know if that helps. I'll add more comments if it's still not readable.

> Also for the last assertion part, seems like it would be simpler, if possible, to convert the expected metadata to OP_RETURN bytes and just compare with the instructions bytes

This would not tell us which part of the metadata was faulty unless we manually inspect the bytes. But that should be fine.

bewakes (approved):

PR #112, opened 2025-04-21
Replace x-only pk in dt metadata with merkle proof

Rajil1213:
## Description

This PR replaces the x-only public key in the deposit tx metadata with the merkle proof required to spend the DRT via the $$N/N$$ script spend path. This merkle proof is basically the hash of the take back script that can be used by the depositor to retrieve their funds should the deposit not succeed within a given time period.

### Type of Change

- [x] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/primitives/src/deposit.rs
+++ b/crates/primitives/src/deposit.rs
@@ -133,6 +133,7 @@ impl DepositInfo {
             *deposit_amount,
             tag.as_bytes(),
             sidesystem_params.address_length as usize,
+            pegout_graph_params.refund_delay,
         )?;
 
         let mut psbt = Psbt::from_unsigned_tx(unsigned_tx)?;
@@ -224,6 +225,7 @@ impl DepositInfo {
         deposit_amt: Amount,
         tag: &[u8],
         ee_address_size: usize,
+        refund_delay: u16,
     ) -> BridgeTxBuilderResult<Transaction> {
         // First, create the inputs
         let outpoint = self.deposit_request_outpoint();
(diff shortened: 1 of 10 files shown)


crates/primitives/src/deposit.rs
```diff
@@ -386,9 +391,9 @@ mod tests {
         let deposit_request_outpoint = OutPoint::null();
         let recovery_xonly_pk = generate_xonly_pubkey();
 
-        let refuned_delay = 1008;
+        let refund_delay = 1008;
```
storopoli, line 394: This should be a constant somewhere and
```suggestion
        let refund_delay = 1_008;
```

storopoli (changes requested):
minor nit.

I don't understand why are we reverting the X-only PK and going back to the root hash?

crates/primitives/src/deposit.rs
Rajil1213: The diff does not show it but it's actually inside a test. So I don't think we need to use a const.

Rajil1213: Will make the other change.

Rajil1213:
> I don't understand why are we reverting the X-only PK and going back to the root hash?

The reason for this is that using the x-only PK means that the OL needs to construct the take back script itself. This requires that the OL knows what the take back period is. This then means that the rollup params must also hold the take back period of the DRT. We want to reduce coupling between the OL and the bridge as far as possible, and using the hash means that the OL does not need care about the internals of the take back mechanism -- it just needs to care about the tapleaf hash (in order to construct the address).

storopoli (approved):
ACK 69d68f799dfa5638e52be2860bd5af029f684286

bewakes (dismissed):

Rajil1213:
Had to rebase this PR after the recent merges. So, re-requesting reviews.

bewakes (approved):

PR #114, opened 2025-04-22
feat(tx-graph): make PogInput serde compatible by using strata-p2p-types

Rajil1213:
## Description

This PR replaces the `wots` type inside the `PogInput` with those from the `strata-p2p-types`. The `connectors` crate still uses the types from `strata-bridge-primitives::wots` to maintain compatibility with the BitVM scripts. `Into` and `TryInto` have been implemented to convert from one type to another.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10496,6 +10496,7 @@ dependencies = [
  "strata-bridge-stake-chain",
  "strata-bridge-test-utils",
  "strata-btcio",
+ "strata-p2p-types",
  "thiserror 2.0.12",
  "tokio",
  "tracing",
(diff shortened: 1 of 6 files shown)


Rajil1213:
> What if we moved these types into our own crate here in strata-bridge from strata-p2p and then apply To/From conversions?
We will very soon remove the strata-p2p-types from strata-p2p so if we are doing this makes sense to bite the bullet and go all the way.

That's not as straightforward as it seems since all the message types in `strata-p2p` expect the types from `strata-p2p-types`. We could port over all the types into `strata-bridge` but then we'll have two copies of the same type but the compiler will still complain and so we'll have to convert between the two again.

ProofOfKeags (approved):
Yum 😋

PR #115, opened 2025-04-23
Add Ouroboros channel

Rajil1213:
## Description

This PR adds a mechanism to feed p2p gossipsub messages back to the node itself aka self-publishing. This is necessary since we want the state updates to only take place via the event loop input stream. This is achieved by adding a broadcast channel to the p2p message handler that also pushes messages to the channel when they are pushed to the network via the p2p handle. This ouroborus event is prioritized over the normal p2p messages.

Along the way, it also fixes the following issues:

* issue where in the StakeChainExchange message, the key used to advance the stake was not being shared.
* issue where the p2p messages were being processed but their corresponding duties were not being queued for execution.

Finally, it updates the nagging mechanism so that the deposit setup and graph nonces are not queried simultaneously. This is because during race conditions, the node will send out more nags than it should.

The `strata-p2p` dependency has also been updated to the latest.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [x] Dependency Update

## Checklist


bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -459,20 +459,32 @@ async fn init_operator_wallet(
 async fn handle_stakechain_genesis(
     db: SqliteDb,
     s2_client: SecretServiceClient,
-    message_handler: MessageHandler,
+    p2p_handle: P2PHandle,
```
storopoli, line 462: We should still give the `handle_stakechain_genesis` a `MessageHandler` and construct it outside the function.

`MessageHandler::dispatch` is using the ouroboros channel anyways and all `send` commands will pass through that.

storopoli (changes requested):
NACK we need to still keep the `MessageHandler` in the `init` function instead of the `P2PHandle`, see my arguments below.

Also it is [ouroboros](https://en.wikipedia.org/wiki/Ouroboros) and not ouroborus.

bin/alpen-bridge/src/mode/operator.rs
Rajil1213: The only reason we need the ouroboros channel is that we need to update the state post duty execution. That is not  a problem during genesis because the state is constructed only _after_ the genesis step, so there's nothing to update yet.

Rajil1213: Will correct the spelling. 😅

Rajil1213: Fixed the typo in https://github.com/alpenlabs/strata-bridge/commit/60903dfb21b8ef3b820ea15832243fdfb265b0e0.

storopoli: Yes but you are constructing the thing inside the function. Why not just leave as it was, accepting a `MessageHandler` and not construct it inside anymore from a `P2PHandle`?

Also `init_duty_tracker` should also keep `MessageHandler` as the inputs and not a `P2PHandle`.

Rajil1213: The reason for not leaving things as they were is that constructing the `MessageHandler` now requires us to pass in a `broadcast::sender` as well (for ouroboros). This is not required for the `stake_chain_exchange`. The sender/receiver are only relevant in the event loop created in the `ContractManager`. So, the `MessageHandler` is constructed in the `ContractManager` constructor and passed in the channels created there.

Rajil1213: I've updated the docs on the initializers and also updated the `MessageHandler` so that the underlying `p2p_handle` is not `pub`.

storopoli (approved):
ACK f66ae82b4fa7b5b5fae2e9ff94a98e0b4d26a514

PR #118, opened 2025-04-29
STR-970 use keyspend in drt output

Rajil1213:
## Description

This PR:

* Update the Deposit Transaction so that it spends the Deposit Request via the key-spend path instead of the script-spend path.
* Moves most of the structures in `strata_bridge_primitives::deposit` to `strata_bridge_tx_graph::deposit` so that all transactions relevant to the tx graph are now inside that crate. It also updates the API for the Deposit transaction so that it aligns with the rest of the transactions in the `tx-graph` crate.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The commits are atomic -- the first commit updates the spend path and the second one moves the crates around. Any changes to the `agent` crate can be ignored in the review process.

## Checklist

## Related Issues

Closes STR-970


storopoli (dismissed):
ACK 2b0fc3ff9c176f93f3413a2c5fec90d461b8a936

I think we should break straight away and avoid deprecating. Deprecating is useful when we have users that we need to gently steer towards better APIs etc.
We don't have that. We are the users now.

Rajil1213:
> We don't have that. We are the users now.

We are still using these in the `strata-bridge-agent` crate which is used in the `dev-bridge` binary. I'm keeping these around till we have an end-to-end working version of the testent-i bridge.

storopoli:
> > We don't have that. We are the users now.
> 
> We are still using these in the `strata-bridge-agent` crate which is used in the `dev-bridge` binary. I'm keeping these around till we have an end-to-end working version of the testent-i bridge.

Why not tackle these changes in this PR as well? I know we don't want to touch the `agent` crate but I prefer touching it a little bit than having deprecations in a pre-0.1.0 codebase.

Rajil1213:
> Why not tackle these changes in this PR as well? I know we don't want to touch the agent crate but I prefer touching it a little bit than having deprecations in a pre-0.1.0 codebase.

That's because it's not a trivial change (it depends on the duty system in `strata` that we had in devnet that relies on the older deposit types). There are a lot of changes that need to be made to remove all the things that that the `agent` crate depends upon. And since the near-term plan is to just remove it, that kind of effort is not worth it.

storopoli (approved):
ACK d80738b65be95e94392af02431f876bd2c7dccd2

PR #121, opened 2025-04-30
feat(s2) sig/nonce batched receives

Zk2u:
## Description

Changes the musig2 rounds API to allow batched calls to the `receive_pub_nonces` and `receive_signatures` methods. 

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/secret-service-proto/src/v1/wire.rs
+++ b/crates/secret-service-proto/src/v1/wire.rs
@@ -93,7 +93,7 @@ pub enum ServerMessage {
     },
 
     /// Response for
-    /// [`Musig2SignerFirstRound::receive_pub_nonce`](super::traits::Musig2SignerFirstRound::receive_pub_nonce).
+    /// [`Musig2SignerFirstRound::receive_pub_nonces`](super::traits::Musig2SignerFirstRound::receive_pub_nonces).
     Musig2FirstRoundReceivePubNonce(
         /// Errors indicating whether the server was unable to process the request, indexed by the
         /// pubnonce's signer's xonly pubkey. If empty, should be returned as a Ok(())
@@ -139,7 +139,7 @@ pub enum ServerMessage {
     },
 
     /// Response for
-    /// [`Musig2SignerSecondRound::receive_signature`](super::traits::Musig2SignerSecondRound::receive_signature).
+    /// [`Musig2SignerSecondRound::receive_signatures`](super::traits::Musig2SignerSecondRound::receive_signatures).
     Musig2SecondRoundReceiveSignature(
         /// Any errors that occurred during signature reception, keyed by the partial signature's
         /// signer's xonly pubkey If empty, should be returned as a Ok(())
@@ -315,7 +315,7 @@ pub enum ClientMessage {
     },
 
     /// Request for
-    /// [`Musig2SignerFirstRound::receive_pub_nonce`](super::traits::Musig2SignerFirstRound::receive_pub_nonce).
+    /// [`Musig2SignerFirstRound::receive_pub_nonces`](super::traits::Musig2SignerFirstRound::receive_pub_nonces).
     Musig2FirstRoundReceivePubNonce {
         /// Session that this server is requesting for.
         #[rkyv(with = rkyv_wrappers::OutPoint)]
@@ -369,7 +369,7 @@ pub enum ClientMessage {
     },
 
     /// Request for
-    /// [`Musig2SignerSecondRound::receive_signature`](super::traits::Musig2SignerSecondRound::receive_signature).
+    /// [`Musig2SignerSecondRound::receive_signatures`](super::traits::Musig2SignerSecondRound::receive_signatures).
     Musig2SecondRoundReceiveSignature {
         /// Session that this server is requesting for.
         #[rkyv(with = rkyv_wrappers::OutPoint)]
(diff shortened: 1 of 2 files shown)


storopoli (commented):
The code seems fine. But I do think that the user (us for now) the order matter since the keys are `XOnlyPublicKey`s. Hence, I think it should be a `BTreeMap` and not a `HashMap`.

storopoli:
> @storopoli hold my cup

Why not also returning an `BTreeMap` instead of `HashMap` in these functions.

ProofOfKeags:
No notes beyond what @storopoli has already pointed out.

Rajil1213:
> switched the api to btreemaps, the tests are failing because the rest of the code isn't using the batched receives.

Now that the rest of the deposit stuff has been merged that ought to use this API, could you also update those?

On the subject of `BTreeMap` vs `HashMap`, either choice is fine because the order that the `BTreeMap` gives us is not the order that we actually use, and the `RoundContributionError` from musig2 also holds the index of the signer so we do have visibility over what signer screwed up.

bin/secret-service/src/seeded_impl/musig2.rs
```diff
@@ -150,17 +153,29 @@ impl Musig2SignerFirstRound<Server, ServerSecondRound> for ServerFirstRound {
...
+                Some(idx) => idx,
+                None => {
+                    errs.insert(pubkey, RoundContributionError::out_of_range(0, 0));
+                    continue;
+                }
+            };
+            if let Err(e) = self.first_round.receive_nonce(signer_idx, nonce) {
+                errs.insert(pubkey, e.clone());
+            }
+        }
```
ProofOfKeags, line 173: Fun Fact (irrelevant to review but you may find interesting): this is the [Validation Applicative](https://hackage.haskell.org/package/validation-1.1.3/docs/Data-Validation.html#t:Validation)

storopoli: Damn it accumulates errors!

crates/secret-service-client/src/musig2.rs
```diff
@@ -159,13 +159,12 @@ impl Musig2SignerFirstRound<Client, Musig2SecondRound> for Musig2FirstRound {
 
     async fn receive_pub_nonces(
         &mut self,
-        nonces: HashMap<XOnlyPublicKey, PubNonce>,
+        nonces: impl Iterator<Item = (XOnlyPublicKey, PubNonce)> + Send,
```
ProofOfKeags, line 162: since you're serializing them anyway don't you only need an iterator of borrowed items? For the server the owned ones make sense I think. Does your trait system make it such that it's required for us to make the types the same on the client and server? I don't think the extra clone is an issue fwiw but I've just been paying more attention to owned vs borrowed stuff lately and I'm curious.

crates/secret-service-server/src/lib.rs
```diff
@@ -7,7 +7,13 @@
 
 pub mod musig2_session_mgr;
 
-use std::{collections::HashMap, io, marker::Sync, net::SocketAddr, sync::Arc};
+use std::{
+    collections::{BTreeMap, HashMap},
```
ProofOfKeags, line 11: Do you still need the HashMap import?

ProofOfKeags (changes requested):
A couple small changes I'd like to see, namely a removal of the nested Result structure, and then also a systematic review of the `InvalidClientMessage` error. I think we should only use that for errors resulting from improper "syntax": if we can't parse messages. Otherwise we should use errors that indicate what semantics were violated.

crates/secret-service-proto/src/v1/wire.rs
```diff
@@ -314,12 +321,8 @@ pub enum ClientMessage {
...
         session_id: Musig2SessionId,
 
-        /// The serialized [`XOnlyPublicKey`](bitcoin::XOnlyPublicKey) of the signer whose public
-        /// nonce this is.
-        pubkey: [u8; 32],
-
-        /// Serialized public nonce
-        pubnonce: [u8; 66],
+        /// Public nonces keyed by the signer's xonly public key
+        nonces: HashMap<[u8; 32], [u8; 66]>,
```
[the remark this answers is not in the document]
storopoli: https://docs.rs/musig2/latest/musig2/struct.PubNonce.html

> Represents a public nonce derived from a secret nonce. It is composed of two public points, `R1` and `R2`, derived by base-point multiplying the two scalars in a `SecNonce`.

To be fair, I should probably do a PR in `musig2` to add a `{SEC,PUB}_NONCE_SIZE` (64 and 66) and `PARTIAL_SIGNATURE_SIZE` (32) similar to https://docs.rs/musig2/latest/musig2/constant.SCHNORR_SIGNATURE_SIZE.html

crates/secret-service-client/src/musig2.rs
ProofOfKeags: It's not super important. Don't worry about it. Just something to let simmer in the back of your mind.

storopoli (approved):
ACK b2f2021165ee0c978f4fe02869a4d04751af82da

ProofOfKeags (approved):

PR #122, opened 2025-04-30
STR-1309: Implement the RPC server

storopoli:
## Description

Implements the RPC server than can query hot state from the `ContractManager` with respect to active contracts and fallback to the `contracts` SQL table for non-active contracts (or contracts that cannot be found in the hot state).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The RPC needs a fresh state from the `ContractManager`.
However both are on separate tokio threads, hence we need a tokio channel to send the latest hot state from the `ContractManager` to the RPC server thread.
I've used [`watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) channels since per [tokio doc](https://docs.rs/tokio/latest/tokio/sync/index.html#watch-channel):

> The [watch channel](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) supports sending many values from a many producer to many consumers. However, only the most recent value is stored in the channel. Consumers are notified when a new value is sent, but there is no guarantee that consumers will see all values.

> The [watch channel](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) is similar to a [broadcast channel](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) with capacity 1.

It also adds a config to for the RPC state refresh interval, similar to the "nag interval".
Note that this adds a new entry in the "event loop" of the `ContractManager`.

## Checklist

## Related Issues

STR-1309

### DIFF
--- a/bin/alpen-bridge/src/mode/operator.rs
+++ b/bin/alpen-bridge/src/mode/operator.rs
@@ -1,5 +1,6 @@
 //! Defines the main loop for the bridge-client in operator mode.
 use std::{
+    collections::BTreeMap,
     env, fs, io,
     path::{Path, PathBuf},
     sync::Arc,
@@ -12,7 +13,7 @@ use bitcoin::{
     hashes::Hash,
     secp256k1::SecretKey,
     sighash::{Prevouts, SighashCache, TapSighashType},
-    FeeRate, OutPoint, TxOut, XOnlyPublicKey,
+    FeeRate, OutPoint, TxOut, Txid, XOnlyPublicKey,
 };
 use bitcoind_async_client::{
     traits::{Broadcaster, Reader},
@@ -21,7 +22,8 @@ use bitcoind_async_client::{
 use btc_notify::client::BtcZmqClient;
 use duty_tracker::{
     contract_manager::ContractManager, contract_persister::ContractPersister,
-    stake_chain_persister::StakeChainPersister, tx_driver::TxDriver,
+    contract_state_machine::ContractSM, stake_chain_persister::StakeChainPersister,
+    tx_driver::TxDriver,
 };
 use libp2p::{
     identity::{secp256k1::PublicKey as LibP2pSecpPublicKey, PublicKey as LibP2pPublicKey},
@@ -52,8 +54,13 @@ use strata_bridge_primitives::{
 use strata_bridge_stake_chain::prelude::OPERATOR_FUNDS;
 use strata_p2p::swarm::handle::P2PHandle;
 use strata_p2p_types::{P2POperatorPubKey, StakeChainId};
-use tokio::{spawn, sync::broadcast, task::JoinHandle, try_join};
-use tracing::{debug, info};
+use tokio::{
+    spawn,
+    sync::{broadcast, watch},
+    task::JoinHandle,
+    try_join,
+};
+use tracing::{debug, info, trace};
 
 use crate::{
     config::{Config, P2PConfig, SecretServiceConfig},
@@ -140,7 +147,7 @@ pub(crate) async fn bootstrap(params: Params, config: Config) -> anyhow::Result<
     let zmq_client = BtcZmqClient::connect(&config.btc_zmq)
         .await
         .expect("should be able to connect to zmq");
-    let contract_manager_task = init_duty_tracker(
+    let (contract_manager_task, state_snapshot_rx) = init_duty_tracker(
         &params,
         &config,
         bitcoin_rpc_client.clone(),
@@ -155,7 +162,14 @@ pub(crate) async fn bootstrap(params: Params, config: Config) -> anyhow::Result<
 
     info!("starting the RPC server");
     let rpc_address = config.rpc_addr.clone();
-    let rpc_task = start_rpc_server(rpc_address, db_rpc, p2p_handle_rpc, params.clone()).await?;
+    let rpc_task = init_rpc_server(
+        rpc_address,
+        db_rpc,
+        p2p_handle_rpc,
+        state_snapshot_rx,
+        params.clone(),
+    )
+    .await?;
     info!("started the RPC server");
 
     // Wait for all tasks to run
@@ -339,9 +353,10 @@ async fn init_duty_tracker(
     p2p_handle: P2PHandle,
     operator_wallet: OperatorWallet,
     db: SqliteDb,
-) -> anyhow::Result<JoinHandle<()>> {
+) -> anyhow::Result<(JoinHandle<()>, watch::Receiver<BTreeMap<Txid, ContractSM>>)> {
     let network = params.network;
     let nag_interval = config.nag_interval;
+    let rpc_interval = config.rpc_interval;
     let connector_params = params.connectors;
     let pegout_graph_params = params.tx_graph.clone();
     let stake_chain_params = params.stake_chain;
@@ -377,6 +392,7 @@ async fn init_duty_tracker(
     Ok(ContractManager::new(
         network,
         nag_interval,
+        rpc_interval,
         connector_params,
         pegout_graph_params,
         stake_chain_params,
@@ -394,17 +410,29 @@ async fn init_duty_tracker(
     ))
 }
 
-async fn start_rpc_server(
+async fn init_rpc_server(
     rpc_address: String,
     db: SqliteDb,
     p2p_handle: P2PHandle,
+    state_snapshot_rx: watch::Receiver<BTreeMap<Txid, ContractSM>>,
     params: Params,
 ) -> anyhow::Result<JoinHandle<()>> {
-    let rpc_client = BridgeRpc::new(db, p2p_handle, params);
+    let mut rpc_client = BridgeRpc::new(db, p2p_handle, state_snapshot_rx, params);
     let handle = spawn(async move {
         start_rpc(&rpc_client, rpc_address.as_str())
             .await
             .expect("failed to start RPC server");
+        let mut rx = rpc_client.get_state_snapshot_rx();
+        loop {
+            if rx.changed().await.is_ok() {
+                debug!("state snapshot changed");
+                let snapshot = rx.borrow();
+                trace!(?snapshot, "state snapshot");
+                rpc_client.update_state_snapshot(snapshot.clone());
+            } else {
+                trace!("state snapshot not changed");
+            }
+        }
     });
     Ok(handle)
 }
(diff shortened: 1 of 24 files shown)


Rajil1213 (changes requested):
I've only skimmed through this since the PR is at a relatively early stage but the main issue is that the implementation depends on the `db` crate whereas the duty tracker does not fully use it at the moment. As discussed in one of our calls (last week, I think), the RPC struct should have a watch channel (or a similar construct) that always receives the latest copy of the `ExecutionState` which holds all the contracts along with the `StakeChainSM`. This will always provide the latest state of all the contracts and can be used to extract the information required by the RPC traits. It may also be the case that some information is not currently readily available in which case, we'll need to update the states.

bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -394,17 +410,31 @@ async fn init_duty_tracker(
...
+            if rx.changed().await.is_ok() {
+                debug!("state snapshot changed");
+                let snapshot = rx.borrow();
+                trace!(?snapshot, "state snapshot");
+                rpc_client.update_state_snapshot(snapshot.clone());
+            } else {
+                trace!("state snapshot not changed");
+            }
+            // Sleep for 1 second to avoid busy-waiting.
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
Rajil1213, line 436: I don't think this is required. It won't busy loop because tokio would wait on the `await` point. This is easy to verify if we replace this sleep with a log. If you don't see a bazillion logs, I think it's okay to remove this.

Rajil1213 (changes requested):

bin/alpen-bridge/src/mode/operator.rs
storopoli: Removed in e4bbdea577a7abc856042a80fdbce88b6136fc35

migrations/20241113060253_init_schema.sql
```diff
@@ -166,6 +166,14 @@ CREATE TABLE IF NOT EXISTS duty_tracker (
     status TEXT NOT NULL                  -- Status of the duty as JSON
 );
 
+-- Table to store bridge duties
+CREATE TABLE IF NOT EXISTS bridge_duties (
+    duty_id TEXT PRIMARY KEY,             -- Unique identifier for each duty as an encoded txid
+    operator_pk TEXT,                     -- Operator's public key that the duty is for
+    duty_type TEXT NOT NULL,              -- Type of duty (SignDeposit or FulfillWithdrawal)
+    duty_data BLOB NOT NULL              -- Serialized duty data (DepositInfo or WithdrawalInfo)
+);
```
ProofOfKeags, line 200: I'm highly suspicious of this. Is this table a redundant summary table?

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -9,7 +9,7 @@ use libp2p::{identity::PublicKey as LibP2pPublicKey, PeerId};
 use secp256k1::Parity;
 use strata_bridge_db::{persistent::sqlite::SqliteDb, tracker::DutyTrackerDb};
 use strata_bridge_primitives::duties::{
-    BridgeDuty, ClaimStatus, DepositRequestStatus, WithdrawalStatus,
+    BridgeDuties, ClaimStatus, DepositRequestStatus, WithdrawalStatus,
```
ProofOfKeags: What is this and why have I never seen it before o.O

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -746,6 +746,16 @@ impl ContractSM {
         (contract_sm, duty)
     }
 
+    /// Gets the configuration of the contract.
+    pub fn get_cfg(&self) -> &ContractCfg {
```
ProofOfKeags, line 840: I don't like the naming of `get` anywhere. We also have these functions already: `ContractSM::cfg` and `ContractSM::state`

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -181,25 +181,32 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
...
-            Some(deposit) => Ok(deposit),
-            None => Err(ErrorObjectOwned::owned::<_>(
-                -32001,
-                "Deposit request outpoint not found",
-                Some(deposit_request_outpoint),
-            )),
+        // If it is Requested, then we can just return the InProgress status.
+        // Otherwise, it is complete.
+        if let Some(contract_sm) = self.current_state.get(&deposit_request_outpoint.txid) {
+            match contract_sm.get_state() {
```
ProofOfKeags: you can lift out the `return Ok(...)` to outside the match statement if you want and it may clean things up a bit.

ProofOfKeags: Also this is one of those situations where a `let...else` may be more appropriate.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -213,6 +213,9 @@ pub enum ContractState {
     /// This state describes everything from the moment the deposit confirms, to the moment the
     /// strata state commitment that assigns this deposit confirms.
     Deposited {
+        /// The txid of the deposit request transaction that kicked off this contract.
+        deposit_request_txid: Txid,
```
ProofOfKeags, line 217: is there a reason you aren't using `ContractSM::deposit_request_txid`? If you go this route, please don't put it in state, put it in config where it belongs as this doesn't change for the lifetime of the contract.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -446,8 +476,68 @@ impl ContractState {
...
             ContractState::Asserted { peg_out_graphs, .. } => get_summaries(peg_out_graphs),
-            ContractState::Disproved {} => vec![],
-            ContractState::Resolved {} => vec![],
+            ContractState::Disproved { .. } => vec![],
+            ContractState::Resolved { .. } => vec![],
+        }
+    }
+
+    /// Returns the transaction ID of the deposit request for this contract state.
+    pub fn deposit_request_txid(&self) -> Option<Txid> {
```
ProofOfKeags, line 485: This function should not exist.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1842,4 +1951,48 @@ impl ContractSM {
...
+            } => Some(*assignment_txid),
+            ContractState::StakeTxReady {
+                assignment_txid, ..
+            } => Some(*assignment_txid),
+            ContractState::Fulfilled { .. } => None,
+            ContractState::Claimed { .. } => None,
+            ContractState::Challenged { .. } => None,
+            ContractState::Asserted { .. } => None,
+            ContractState::Disproved {} => None,
+            ContractState::Resolved {} => None,
```
ProofOfKeags, line 1974: Good choice making this explicit as opposed to catch all 🫡

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -47,11 +47,11 @@ impl ContractPersister {
             // TODO(proofofkeags): make state not opaque at the DB level
             r#"
             CREATE TABLE IF NOT EXISTS contracts (
-                deposit_txid CHAR(64) PRIMARY KEY,
+                deposit_txid TEXT PRIMARY KEY,
```
ProofOfKeags, line 50: why?

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -181,27 +182,96 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
...
-                ContractState::Requested {
-                    deposit_request_txid,
-                    ..
-                } => {
-                    return Ok(DepositRequestStatus::InProgress {
-                        deposit_request_txid: *deposit_request_txid,
-                    })
+        let deposit_request_txid = deposit_request_outpoint.txid;
+        // Iterate over all contract states to find the matching deposit request
+        for contract_sm in self.current_state.values() {
```
ProofOfKeags, line 188: are you duplicating the CSMs?

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -279,14 +280,162 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
     }
 
     async fn get_bridge_duties(&self) -> RpcResult<Vec<RpcBridgeDutyStatus>> {
-        todo!()
+        // we don't care about the hot state here, we only care about the database
```
ProofOfKeags, line 284: Is this just so we can track the status of duties that are in-flight?

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -667,6 +667,7 @@ pub struct MachineState {
     pub state: ContractState,
 }
 
+#[derive(Debug, Clone)]
```
ProofOfKeags: No. This summons demons.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -342,11 +350,20 @@ impl ContractManager {
...
+                        // Process nag messages
                         let nags = ctx.nag();
                         for nag in nags {
                             p2p_handle.send_command(nag).await;
                         }
+                    },
+
+                    _ = rpc_interval.tick() => {
+                        // Send state snapshot
+                        let snapshot = ctx.state.active_contracts.clone();
```
ProofOfKeags, line 363: No. Demons.

ProofOfKeags (changes requested):
We need to have a conversation about two things mainly:
1. The role of the database tables you're introducing here. It's important that we don't really persist derived state unless we have a good reason for it. I'm open to this being fine but there need to be rules around it.
2. Why making the CSM Clone is a demon summoning spell. The synopsis is that the CSM is the authoritative resource governing the state transitions. What I think you want here is a copy of the CSM State and a copy of the CSM config. The CSM should fundamentally not be clone because we should not be able to call any of the `process_` functions on any clones of the CSM. This has the potential to let us have divergent state and that is where the demons lie.

storopoli:
Closing in favor of #134

crates/duty-tracker/src/contract_persister.rs
storopoli: To make consistent with all other tables, but let's move discussion to #134

PR #128, opened 2025-05-03
Parallelize graph generation

ProofOfKeags:
## Description

This PR is a first draft of parallelizing graph generation inside of a single CSM. Currently it takes a very naive approach where we do not consider any thread pools, nor do we use async/green threads. This choice was made for expediency so we could get immediate practical benefits in testing while we consider the best way to handle this at larger scales.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/duty-tracker/src/contract_state_machine.rs
+++ b/crates/duty-tracker/src/contract_state_machine.rs
@@ -1010,7 +1010,13 @@ impl ContractSM {
                         (
                             signer,
                             // TODO(proofofkeags): use async thread pool in future commit
-                            thread::spawn(move || thread_cfg.build_graph(input)),
+                            thread::Builder::new()
+                                .stack_size(8 * 1024 * 1024)
+                                .spawn(move || {
+                                    info!("building graph...");
+                                    thread_cfg.build_graph(input)
+                                })
+                                .expect("spawn succeeds"),
                         )
                     })
                     .collect::<BTreeMap<_, _>>();
--- a/crates/duty-tracker/src/contract_state_machine.rs
+++ b/crates/duty-tracker/src/contract_state_machine.rs
@@ -1,7 +1,7 @@
 //! This module defines the core state machine for the Bridge Deposit Contract. All of the states,
 //! events and transition rules are encoded in this structure. When the ContractSM accepts an event
 //! it may or may not give back an OperatorDuty to execute as a result of this state transition.
-use std::{collections::BTreeMap, fmt::Display};
+use std::{collections::BTreeMap, fmt::Display, sync::Arc, thread};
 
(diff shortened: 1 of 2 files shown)


crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -979,41 +995,68 @@ impl ContractSM {
...
-                let stake_txid = pog_input.stake_outpoint.txid;
-                let pog = if let Some(pog) = self.pog.get(&stake_txid) {
-                    debug!(reimbursement_key=%operator_pubkey, %stake_txid, "retrieving peg out graph from cache");
-                    pog.clone()
-                } else {
-                    debug!(reimbursement_key=%operator_pubkey, %stake_txid, "generating and caching peg out graph");
-                    let pog = self.cfg.build_graph(pog_input.clone());
-                    self.pog.insert(stake_txid, pog.clone());
+                if peg_out_graph_inputs.len() != self.cfg.operator_table.cardinality() {
+                    return Ok(vec![]);
```
Rajil1213, line 1001: You added an early return? 🤯

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -979,41 +995,68 @@ impl ContractSM {
...
+
+                let graphs = jobs
+                    .into_iter()
+                    .map(|(signer, job)| {
+                        (
+                            signer,
+                            job.join().expect("peg out graph generation panic'ed"),
+                        )
+                    })
+                    .collect::<BTreeMap<_, _>>();
```
Rajil1213, line 1032: Can we not just do `join_all` here?

Rajil1213 (changes requested):

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: Sadly I did, yes.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: I think that join_all is just a futures thing? I can't find it in std.

storopoli: https://docs.rs/futures/latest/futures/future/fn.join_all.html

ProofOfKeags: These are not async tasks, they are os threads.

storopoli (approved):
ACK f32249a2a5533aac09ea0c80180d25f6621a2615

But I'm not yet completely sold on the OS threads vs async green threads with `tokio`.

PR #130, opened 2025-05-06
chore(deps): bump corepc-node to 0.7.1

storopoli:
## Description

There's a RUSTSEC advisory on time that should be updated to >=0.2.23. It gets pulled down from corepc-node with download feature from the zip dependency, see the cargo tree below:

```
time 0.1.45
└── zip 0.5.13
    └── corepc-node 0.7.0
```

This was fixed in corepc-node 0.7.1, see
https://github.com/rust-bitcoin/corepc/pull/143

Closes https://github.com/alpenlabs/strata-bridge/security/dependabot/14.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [x] Dependency Update
- [ ] Security Update

## Notes to Reviewers

This also moves the `corepc-node` from the workspace `dependencies` to crates `dev-dependencies` since `corepc-node` build times with the `download` feature incurs in a bitcoin core download every time we do `cargo check` or `cargo clippy` or a rust-analyzer incremental build.

## Checklist


barakshani (approved):

Rajil1213 (changes requested):
I don't think `corepc-node` needs to be removed from `workspace.dependencies`. In general, simply adding a dependency as a workspace dependency does not actually cause it to be built if it is not used in any of the member crates.

In our case, even if you do add it to dev-dependency, `corepc-node` will still be built if you pass in the `--workspace` flag to `clippy` or `check` because it is _not_ a dev-dependency in the `test-utils` crate. If you do not pass the `--workspace` flag, it will only build the default members and so it won't build `corepc-node` even though it is part of the workspace dependencies. I've verified this locally but I do recommend validating it just to make sure.

storopoli:
Oh the `test-utils` crate, I see.
Done in 22b7642e214bceef949eb404ecdb3ed91dc0cbbb

Rajil1213 (approved):

PR #137, opened 2025-05-08
STR-1331 Optimistic Fulfiller Duties Execution

Rajil1213:
## Description

This PR adds the ability for operator nodes to fulfill withdrawal requests and to get reimbursed provided that they are not challenged.

In so doing, it also makes the following additional changes:

* updates the `strata` dependency to use the updated checkpoint/chain-state structure.
* updates the `bitvm` dependency to use the optimized graph generation.
* updates most of the STFs to handle cases where a contract may receive txs belonging to other graphs tied to the same deposit.
* updates the transaction APIs in the `tx-graph` crate to not require connectors when finalizing.
* updates the dockerfile for bitcoind to automine blocks given a certain env variable.
* handles an edge case where we proceed to the next step before all nonces are received for all graphs (previously, we only used to check the number of nonces per graph and not the number of graphs themselves).
* updates the stake chain API to compute sighashes just-in-time with the supplied prevouts

### TODO

- [x] rebase with `main`.
- [x] update the prover test files to fix failing tests.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

Good luck. 👀 

## Checklist

## Related Issues

Closes STR-1331


Rajil1213:
I've rebased this with `main` and also updated the proof test-data. So this is actually ready for review now.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -214,6 +215,9 @@ pub enum ContractState {
         /// contract through its lifecycle, as well as reconstructing the graph when necessary.
         peg_out_graphs: BTreeMap<Txid, (PegOutGraphInput, PegOutGraphSummary)>,
 
+        /// This is a collection of all partial signatures for all graphs and for all operators.
+        graph_partials: BTreeMap<Txid, BTreeMap<P2POperatorPubKey, PogMusigF<PartialSignature>>>,
```
ProofOfKeags, line 219: if my suspicions are correct, this is something you'll undo in a later commit but I don't think we should be threading the graph partials to future states. We should extract the signatures as soon as we can at a minimum.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1084,13 +1084,25 @@ impl ContractSM {
...
             ContractState::StakeTxReady { .. } => self.process_fulfillment_confirmation(tx),
             ContractState::Fulfilled { .. } => self.process_claim_confirmation(height, tx),
-            ContractState::Claimed { .. } => self
-                .process_challenge_confirmation(tx)
-                .or_else(|_| self.process_optimistic_payout_confirmation(tx)),
+            ContractState::Claimed { .. } => {
+                // could be challenged
+                if let Some(duty) = self.process_challenge_confirmation(tx)? {
+                    return Ok(Some(duty));
+                }
```
ProofOfKeags, line 1093: This doesn't seem right. This seems to imply that if you pass in a optimistic payout tx to `process_challenge_confirmation` that it comes back as `Ok(...)`. This is wrong and should be changed. It should be an error and we should be cascading on error.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1084,13 +1084,25 @@ impl ContractSM {
...
+            }
             ContractState::Challenged { .. } => self.process_assert_chain_confirmation(height, tx),
-            ContractState::Asserted { .. } => self
-                .process_disprove_confirmation(tx)
-                .or_else(|_| self.process_defended_payout_confirmation(tx)),
+            ContractState::Asserted { .. } => {
+                // could be disproved
+                if let Some(duty) = self.process_disprove_confirmation(tx)? {
+                    return Ok(Some(duty));
+                }
```
ProofOfKeags, line 1103: same goes here.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1373,10 +1375,12 @@ impl ContractSM {
...
 
-                let required_partials = self.cfg.operator_table.cardinality();
-                let have_all_partials = graph_partials
+                let num_operators = self.cfg.operator_table.cardinality();
+                let have_all_graphs = graph_partials.values().count() == num_operators;
+                let have_all_partials_for_all_graphs = graph_partials
                     .values()
-                    .all(|session_partials| session_partials.len() == required_partials);
+                    .all(|session_partials| session_partials.len() == num_operators);
+                let have_all_partials = have_all_graphs && have_all_partials_for_all_graphs;
```
ProofOfKeags, line 1388: this is one of those times where laziness by default just makes performance naturally optimized

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1667,7 +1667,9 @@ impl ContractSM {
         }
 
         let current = std::mem::replace(&mut self.state.state, ContractState::Resolved {});
-        match &current {
+        let copy_of_current = current.clone();
```
ProofOfKeags, line 1675: don't do this. instead use an `@` binding

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1785,9 +1790,10 @@ impl ContractSM {
                 graph_partials,
             } => {
                 if tx.compute_txid() != active_graph.1.stake_txid {
-                    return Err(TransitionErr(format!(
-                        "stake chain advancement txid ({}) doesn't match the stake txid of the active graph ({})", tx.compute_txid(), active_graph.1.stake_txid,
-                    )));
+                    // might be somebody else's stake txid
+                    self.state.state = copy_of_current;
+
+                    return Ok(None);
```
ProofOfKeags, line 1800: yeah definitely don't do this, this violates a desired property of the CSM:

Ok(...) == csm.process_x(evt) => state prior != state after
Err(...) == csm.process_x(evt) => state prior == state after

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1919,7 +1919,10 @@ impl ContractSM {
         height: BitcoinBlockHeight,
         tx: &Transaction,
     ) -> Result<Option<OperatorDuty>, TransitionErr> {
+        debug!(txid=%tx.compute_txid(), %height, "processing confirmation of claim tx");
+
         let current = std::mem::replace(&mut self.state.state, ContractState::Resolved {});
+        let copy_of_current = current.clone();
```
ProofOfKeags, line 1929: again we need to not be doing this copying stuff

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1995,10 +1996,10 @@ impl ContractSM {
...
             } => {
                 if !is_challenge(active_graph.1.claim_txid)(tx) {
-                    return Err(TransitionErr(format!(
-                        "invalid challenge transaction ({}) in process_challenge_confirmation",
-                        tx.compute_txid()
-                    )));
+                    // could be an optimistic payout
+                    self.state.state = copy_of_current;
+
+                    return Ok(None);
```
ProofOfKeags, line 2006: don't do this. same issue regarding the state diff property mentioned elsewhere

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -482,11 +500,12 @@ impl ContractManagerCtx {
                     self.cfg.stake_chain_params,
                     height,
                     height + self.cfg.pegout_graph_params.refund_delay as u64,
-                    stake_index,
+                    stake_index + new_deposit_requests,
                     deposit_request_txid,
                     deposit_tx,
                     stake_chain_inputs,
                 );
+                new_deposit_requests += 1;
```
ProofOfKeags, line 507: you don't need to do this. just use the length of the new_contracts vec.

crates/db/src/persistent/sqlite.rs
```diff
@@ -408,7 +408,7 @@ impl PublicDb for SqliteDb {
                 let stake_data = DbStakeTxData::from(stake_data);
 
                 sqlx::query!(
-                    "INSERT OR REPLACE INTO operator_stake_data
+                    "INSERT OR IGNORE INTO operator_stake_data
```
ProofOfKeags, line 411: 🫡

crates/duty-tracker/src/executors/deposit.rs
```diff
@@ -237,22 +237,6 @@ pub(crate) async fn handle_publish_deposit_setup(
...
-        hash: stakechain_preimg_hash,
-        withdrawal_fulfillment_pk: strata_bridge_primitives::wots::Wots256PublicKey(
-            withdrawal_fulfillment_pk,
-        ),
-    };
-
-    info!(%deposit_txid, %deposit_idx, "adding stake data to the database");
-    debug!(%deposit_txid, %deposit_idx, ?stake_data, "adding stake data to the database");
-
-    db.add_stake_data(pov_idx, deposit_idx, stake_data).await?;
```
ProofOfKeags: I assume you did this due to locking issues?

ProofOfKeags (changes requested):
Beautifully organized PR. Content wise I have a lot to say, though. I'll leave the details to the individual comments. Any 🫡 emojis should be understood as explicit acks. This helps as we iterate on this PR to know what we've already looked at.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -1262,89 +1264,31 @@ async fn execute_duty(
...
-            witness,
-        } => {
-            handle_publish_root_nonce(
-                &output_handles.s2_session_manager,
-                &output_handles.msg_handler,
-                OutPoint::new(deposit_request_txid, 0),
-                witness,
-            )
-            .await
-        }
```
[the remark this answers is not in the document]
Rajil1213: All the executors have been moved to a separate module inside the duty tracker crate.

crates/duty-tracker/src/executors.rs
```diff
@@ -0,0 +1,435 @@
...
+            .push(signature.to_vec());
+    }
+
+    info!(
+        txid = %tx.compute_txid(),
+        "submitting claim funding tx to the tx driver"
+    );
+    tx_driver
+        .drive(tx)
+        .await
```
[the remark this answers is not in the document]
Rajil1213: It is _definitely_ a good idea to have more options here. But we should defer on that.

crates/duty-tracker/src/executors.rs
```diff
@@ -51,22 +55,24 @@ pub(super) async fn handle_publish_deposit_setup(
...
     );
     const NUM_FQS: usize = NUM_U256;
     const NUM_PUB_INPUTS: usize = NUM_PUBS;
     const NUM_HASHES: usize = NUM_HASH;
-    let public_inputs_ftrs: [_; NUM_PUB_INPUTS] =
-        std::array::from_fn(|i| wots_client.get_256_public_key(deposit_txid, VOUT, i as u32));
+
+    let public_inputs_ftrs: [_; NUM_PUB_INPUTS] = std::array::from_fn(|i| {
+        wots_client.get_256_public_key(deposit_txid, VOUT, (i + PUBLIC_INPUTS_PK_OFFSET) as u32)
+    });
```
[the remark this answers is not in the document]
Rajil1213: Shouldn't be. It is supposed to take the config from `rustfmt.toml`. As long as the CI does not complain, it should be fine.

crates/duty-tracker/src/executors.rs
```diff
@@ -439,3 +446,91 @@ pub(super) async fn handle_withdrawal_fulfillment(
...
+    deposit_txid: Txid,
+    withdrawal_fulfillment_txid: Txid,
+) -> Result<(), ContractManagerErr> {
+    let pov_idx = cfg.operator_table.pov_idx();
+
+    // the input to the claim transaction is the input to the stake transaction minus the two dust
+    // outputs in the stake transaction.
+    let input_amount = OPERATOR_FUNDS
+        .checked_sub(SEGWIT_MIN_AMOUNT * 2)
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
Rajil1213: This has been refactored out to inside the individual transaction(s) where the test coverage is comprehensive enough that I'm not too worried about it accidentally being zero. I'll throw some assertions in there just to be extra safe.

crates/duty-tracker/src/executors.rs
```diff
@@ -534,3 +540,151 @@ pub(super) async fn handle_publish_claim(
...
+    deposit_txid: Txid,
+    claim_txid: Txid,
+    stake_txid: Txid,
+    stake_index: u32,
+) -> Result<(), ContractManagerErr> {
+    const DUST_OUTPUTS_IN_STAKE_TX: u64 = 2;
+    const DUST_OUTPUTS_IN_CLAIM_TX: u64 = 3;
+
+    let input_amount = OPERATOR_FUNDS
+        .checked_sub(SEGWIT_MIN_AMOUNT * (DUST_OUTPUTS_IN_CLAIM_TX + DUST_OUTPUTS_IN_STAKE_TX))
```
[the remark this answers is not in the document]
Rajil1213: These have been refactored out.

crates/duty-tracker/src/predicates.rs
```diff
@@ -195,10 +195,18 @@ pub(crate) fn parse_strata_checkpoint(
...
             }
-            if let Ok(signed_batch_checkpoint) =
+
+            let Ok(signed_checkpoint) =
                 borsh::from_slice::<SignedCheckpoint>(inscription[0].data())
-            {
-                return Some(signed_batch_checkpoint.into());
+            else {
+                return None;
+            };
```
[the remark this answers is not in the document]
Rajil1213: Oh, neat!

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1558,11 +1573,14 @@ impl ContractSM {
                 fulfiller,
                 active_graph,
                 recipient,
+                stake_txid,
                 ..
             } => {
                 // TODO(proofofkeags): we need to verify that this is bound properly to the correct
                 // operator.
                 let cfg = self.cfg();
+                let deposit_txid = self.cfg.deposit_tx.compute_txid();
```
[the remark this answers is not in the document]
Rajil1213: Oh, I'll make that change although computationally both are equivalent.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -979,25 +979,31 @@ impl ContractSM {
...
+                info!(%deposit_txid, "updating contract state to deposited");
+                self.state.state = ContractState::Deposited {
+                    peg_out_graphs,
+                    claim_txids,
+                }
+            }
+            ContractState::Deposited { .. } => {
+                // somebody else may have deposited already.
+                info!("contract already in deposited state");
+            }
```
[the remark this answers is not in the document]
Rajil1213: This is actually not weird.

All nodes compute the deposit transaction independently. Whichever node computes the aggregated root sig first will broadcast the transaction first. Another node may notice this transaction on chain before they can broadcast it on their own.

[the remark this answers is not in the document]
Rajil1213: I can definitely add a note there with some further clarification if that helps.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1420,6 +1431,11 @@ impl ContractSM {
                     },
                 )
             }
+            ContractState::Deposited { .. } => {
+                // somebody else may have deposited already.
+                info!("contract already in deposited state, skipping root signature generation");
+                Ok(None)
+            }
```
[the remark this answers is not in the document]
Rajil1213: Yes. I'll update it.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1384,6 +1390,11 @@ impl ContractSM {
                     },
                 )
             }
+            ContractState::Deposited { .. } => {
+                // somebody else may have deposited already.
+                info!("contract already in deposited state, skipping root nonce generation");
+                Ok(None)
+            }
```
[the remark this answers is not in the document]
Rajil1213: Yes. I'll update it.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1542,7 +1542,8 @@ impl ContractSM {
             )));
         }
 
-        match std::mem::replace(&mut self.state.state, ContractState::Resolved {}) {
+        let current = std::mem::replace(&mut self.state.state, ContractState::Resolved {});
+        match &current {
```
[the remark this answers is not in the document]
Rajil1213: Turns out borrowing here is actually a bug. It was changed to take ownership in an earlier commit.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -556,56 +559,65 @@ impl ContractManagerCtx {
...
+
+                        let pov_op_p2p_key = self.cfg.operator_table.pov_op_key();
+                        let stake_index = entry.idx();
+                        let Ok(Some(stake_tx)) = self
+                            .state
+                            .stake_chains
+                            .stake_tx(pov_op_p2p_key, stake_index as usize)
+                        else {
+                            warn!(%stake_index, %pov_op_p2p_key, "deposit assigned but stake chain data missing");
+                            continue;
```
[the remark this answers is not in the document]
Rajil1213: May be. But this is more of a side-effect of the database design/api. The data for an arbitrary operator, stake_index combo may or may not exist in the database, so it has to return an `Option<T>` but in this case, it presents as this `else` clause. I should probably bump it to an `error` though, wdyt?

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -416,7 +416,10 @@ impl ContractManagerCtx {
 
         for tx in block.txdata {
             let assignment_duties = self.process_assignments(&tx).await?;
-            duties.extend(assignment_duties.into_iter());
+            if !assignment_duties.is_empty() {
+                info!(num_duties=%assignment_duties.len(), "queueing assignment duties");
+                duties.extend(assignment_duties.into_iter());
```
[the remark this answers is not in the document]
Rajil1213: Right. Will fix.

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
```diff
@@ -230,28 +230,20 @@ pub(crate) async fn handle_publish_claim(
...
     let wots_client = s2_client.wots_signer();
 
-    let OutPoint {
-        txid: prestake_txid,
-        vout: prestake_vout,
-    } = output_handles.db.get_pre_stake(pov_idx).await?.ok_or(
-        StakeChainErr::StakeSetupDataNotFound(cfg.operator_table.pov_op_key().clone()),
-    )?;
-
+    const VOUT: u32 = 0;
```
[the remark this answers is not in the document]
Rajil1213: This is the deposit vout. Will rename.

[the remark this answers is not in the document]
Rajil1213: It is irrelevant though. It just has to be the same in both the deposit and withdrawal executors.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Yes. I definitely don't feel comfortable passing this map through.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: No, if we pass in a non-challenge tx to the `process_challenge_confirmation`, it just returns `Ok(None)`. The only time it returns a duty is if it observes a challenge transaction.

We cannot return a hard error from `process_challenge_transaction` because it may receive an optimistic payout tx. An alternative is to check the txid outside the `process_*` functions and then, convert the `Ok(None)` into an `Err<T>` but it will look weird.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: What would that construction look like? Iterators are still lazy in rust so maybe the compiler does optimize this a bit.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: We need to do this because of the `mem::replace` hack. Otherwise, we'll be returning an `Ok<T>` but mutating the CSM state to `Resolved`.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Maybe I'm doing something wrong but the borrow checker does not like this. We are moving the internals of the variant out (the very reason for the `mem::replace` hack) so we need to clone it outside.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: See my earlier comment.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Changed it back!

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Ended up reverting the change so that it does not violate the CSM property.

crates/duty-tracker/src/executors/deposit.rs
Rajil1213: No, previously the database state was being updated _before_ the in-memory state because this call was being made in the deposit executor. I removed it so that we only update the database in the CSM (both in-memory and persistent state).

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/a0fd9ae74933ae2d8566ff283f37acc22aee657b.

crates/duty-tracker/src/executors.rs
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/41fa94260f5db08d0a23c92fe2dd642134322adc.

crates/duty-tracker/src/predicates.rs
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/302646cd974250ea36fe4e05a04e18bdec344126.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/192ecd8ca79e1baa052cf7fef4776655a04268e1.

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/aa0b7f6920cf9ef15badae1333385ccd75651429.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/d733649112014115f2349ebfc22edff70defeb3d.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/d733649112014115f2349ebfc22edff70defeb3d.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: In https://github.com/alpenlabs/strata-bridge/commit/d733649112014115f2349ebfc22edff70defeb3d.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Nice catch. Done in https://github.com/alpenlabs/strata-bridge/commit/daef2eb8d723dcac1fc1b443b3b1d8df5d025ff9.

crates/operator-wallet/src/lib.rs
```diff
@@ -150,12 +146,14 @@ impl OperatorWallet {
             .into_iter()
             .map(|lo| lo.outpoint)
             .collect();
+
         let mut tx_builder = self.general_wallet.build_tx();
         // DON'T spend any of the cpfp outputs
         tx_builder.unspendable(cpfp_utxos);
         tx_builder.fee_rate(fee_rate);
-        tx_builder.add_recipient(user_p2tr_address.script_pubkey(), amount);
+        tx_builder.add_recipient(user_script_pubkey, amount);
```
storopoli, line 154: Just an FYI, `bdk_wallet` 1.2.0 relaxes the constraints around `add_recipient`: https://github.com/bitcoindevkit/bdk/pull/1841

crates/secret-service-client/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 bitcoin.workspace = true
+bitvm.workspace = true
```
storopoli, line 8: I wish we didn't had to do this. But yeah fine for now.

crates/secret-service-proto/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 bitcoin.workspace = true
+bitvm.workspace = true
```
storopoli, line 8: I wish we didn't had to do this. But yeah fine for now.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -345,25 +311,65 @@ impl StakeTx {
...
+        self.compute_sighash_with_prevouts(prevouts)
+    }
 
-        let prevouts = self
+    /// Generates the transaction message sighash for a stake transaction.
+    ///
+    /// # CAUTION
+    ///
+    /// This should only be invoked for stake_index >= 1, i.e. from at least second stake
+    /// transaction.
```
storopoli, line 339: What if we added an `idx` field to `StakeTx` and then did an `assert!` here (or even a `debug_assert!`)?

crates/tx-graph/src/transactions/post_assert.rs
```diff
@@ -26,7 +26,8 @@ pub struct PostAssertTxData {
     pub deposit_txid: Txid,
 }
 
-pub(crate) const NUM_POST_ASSERT_INPUTS: usize = NUM_ASSERT_DATA_TX;
+/// The number of inputs that require an $N-of-N$ signature in the [`PostAssertTx`].
```
storopoli, line 29:
```suggestion
/// The number of inputs that require an $N$-of-$N$ signature in the [`PostAssertTx`].
```

crates/tx-graph/src/transactions/slash_stake.rs
```diff
@@ -29,6 +29,7 @@ pub struct SlashStakeData {
     pub network: Network,
 }
 
+/// The number of inputs that require an $N-of-N$ signature in the [`SlashStakeTx`].
```
storopoli, line 32:
```suggestion
/// The number of inputs that require an $N$-of-$N$ signature in the [`SlashStakeTx`].
```

storopoli (changes requested):
minor nits and agree with Keagan's review

crates/stake-chain/src/transactions/stake.rs
Rajil1213: We could but I don't know if that is any better -- we'd have panics either way. The only difference will be when we invoke this function rather than when we broadcast the transaction, and both those things happen at about the same time.

crates/bridge-proof/protocol/src/tx_info.rs
```diff
@@ -70,40 +70,72 @@ pub(crate) struct WithdrawalInfo {
...
     let withdrawal_fulfillment_output = &tx.output[0];
     let withdrawal_metadata_output = &tx.output[1];
 
     let metadata_script = withdrawal_metadata_output.script_pubkey.as_bytes();
-    const EXPECTED_METADATA_SIZE: usize = 2 + 4 + 4 + 32; // OP_RETURN + OP_PUSHBYTES + operator_id + deposit_id + deposit_txid
-    if metadata_script.len() != EXPECTED_METADATA_SIZE {
+
+    const OP_RETURN_INSTRUCTION_SIZE: usize = 2; // OP_RETURN + OP_PUSHBYTES
+    const OPERATOR_IDX_SIZE: usize = 4;
+    const DEPOSIT_IDX_SIZE: usize = 4;
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/6f450300f48522adc66613027a455c28f233e5f3.

crates/bridge-proof/protocol/src/tx_info.rs
```diff
@@ -70,40 +70,72 @@ pub(crate) struct WithdrawalInfo {
...
+    const DEPOSIT_TXID_SIZE: usize = 32;
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/6f450300f48522adc66613027a455c28f233e5f3.

crates/bridge-proof/protocol/src/tx_info.rs
```diff
@@ -70,40 +70,72 @@ pub(crate) struct WithdrawalInfo {
...
 
     let withdrawal_fulfillment_output = &tx.output[0];
     let withdrawal_metadata_output = &tx.output[1];
 
     let metadata_script = withdrawal_metadata_output.script_pubkey.as_bytes();
-    const EXPECTED_METADATA_SIZE: usize = 2 + 4 + 4 + 32; // OP_RETURN + OP_PUSHBYTES + operator_id + deposit_id + deposit_txid
-    if metadata_script.len() != EXPECTED_METADATA_SIZE {
+
+    const OP_RETURN_INSTRUCTION_SIZE: usize = 2; // OP_RETURN + OP_PUSHBYTES
+    const OPERATOR_IDX_SIZE: usize = 4;
```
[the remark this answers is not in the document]
Rajil1213: Nice suggestion. Done in https://github.com/alpenlabs/strata-bridge/commit/6f450300f48522adc66613027a455c28f233e5f3.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -2,53 +2,32 @@
...
-    sighash::{Prevouts, SighashCache},
-    taproot, Block, FeeRate, Network, OutPoint, Psbt, TapSighashType, Transaction, Txid,
+    hashes::Hash, hex::DisplayHex, Address, Block, Network, OutPoint, ScriptBuf, Transaction, Txid,
 };
 use bitcoind_async_client::{client::Client as BitcoinClient, traits::Reader};
-use bitvm::chunk::api::{NUM_HASH, NUM_PUBS, NUM_U256};
 use btc_notify::client::{BlockStatus, BtcZmqClient};
 use futures::{
-    future::{self, join3, join_all},
+    future::{self},
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/6f450300f48522adc66613027a455c28f233e5f3.

crates/tx-graph/src/transactions/slash_stake.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/6f450300f48522adc66613027a455c28f233e5f3

storopoli (dismissed):
ACK 6f450300f48522adc66613027a455c28f233e5f3

crates/stake-chain/src/transactions/stake.rs
storopoli: Cc @ProofOfKeags who might have a good argument here with enforcing invariants instead of warning about them in docstrings.

Rajil1213: That's the thing -- adding a runtime assertion does not enforce the invariant any more than the docstring.

The only way to really enforce it is to have separate types for the first and subsequent stake txs.

storopoli: Yes, agreed. Maybe add a TODO then or a ticket?

Rajil1213: Okay, I'm cooking up something.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
ProofOfKeags: Yeah it's just nice to delegate things to the subcomputations responsible for actually figuring it out.

crates/duty-tracker/src/contract_manager.rs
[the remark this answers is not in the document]
ProofOfKeags: it's fine as is. I'll do a knowledge sharing session post-launch on functional decomposition and we can use this as an example.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: we don't need to change it, it's not an actual issue, but just a teaching moment.

`have_all_graphs` can fail and you could sidestep having to compute `have_all_partials_for_all_graphs`. To do that here using the laziness of iterators you'd have to get clever with the iterator's `chain` operator most likely to encode both conditions as a single stream. Laziness a la Haskell makes the code you've written optimal with respect to skipping computation that isn't needed since it starts computation at the end, evaluating && in such a way that if `have_all_graphs` is false, then it won't bother doing `have_all_partials_for_all_graphs`

Secondly, it occurs to me that there is probably a metadata element on the map that will give you its entry count. It is very possible that the iterator count is O(N), but maybe it's more clever than that.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
Rajil1213: Maybe it has to something to do with the other issue with burial depths not working but I was definitely seeing this issue when I ran it. I'll try to revert this change and see if it works without this now.

crates/duty-tracker/src/contract_manager.rs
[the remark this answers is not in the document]
Rajil1213: Nice. Looking forward to it (both the `post-launch` and the session 😆).

crates/duty-tracker/src/contract_manager.rs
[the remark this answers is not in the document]
Rajil1213: Forgot to mention that this is fixed too https://github.com/alpenlabs/strata-bridge/commit/daef2eb8d723dcac1fc1b443b3b1d8df5d025ff9.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: I think you are. match current destroys current but if you rebind it to `current@ContractState::Deposited {..}` you should not need to be cloned to move it back into state.

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -121,12 +121,11 @@ impl TxDriver {
     }
 
     /// Instructs the TxDriver to drive a new transaction to confirmation by the supplied deadline.
-    pub async fn drive(&self, tx: Transaction, _deadline: usize) -> Result<(), DriveErr> {
+    pub async fn drive(&self, tx: Transaction) -> Result<(), DriveErr> {
```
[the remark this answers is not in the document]
Rajil1213: Just noticed that the doc string has not been updated on this.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: if an event cannot be consumed by the CSM it should give back an error. if it does get consumed by the CSM it should not have the same state prior. Not all errors need to be fatal but semantically there's no difference between rejecting an event because it has the wrong internal state or rejecting an event because the event doesn't apply to the machine. Either way the error semantics should be about whether or not the event was accepted or rejected. We can annotate it with different reasons still if we use errors.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: The call to `count()` is `O(N)`. It iterates through the iterator to get the `count()`. But that `N` is the number of operators so it shouldn't immediately cause any issues.

As for the laziness optimization, I'm hoping the compiler is smart enough to make that optimization. I'm pretty sure it would do that if I didn't have the two variables and inlined both of them. I'm not sure if that is still true here.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
ProofOfKeags: Yeah I'm just trying to say that even if we can swallow it, it qualifies as weirdness. It should not happen. If it happens, we do not understand how our system behaves.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Rebinding works but since we need to move its members (`peg_out_graphs`, etc.), it complains that we're binding to a partially moved value. The only solution is to add a `ref` and if we're doing that, there's no point in doing `mem::replace`. Could you give it a shot as well? It may well be that I'm missing something obvious here.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: yeah depends on how good the inline heuristics are. Inlining is not safe *in general* in the presence of side-effecting functions and so the compiler has to reason about those and reasoning about it is quite difficult. Given that there's no `#[pure]` pragma or similar, it would have to annotate functions with IO or that contained Mutexes anywhere in its reference stack to know it can't.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
Rajil1213: I've removed this and it is still working!

crates/stake-chain/src/transactions/stake.rs
Rajil1213: Okay check this out: https://github.com/alpenlabs/strata-bridge/commit/4fb3c310e338f1ff12e156b5140536895f3b3a1e

I've completely overhauled the stake tx API so that it distinguishes between the first one and the rest.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: I've copied your comments over to the code and also planted flags in every place where we expect to return a specific error in the future: https://github.com/alpenlabs/strata-bridge/commit/61d420bb202a170260a550579d3505b43db4ea61.

crates/duty-tracker/src/tx_driver.rs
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/50e61a1dbab851d785d7d080e36f027de983c07e.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -821,6 +821,35 @@ pub struct ContractSM {
...
+            debug!(reimbursement_key=%$pog_input.operator_pubkey, %stake_txid, "retrieving peg out graph from cache");
+            pog.clone()
+        } else {
+            debug!(reimbursement_key=%$pog_input.operator_pubkey, %stake_txid, "generating and caching peg out graph");
+            let pog = $self.cfg.build_graph($pog_input);
+            $self.pog.insert(stake_txid, pog.clone());
+            pog
+        }
+    }}
+}
```
ProofOfKeags, line 851: I really very strongly recommend against doing this.

crates/stake-chain/src/transactions/stake.rs
ProofOfKeags: This commit is amazing.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: Is this being tracked by STR-1343? If so I think we can resolve this conversation and deal with it in a follow on PR?

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -57,12 +57,53 @@ impl std::hash::Hash for StakeTxData {
...
 pub const NUM_STAKE_TX_INPUTS: usize = 2;
 
+/// A marker for the first stake transaction.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct HEAD;
+
+/// A marker for the subsequent stake transactions i.e., all transactions in the stake chain other
+/// than the first.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TAIL;
```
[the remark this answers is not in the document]
Rajil1213: I know it from the [typestate](https://cliffle.com/blog/rust-typestate/) pattern in Rust but didn't know it was related to dependent types. Good to know!

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: I could revert this commit if you feel that strongly about this. The alternative is to copy-paste this commit in the places that it is currently being used.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -88,17 +129,170 @@ pub const NUM_STAKE_TX_INPUTS: usize = 2;
 /// 4. A dust output, [`ConnectorCpfp`], for the operator to use as CPFP in future transactions that
 ///    spends this one.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct StakeTx {
+pub struct StakeTx<StakeTxType = HEAD> {
     /// The PSBT that contains the inputs and outputs for the transaction.
     pub psbt: Psbt,
 
-    /// The type of witness required to spend the inputs of this transaction.
+    /// The type of witness required to spend the inputs of this tran tssaction.
```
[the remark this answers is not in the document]
Rajil1213: 🤦

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Reverted in https://github.com/alpenlabs/strata-bridge/commit/e57ce8c51ee8ba2a7cd42a2213b7bcc624748e61.

crates/stake-chain/src/transactions/stake.rs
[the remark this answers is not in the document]
Rajil1213: Fixed in https://github.com/alpenlabs/strata-bridge/commit/e57ce8c51ee8ba2a7cd42a2213b7bcc624748e61.

crates/stake-chain/src/transactions/stake.rs
```diff
@@ -57,12 +54,56 @@ impl std::hash::Hash for StakeTxData {
...
+pub const NUM_STAKE_TX_INPUTS: usize = 2;
+
+/// A marker for the first stake transaction.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Head;
+
+/// A marker for the subsequent stake transactions i.e., all transactions in the stake chain other
+/// than the first.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Tail;
```
storopoli, line 67: These markers can be `Copy` no?

storopoli (approved):
ACK e0e3c9f3e4587fafb1d1fcb280fbc74fba55ec02

crates/stake-chain/src/transactions/stake.rs
Rajil1213: I could derive that but it has no benefits because it is only used as part of StakeTx which is not Copy.

storopoli (approved):
ACK https://github.com/alpenlabs/strata-bridge/commit/e0e3c9f3e4587fafb1d1fcb280fbc74fba55ec02

ProofOfKeags (approved):

PR #138, opened 2025-05-08
Only RPC sync to the buried height

ProofOfKeags:
## Description

This PR makes it such that we don't treat the RPC notion of "current" differently from the zmq notion of "current".

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Commits are atomic.

## Checklist

### DIFF
--- a/bin/alpen-bridge/src/config.rs
+++ b/bin/alpen-bridge/src/config.rs
@@ -13,13 +13,6 @@ use strata_bridge_db::persistent::config::DbConfig;
 /// bridge.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub(crate) struct Config {
-    /// The number of confirmations required for a transaction to be considered final.
-    ///
-    /// This is not consensus-critical as difference in what value is set by individual bridge node
-    /// operators will only cause delays while exchanging information but will not halt the
-    /// functioning of the bridge.
-    pub finality_depth: u8,
-
     /// The directory to store all the data in.
     pub datadir: PathBuf,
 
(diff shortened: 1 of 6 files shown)


storopoli (dismissed):
ACK d0617abed184e5df205869196f9dd452c60e1b93

Need to fix docs CI check

Rajil1213 (approved):
It'd be great if we could actually unit-test as well but the changes themselves look good. I'm assuming that these changed have been tested locally?

Rajil1213 (changes requested):
Just realized that our `config` (see `config.toml` file) already has a parameter named `finality_depth`. Since we are applying this same config to both the zmq and rpc interfaces, I recommend updating the config file to remove the `bury_depth` field from the `zmq` and use the root field instead (I'm fine with removing the root field and keeping the `bury_depth` as well since technically, that is what the code is doing).

storopoli:
> Just realized that our `config` (see `config.toml` file) already has a paremeter named `finality_depth`. Since we are applying this same config to both the zmq and rpc interfaces, I recommend updating the config file to remove the `bury_depth` field from the `zmq` and use the root field instead (I'm fine with removing the root field and keeping the `bury_depth` as well since technically, that is what the code is doing).

Nice catch!

ProofOfKeags:
Yeah that seems better.

storopoli:
still needs to fix the `docs` CI failure: 
image

Rajil1213 (approved):

storopoli (approved):
ACK 0481c5487c8430854cffda7222d9478dc6d69de6

PR #143, opened 2025-05-14
STR-1343 commit aggregated sig to state and serve via rpc

Rajil1213:
## Description

This PR:

* adds a synthetic event `AggregateSigs` to signify that all the partials have been aggregated.
* adds a duty `CommitSig` to put the partials in the s2 and get the aggregated signatures.
* updates the STF to process graph signatures to also verify the partials as they are received using the `musig2` crate directly.
* updates the STF to process graph signatures to produce the `CommitSig` duty if all the partials have been collected.
* adds another ouroboros channel to self-feed the `AggregateSig` event that reuses the `process_contract_event` STF.
* updates all states to hold the aggregated signatures.
* updates the duty to publish payout optimistic to use the aggregated signature from the graph.
* adds an RPC trait to serve signatures for challenge and disprove transactions for a particular claim.
* updates the stake chain state machine to hold the stake data in a BTreeMap instead of an IndexSet to make the stake index explicit.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

RPC has not been tested yet because of the state update issue.
All commits are atomic.

## Checklist

## Related Issues

Closes STR-1343


crates/rpc/src/traits.rs
```diff
@@ -61,3 +61,15 @@ pub trait StrataBridgeMonitoringApi {
     #[method(name = "claimInfo")]
     async fn get_claim_info(&self, claim_txid: Txid) -> RpcResult<RpcClaimInfo>;
 }
+
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "stratabridge"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "stratabridge"))]
+pub trait StrataBridgeDaApi {
```
storopoli, line 67: Why this is called `DaApi`?
DA means, at least for me, data availability. Or maybe "data admin"?
Also no docstrings?

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -269,7 +273,7 @@ impl StrataBridgeControlApiServer for BridgeRpc {
         // The user might care about their system time being incorrect.
         if current_time <= start_time {
             return Err(rpc_error(
-                -32_000,
+                ErrorCode::InternalError,
```
storopoli, line 276: Nice! Thanks! This is much cleaner!

storopoli (changes requested):

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -196,10 +209,18 @@ pub enum ContractState {
         /// This is a collection of nonces for all graphs and for all operators.
         graph_nonces: BTreeMap<Txid, BTreeMap<P2POperatorPubKey, PogMusigF<PubNonce>>>,
 
+        /// This is a collection of all the packed aggregated nonces for each graph.
+        ///
+        /// This is used to validate partial signatures as they are received from the p2p.
+        agg_nonces: BTreeMap<Txid, Vec<AggNonce>>,
+
         /// This is a collection of all partial signatures for all graphs (indexed by the claim
-        /// txid) and for all operators.
```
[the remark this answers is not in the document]
Rajil1213: That was a bad rebase. Nice catch!

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -477,6 +526,42 @@ impl ContractState {
...
+            ContractState::Fulfilled { claim_txids, .. } => claim_txids,
+            ContractState::Claimed { claim_txids, .. } => claim_txids,
+            ContractState::Challenged { claim_txids, .. } => claim_txids,
+            ContractState::Asserted { claim_txids, .. } => claim_txids,
+            ContractState::Disproved { .. } => &BTreeMap::new(),
+            ContractState::Resolved { .. } => &BTreeMap::new(),
+        };
+
+        claim_txids.values().copied().collect()
+    }
```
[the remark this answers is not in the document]
Rajil1213: I think that comment was about the `deposit_request_txid` iirc? That is something that exists in the config. This isn't information that is available in the config. But @ProofOfKeags is definitely the source of truth on this, so I'll defer it to your judgment.

crates/rpc/src/traits.rs
Rajil1213: I'll add the docstring.

DA here does stand for Data Availability. The idea is that methods in this category make the data necessary to enable permissionless challenging available -- this mainly includes signatures as those are not available on chain.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -168,6 +178,9 @@ pub enum ContractEvent {
 /// - Claimed -> Asserted
 /// - Asserted -> Disproved
 /// - Asserted -> Resolved
+///
+/// The transaction ID used to index transactions and other data is the transaction ID of the claim
+/// transaction.
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/fbb7199eacb3a0cac81a61e88c8ec4657a9a2a48.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -530,6 +615,21 @@ pub enum OperatorDuty {
         pog_sighashes: PogMusigF<Message>,
     },
 
+    /// Instruscts us to commit the aggregated signatures to state.
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/fbb7199eacb3a0cac81a61e88c8ec4657a9a2a48.

crates/rpc/src/traits.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/214cca53258eec921a674b9cd9b7f6ecc5c43d01.

storopoli: Oh nice! Now it makes sense. Please add this exactly rationale to the docstring!

storopoli (approved):
ACK 214cca53258eec921a674b9cd9b7f6ecc5c43d01

Just pending the `claim_txids()` either in `ContractState` or `ContractCfg`. I think it should go in the `ContractCfg`.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
ProofOfKeags: Claim txids are unfortunately a function of state. We don't know them until we have the deposit setup messages so we can't make them part of config. If we break up the CSM into smaller sub automata we could make them part of config for the later ones.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -193,10 +193,14 @@ impl ContractManager {
             let (ouroboros_sender, mut ouroboros_receiver) = broadcast::channel(OUROBOROS_CAP);
             let msg_handler = MessageHandler::new(p2p_handle.clone(), ouroboros_sender);
 
+            let (ouroboros_event_sender, mut ouroboros_event_receiver) =
```
ProofOfKeags, line 196: nit: I'd like to call this `synthetic_event_*`

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1455,6 +1705,33 @@ impl ContractSM {
         }
     }
 
+    fn process_aggregate_sigs(
```
ProofOfKeags, line 1638: ok if I understand this properly this is just loading the aggregate signatures back into the CSM. It makes me wonder though... do we need the signatures back in the CSM? Do we use them later?

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -270,6 +274,25 @@ impl ContractManager {
                 tokio::select! {
                     biased; // follow the same order as specified below
 
+                    ouroboros_event = ouroboros_event_receiver.recv() => {
+                        if let Some(ContractEvent::AggregateSigs(deposit_txid, agg_sigs)) = ouroboros_event {
+                            let contract = ctx.state.active_contracts.get_mut(&deposit_txid).expect("contract must exist in the state");
+
+                            match contract.process_contract_event(ContractEvent::AggregateSigs(deposit_txid, agg_sigs)) {
+                                Ok(ouroboros_duties) if !ouroboros_duties.is_empty() => duties.extend(ouroboros_duties),
+                                Ok(ouroboros_duties) => { trace!(?ouroboros_duties, "got no duties when processing ouroboros event to aggregate signatures")},
```
ProofOfKeags, line 284: there's some weird conceptualization going on here. Events are descriptions of ... well ... events. You've formulated this as a command of sorts, if I'm understanding correctly. This trace suggests that this event is instructing the CSM to aggregate signatures but from what I understand in the actual event handler all it's doing is loading the aggregated signatures back into the CSM for later, though I can't tell where we use them later.

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -633,6 +635,47 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
     }
 }
 
+#[async_trait]
+impl StrataBridgeDaApiServer for BridgeRpc {
```
ProofOfKeags, line 647: so for a lot of the other RPC stuff we are going straight to disk. Perhaps we can sidestep having to re-load the aggregate signatures back into the CSM here and instead go to disk to find them?

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -299,7 +303,7 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
         // Avoid DoS attacks by just returning an error if the public key is invalid
         if conversion.is_err() {
             return Err(rpc_error(
-                -32001,
+                ErrorCode::InvalidRequest,
```
ProofOfKeags, line 306: 😘👌🏼

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -1434,7 +1434,7 @@ async fn execute_duty(
                     claim_txid,
                     stake_txid,
                     stake_index,
-                    partials,
+                    *agg_sigs,
```
ProofOfKeags, line 1438: so this is an approach, the other approach would be to ditch the sigs altogether and go straight to disk for them. I understand how we're at least using these signatures now.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1927,56 +1933,6 @@ impl ContractSM {
         Ok(duty)
     }
 
-    /// Transposes an array of partials for each input per operator into an array of partials for
-    /// each operator per input.
-    ///
-    /// It accepts a map of graph partials such that each map holds the partials for an operator for
-    /// all inputs and then changes them so that the end result is an array of partials from all
-    /// operators for each input such that they can be aggregated on a per input basis.
-    fn transpose_partials<const NUM_INPUTS: usize>(
```
ProofOfKeags, line 1659: 🔪🩸

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -91,12 +91,14 @@ impl TxDriver {
...
                                 }
                             }
-                            btc_notify::client::TxStatus::Buried { .. } => {
+                            btc_notify::client::TxStatus::Buried { height, .. } => {
                                 // Since our responsibility ends at block inclusion we will send an
                                 // answer on the response channel now. It is the API caller's
                                 // responsibility for handling reorgs after inclusion.
                                 match active_jobs.remove(&event.rawtx.compute_txid()) {
                                     Some(job) => {
+                                        info!(txid=%event.rawtx.compute_txid(), %height, "transaction confirmed in block");
```
ProofOfKeags, line 100: 🫡

ProofOfKeags (changes requested):

crates/duty-tracker/src/contract_manager.rs
Rajil1213: You're right. This needs to be rephrased. The duty that should be generated is `PublishRootNonce`. The aggregated signature has already been committed inside the `process_contract_event`.

It's used when finalizing the pog transactions.  You can check the fulfiller duty for payout optimistic.

bin/alpen-bridge/src/rpc_server.rs
Rajil1213: Unfortunately, it's not in "one place" in the disk. It's part of the state.

Rajil1213: And since that state should already be available in memory, it just makes sense to use that.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: I could do that. But I don't know if that's better than the approach that we currently have. At some point, I do want us to normalize the database schema for the ContractState and this is as good a place as any but I'd rather cut a ticket for this and get started with the contested withdrawal execution.

crates/duty-tracker/src/tx_driver.rs
Rajil1213: Now that you've pointed it out, maybe I should compute the txid just once.

Rajil1213: Amended the related commit https://github.com/alpenlabs/strata-bridge/commit/f3f84240007145a9b17942e477fabd2582e8b344.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: 🫡

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: I think you've answered this question yourself but yes, we use it during withdrawal execution.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/65b2307dfeb25600246dcd6d2a384920de00edb3.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/cc5e3b2e1e4e3237f0d85baf1861509e98a6ecaa.

crates/duty-tracker/src/contract_manager.rs
ProofOfKeags: I should have mentioned this as well but I think the synthetic event enum should be distinct from the contract event enum. Similar to how we have a single p2p event stream, I want a single synthetic event stream, and that might span multiple different functions and I don't want all contract events to be sendable to the synthetic event stream.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -280,8 +280,8 @@ impl ContractManager {
 
                             info!(%deposit_txid, "committing aggregate signatures");
                             match contract.process_contract_event(ContractEvent::AggregateSigs(deposit_txid, agg_sigs)) {
-                                Ok(ouroboros_duties) => { trace!(?ouroboros_duties, "got no duties when processing ouroboros event to aggregate signatures")},
                                 Ok(synthetic_duties) if !synthetic_duties.is_empty() => duties.extend(synthetic_duties),
+                                Ok(synthetic_duties) => { trace!(?synthetic_duties, "got no duties when processing synthetic event to commit signatures")},
```
ProofOfKeags, line 284: minor conceptual nit: the duties themselves are never synthetic. What is synthetic is the event. It is synthetic because it is not a reaction to any stimulus from the environment, but originates from inside of the bridge. Duties are actions we queue in response to events (of all kinds). You could say that duties that are generated in response to synthetic events are synthetic duties, as I believe you are doing here. However, I'm not sure if that distinction is useful. It's up to you. I mention it because it occurs to me. Nothing necessarily needs to change about this code.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Yeah that's a good idea. I sort of took the shortest path available here. Will create a separate enum.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: > You could say that duties that are generated in response to synthetic events are synthetic duties, as I believe you are doing here.

Yeah that is what I meant.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/fb53acb0bd25a463c162bc9243aeb026ae90be0e.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Renamed to `synthetic_event_duties` in https://github.com/alpenlabs/strata-bridge/commit/fb53acb0bd25a463c162bc9243aeb026ae90be0e (not sure if that's better).

ProofOfKeags (approved):

PR #146, opened 2025-05-15
STR-1342: `io` error when loading contracts from disk upon restart

storopoli:
## Description

- Makes all `ContractPersister` and RPC server DB operations robust by using `execute_with_retries` from `strata-bridge-db` crate.
- Commits all columns of the `contracts` table when serializing active contracts with the `ContractPersister`.
- Could not get `bincode`, `ciborium`, `postcard` or `rkyv` to play nice with the huge `MachineState` struct (that has a huge enum `ContractState` inside). So we're JSONing the 🖕🏻 thing as a quick fix.
- Adds the remaining RPC trait instantiations that were missing somehow.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Tested with multiple deposits, stopping restart nodes, and wait long enough to have the deposit confirmed.

## Checklist

## Related Issues

STR-1342

### DIFF
--- a/.taplo.toml
+++ b/.taplo.toml
@@ -8,4 +8,5 @@ include = ["**/Cargo.toml"]
 keys = ["workspace.dependencies", "dependencies"]
 
 [rule.formatting]
+indent_string = "  "
 reorder_keys = true
--- a/bin/alpen-bridge/src/rpc_server.rs
+++ b/bin/alpen-bridge/src/rpc_server.rs
@@ -16,7 +16,13 @@ use libp2p::{identity::PublicKey as LibP2pPublicKey, PeerId};
 use secp256k1::Parity;
 use serde::Serialize;
 use sqlx::{query_as, FromRow};
-use strata_bridge_db::persistent::sqlite::SqliteDb;
+use strata_bridge_db::{
+    errors::DbError,
+    persistent::{
+        errors::StorageError,
+        sqlite::{execute_with_retries, SqliteDb},
+    },
+};
 use strata_bridge_primitives::operator_table::OperatorTable;
 use strata_bridge_rpc::{
     traits::{
@@ -213,6 +219,7 @@ impl BridgeRpc {
     fn start_cache_refresh_task(&self) {
         let db = self.db.clone();
         let cached_contracts = self.cached_contracts.clone();
+        let db_config = *self.db.config();
 
         // Clone the params we need before spawning the task
         let network = self.params.network;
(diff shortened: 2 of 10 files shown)


bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -228,83 +229,89 @@ impl BridgeRpc {
...
-                // Strive to always drop the lock as soon as possible to avoid blocking other tasks.
-                drop(cache_lock);
-
-                info!(%after_num_contracts, "RPC server Contracts cache initialized");
-            } else {
-                error!("Failed to initialize contracts cache");
-            }
-
-            // Periodic refresh
+            let _ = execute_with_retries::<_, _, ()>(&db_config, || async {
```
Rajil1213, line 238: This should really just wrap the database operation and not the high-level operation to update in-memory cache.

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -228,83 +229,89 @@ impl BridgeRpc {
...
-                    "#,
-                )
-                .fetch_all(db.pool())
-                .await
-                {
-                    Ok(contracts) => {
+                refresh_interval.tick().await;
+
+                // Each refresh is its own execute_with_retries call
+                let _ = execute_with_retries::<_, _, ()>(&db_config, || async {
```
Rajil1213, line 297: Same here. The only thing that should be retried is the database query. Having cache locks inside a closure like this  will make it harder to reason about.

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -155,7 +155,7 @@ impl ContractPersister {
         execute_with_retries(&self.config, || async {
             let _: SqliteQueryResult = sqlx::query(
                 r#"
-            UPDATE contracts SET state = ? WHERE deposit_txid = ?
+            INSERT OR REPLACE INTO contracts (state, deposit_txid) VALUES (?, ?)
```
Rajil1213: Is this change necessary? Plus I don't know if this actually works. My understanding is that not specifying all the columns during insertion means that the rest of the fields will be NULL or replaced by a default.

crates/duty-tracker/src/contract_persister.rs
```diff
@@ -148,28 +149,39 @@ impl ContractPersister {
...
     ) -> Result<(), ContractPersistErr> {
+        let deposit_tx_bytes = bincode::serialize(&deposit_tx)?;
+        let operator_table_bytes = bincode::serialize(&operator_table)?;
         let state_bytes = bincode::serialize(&state)?;
 
         execute_with_retries(&self.config, || async {
             let _: SqliteQueryResult = sqlx::query(
                 r#"
-            INSERT OR REPLACE INTO contracts (state, deposit_txid) VALUES (?, ?)
+            INSERT OR REPLACE INTO contracts (deposit_txid, deposit_idx, deposit_tx, operator_table, state) VALUES (?, ?, ?, ?, ?)
```
Rajil1213, line 164: Nevermind my previous comment about missing fields. It still feels unnecessary to me to have to commit all the data when it's only the state that has changed (unless not doing this causes things to not work).

.taplo.toml
```diff
@@ -8,4 +8,5 @@ include = ["**/Cargo.toml"]
 keys = ["workspace.dependencies", "dependencies"]
 
 [rule.formatting]
+indent_string = "  "
```
Rajil1213, line 11: Why was this necessary?

Rajil1213 (changes requested):
Just a few nits.

If this temporary fix ends up being non-temporary, we should definitely consider migrating to postgres which has dedicated JSON/JSONB types.

crates/duty-tracker/src/contract_persister.rs
storopoli: With respect to the previous comment and this one, if we don't include all the fields, then the `UPDATE` method only alters the `state` and somehow either removes it or it wasn't there to begin with the other colums like `deposit_idx`. This eventually blows up when querying the `contracts` table.
So, yeah, things break without this.

Rajil1213: That's really weird. Is that the case even when using JSON instead of bincode?

bin/alpen-bridge/src/rpc_server.rs
storopoli: Done in ea4aeedff19aeb88b7ebcacbe9beac347b779df4

Nice catch

crates/duty-tracker/src/contract_persister.rs
storopoli: Yes...
It is a `sqlx` error

Rajil1213 (approved):

migrations/20241113060253_init_schema.sql
```diff
@@ -6,7 +6,7 @@ CREATE TABLE IF NOT EXISTS contracts (
     deposit_idx INTEGER NOT NULL UNIQUE,    -- Index of the deposit in the stake chain
     deposit_tx BLOB NOT NULL,               -- Serialized with bincode
     operator_table BLOB NOT NULL,           -- Serialized with bincode
-    state BLOB NOT NULL                     -- Serialized with bincode
+    state TEXT NOT NULL                     -- JSON
```
delbonis, line 9: oh godd

crates/duty-tracker/src/contract_persister.rs
ProofOfKeags: I refuse to believe this is true.

PR #147, opened 2025-05-16
Add proptest generators to select types in the bridge

ProofOfKeags:
## Description

This PR begins to add some tactically relevant proptest generators so that we can more effectively generate test data without having to go through the full process of integration testing via the docker compose cluster. The main thing we need to figure out is how we want to systematically organize and export these. The nature of these generators is that they *compose* so each crate that defines a type should also define a generator. From there we can use generators of component types as component generators of a type.

The trouble with this is that it forces us to export the proptest generators as part of the main package, making `proptest` a regular dependency as opposed to a dev-dependency. In my mind this tradeoff is well worth it, but an alternative is to ship parallel crates that define these generators. The downside of going that route is that we will wall ourselves off from being able to define canonical generators (via the `proptest::Arbitrary` trait) due to orphan instance rules.

Ideally we'd be able to export the test module in `cfg(test)` and have other `cfg(test)` code depend on that exported test module but idk how the transitivity of `cfg(test)` works. Is it only on the package in question or is the entire package tree compiled with `cfg(test)`?

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The main thing that should be evaluated in this PR is the overall strategy with how we group these functions into modules, those modules' visibility, and the impact it has on cross-package generator composition.

## Checklist


crates/primitives/src/operator_table.rs
```diff
@@ -138,3 +138,57 @@ impl OperatorTable {
...
+            sk in any::<[u8; 32]>()
+                .no_shrink()
+                .prop_filter_map(
+                    "invalid secret key",
+                    |bs| secp256k1::SecretKey::from_slice(&bs).ok()
+                )
+        ) -> secp256k1::PublicKey {
+            sk.public_key(secp256k1::SECP256K1)
+        }
+    }
```
storopoli, line 165: You might need to negate this SKs if it generates odd PKs.

crates/tx-graph/src/transactions/deposit.rs
```diff
@@ -557,3 +557,50 @@ mod tests {
...
+        fn arb_txid()(bs in any::<[u8; 32]>()) -> Txid {
+            Txid::from_raw_hash(*sha256d::Hash::from_bytes_ref(&bs))
+        }
+    }
+
+    prop_compose! {
+        fn arb_outpoint()(txid in arb_txid(), vout in 0..100u32) -> OutPoint {
+            OutPoint { txid, vout }
+        }
+    }
```
storopoli, line 581: The deposit is always the first vout `0` in the spec.

crates/tx-graph/src/transactions/deposit.rs
```diff
@@ -557,3 +557,50 @@ mod tests {
...
+
+    prop_compose! {
+        pub fn arb_deposit_request_data(deposit_amount: Amount)(
+            deposit_request_outpoint in arb_outpoint(),
+            stake_index in 1..100u32,
+            ee_address in proptest::collection::vec(any::<u8>(), 20),
+            excess_deposit_amount in 100000..500000u64,
```
storopoli, line 588:
```suggestion
            excess_deposit_amount in 100_000..500_000u64,
```

ZER000000ES

crates/tx-graph/src/transactions/deposit.rs
```diff
@@ -557,3 +557,50 @@ mod tests {
...
+            x_only_public_key in arb_btc_key().prop_map(|x|x.x_only_public_key().0),
+        ) -> DepositRequestData {
+
+            let script = format!("and_v(v:pk({}),older({}))", x_only_public_key.clone(), 1008);
```
storopoli, line 592: `1_008` should be the constant in https://github.com/alpenlabs/strata-bridge/blob/dbf615860037dc10eb7b5c3a43159b8fcf120827/crates/params/src/default.rs#L33-L35

storopoli (changes requested):
This is really nice, excited to use it!

crates/primitives/src/operator_table.rs
Rajil1213: Yeah, you should just use the `EvenSecretKey` type from the primitives.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2911,12 +2912,19 @@ pub mod prop_tests {
     }
 
     proptest! {
+        #![proptest_config(ProptestConfig::with_cases(0))] // This still does 1 test case. It's weird.
```
storopoli, line 2915: Very strange, indeed!
A wild hunch is that maybe somewhere in `proptest` codebase we have something like:

```rs
fn main() {
    let n_cases = 0;
    for _ in 0..=n_cases {
        // DO STUFF
    }
}
```

This runs once, check https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b0b1f041e0aaa462bea4ad5349f2b159

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2911,12 +2912,19 @@ pub mod prop_tests {
...
         #[test]
-        fn state_serialization_invertible(state in arb_machine_state()) {
+        fn state_serialization_invertible(state in arb_machine_state().no_shrink()) {
+            let mut time = std::time::Instant::now();
```
storopoli, line 2918: full import?

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2911,12 +2912,19 @@ pub mod prop_tests {
...
+            println!("serializing machine state");
             match bincode::serialize(&state) {
                 Ok(serialized) => {
+                    println!("serialization complete. time taken: {:?}", std::time::Instant::now().duration_since(time));
+                    time = std::time::Instant::now();
```
storopoli, line 2923: full import?

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2911,12 +2912,19 @@ pub mod prop_tests {
...
+                    println!("deserializing machine state");
                     match bincode::deserialize(&serialized) {
                         Ok(deserialized) => {
+                            println!("deserialization complete. time taken: {:?}", std::time::Instant::now().duration_since(time));
```
storopoli, line 2927: full import?

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2911,12 +2912,19 @@ pub mod prop_tests {
...
     proptest! {
+        #![proptest_config(ProptestConfig::with_cases(0))] // This still does 1 test case. It's weird.
         #[test]
-        fn state_serialization_invertible(state in arb_machine_state()) {
+        fn state_serialization_invertible(state in arb_machine_state().no_shrink()) {
+            let mut time = std::time::Instant::now();
+            println!("serializing machine state");
             match bincode::serialize(&state) {
                 Ok(serialized) => {
+                    println!("serialization complete. time taken: {:?}", std::time::Instant::now().duration_since(time));
```
storopoli, line 2922: full import?

storopoli (changes requested):
This is awesome. Couple minor nits.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2746,3 +2726,222 @@ mod tests {
...
+                .p2p_keys()
+                .iter()
+                .map(|pk| arb_deposit_setup_from_operator(pk.clone()).boxed()).collect::<Vec<_>>(),
+            cfg in Just(cfg),
+            block_height in Just(block_height),
+        ) -> MachineState {
+            let abort_deadline = block_height + cfg.peg_out_graph_params.refund_delay as u64;
+            let mut csm = ContractSM::new(cfg, block_height, abort_deadline);
+
+            unsafe { std::env::set_var("ZKVM_MOCK", "true"); }
```
Rajil1213, line 2905: Instead of the generator making this choice for us, I think it's best to let whatever invokes this test bear the responsibility to set this environment variable. Is there a reason why this choice was made?

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -1276,8 +1256,8 @@ impl ContractSM {
         signer: P2POperatorPubKey,
         operator_pubkey: XOnlyPublicKey,
         new_stake_hash: sha256::Hash,
-        new_stake_tx: StakeTxKind,
-        new_wots_keys: WotsPublicKeys,
+        new_stake_txid: Txid,
```
Rajil1213, line 1259: That's nice. I'm kinda surprised I didn't notice that the stake txid is the only thing being used downstream.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -952,64 +970,26 @@ impl ContractSM {
     /// Builds a new ContractSM around a given deposit transaction.
     ///
     /// This will be constructible once we have a deposit request.
-    #[expect(clippy::too_many_arguments)]
```
Rajil1213, line 955: The earth heals every time we remove these. 😌

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: In some of the proptest source code I've glanced at, it looks like the number of cases is actually the number of successes. The runner runs the given test once (regardless of the `case` value) and then keeps running it until the number of successes reaches the number of cases specified.

Rajil1213 (changes requested):
Just some nits beyond what Jose has already said. I really like some of the refactors you've done as part of this work.

The only gripe I have is with the test organization. I don't feel very comfortable with files with >1k lines. I think it's almost time we split the `contract_state_machine.rs` to the states, duties, state machine and state transition functions (but of course, we can do that later).

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: Yeah it's because the code that relies on this being set is the graph generation code and that is invoked in the `arb_contract_state` generator. Generator sampling is done before the test case runs when you use the normal macros. I could unroll the proptest macro and set this prior to feeding the generator to the test case. I could also not have `arb_contract_state` at all and instead generate just the events, and then do the event processing in the main body of the test case. I agree that the generator shouldn't make this choice, but I also don't think this choice should be left up to magic environment variables whose consequences involve network calls to unknown places.

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: It's number of successes but also no failures. The successes are in contrast to discards. Some generated cases get discarded because they fail some precondition needed for the test to be valid.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -720,7 +720,7 @@ impl ContractManagerCtx {
     ) -> Result<Vec<OperatorDuty>, ContractManagerErr> {
         let mut duties = Vec::new();
 
-        match msg {
+        match msg.clone() {
```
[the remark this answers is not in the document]
ProofOfKeags: To preserve our ability to put the message in the log message *and* the error value in the error case. I'll look it over again and see if I can shrink the surface area though.

crates/primitives/src/wots.rs
```diff
@@ -106,6 +105,19 @@ impl<'de> Deserialize<'de> for Wots256PublicKey {
     }
 }
 
+impl Arbitrary for Wots256PublicKey {
+    type Parameters = ();
+
+    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+        any::<[u8; std::mem::size_of::<Wots256PublicKey>()]>()
+            .no_shrink()
+            .prop_map(|arr| unsafe { std::mem::transmute(arr) })
```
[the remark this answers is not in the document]
ProofOfKeags: can't transmute without it and I opted for transmuting because these generators runtimes meaningfully impact test generation. I can try backing off of it and doing some memcopying from the array into the different keys but even so it doubles the amount of stack allocation that happens and keep in mind that the wots keys are about 250k a piece

crates/primitives/src/wots.rs
```diff
@@ -307,6 +317,18 @@ impl<'de> Deserialize<'de> for Groth16PublicKeys {
         deserializer.deserialize_seq(Groth16PublicKeysVisitor)
     }
 }
+impl Arbitrary for Groth16PublicKeys {
+    type Parameters = ();
+
+    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+        any::<[u8; std::mem::size_of::<Groth16PublicKeys>()]>()
+            .no_shrink()
+            .prop_map(|arr| unsafe { std::mem::transmute(arr) })
```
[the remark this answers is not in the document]
ProofOfKeags: This one is actually way more important than the other one.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2770,9 +2778,23 @@ pub mod prop_tests {
...
+
+    // Generates a random 32 byte hash as a Txid`.
+    prop_compose! {
+        fn arb_txid()(bs in any::<[u8; 32]>()) -> Txid {
+            Txid::from_raw_hash(*sha256d::Hash::from_bytes_ref(&bs))
+        }
+    }
+
+    prop_compose! {
+        fn arb_hash()(bytes in any::<[u8; 32]>()) -> sha256::Hash {
```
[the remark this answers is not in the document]
ProofOfKeags: These shouldn't even be where they are. I source ripped them from the btc-notify package tests because we don't have canonical generation code for the primitive types. I think we should move them to their own package and import that package everywhere.

crates/primitives/src/wots.rs
```diff
@@ -42,6 +46,63 @@ impl From<Wots256PublicKey> for strata_p2p_types::Wots256PublicKey {
...
+            }
+
+            // Handle the case where input is a sequence (e.g., JSON array)
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let mut packed = [[0u8; 20]; 68];
+                for (key_idx, key) in packed.iter_mut().enumerate() {
+                    for (byte_idx, byte) in key.iter_mut().enumerate() {
```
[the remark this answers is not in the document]
ProofOfKeags: This was clearer since rust doesn't have general monad syntax. `for..in` is just `<-` for the list (iterator) monad.

crates/primitives/src/wots.rs
```diff
@@ -128,6 +189,120 @@ impl From<Groth16PublicKeys> for strata_p2p_types::Groth16PublicKeys {
...
+    {
+        // Create a visitor for our nested array
+        struct Groth16PublicKeysVisitor;
+
+        impl<'de> Visitor<'de> for Groth16PublicKeysVisitor {
+            type Value = Groth16PublicKeys;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str(
+                    "a flattened structure of type ([[[u8; 20]; 68]; NUM_PUBS], [[[u8; 20]; 68]; NUM_U256], [[[u8; 20]; 36]; NUM_HASHES])"
```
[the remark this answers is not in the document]
ProofOfKeags: I'll go ahead and steal the `key_width` code I wrote in s2

crates/primitives/src/wots.rs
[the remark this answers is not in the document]
ProofOfKeags: Also fwiw this code should age just fine (even in the presence of unsafe) so long as all the data in those key types is entropic.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: > Generator sampling is done before the test case runs when you use the normal macros.

Oh I didn't know that. In that case, I recommend leaving it there with a note.

> but I also don't think this choice should be left up to magic environment variables whose consequences involve network calls to unknown places.

It's slightly more subtle than that -- we're not making arbitrary network calls when an environment variable is set, it's actually the opposite. We're _preventing_ these network calls by setting the environment variable. I do agree that it's not ideal though. Sadly, this is still consistent with how most prover-related things are controlled when running things in simulation.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2874,4 +2874,28 @@ pub mod prop_tests {
             }
         }
     }
+
+    prop_compose! {
+        /// Generates a ContractState with all three DepositSetup messages.
+        pub fn arb_contract_state()(
+            cfg in arb_contract_cfg(),
+            block_height in 500_000..800_000u64,
```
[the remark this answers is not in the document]
ProofOfKeags: Can you elaborate more on what you mean here? Do you just want a test-wide constant that defines a range so all tests use the same one? I don't think it's that important since the consistency guarantees are either non-existent or *much* more tightly linked than using the same range everywhere.

crates/primitives/src/operator_table.rs
ProofOfKeags: d166e863c77a7926f58235eb19dedd02216c2432

crates/tx-graph/src/transactions/deposit.rs
ProofOfKeags: 6da7b335e4cd4a0cb226b115578900b0b4d5f777

crates/tx-graph/src/transactions/deposit.rs
ProofOfKeags: 4334f4b47645fac7ab3dda4ec1b8af02d2072488

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2770,9 +2778,23 @@ pub mod prop_tests {
         proof::RollupVerifyingKey,
     };
 
-    use super::ContractCfg;
+    use super::{ContractCfg, ContractEvent};
+
+    // Generates a random 32 byte hash as a Txid`.
```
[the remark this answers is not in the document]
ProofOfKeags: 93b4ff65a7c8b46cbbff7d446c684f5d98a0215a

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2770,9 +2778,23 @@ pub mod prop_tests {
...
+    }
+
+    prop_compose! {
+        fn arb_hash()(bytes in any::<[u8; 32]>()) -> sha256::Hash {
+            sha256::Hash::from_byte_array(bytes)
+        }
+    }
 
     prop_compose! {
+        /// Generates a random ContractCfg.
```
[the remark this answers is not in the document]
ProofOfKeags: 93b4ff65a7c8b46cbbff7d446c684f5d98a0215a

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2860,4 +2882,22 @@ pub mod prop_tests {
             }
         }
     }
+
+    prop_compose! {
+        /// Generates a random ContractEvent::DepositSetup {..}
```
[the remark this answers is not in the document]
ProofOfKeags: 93b4ff65a7c8b46cbbff7d446c684f5d98a0215a

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2874,4 +2874,28 @@ pub mod prop_tests {
             }
         }
     }
+
+    prop_compose! {
+        /// Generates a ContractState with all three DepositSetup messages.
```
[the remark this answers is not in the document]
ProofOfKeags: 93b4ff65a7c8b46cbbff7d446c684f5d98a0215a

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2828,7 +2837,7 @@ pub mod prop_tests {
                 deposit_amount: 1000000000,
```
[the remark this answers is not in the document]
ProofOfKeags: b68cba94bf49c42519f806f6d2ccac0f7dcf3ac5

crates/primitives/src/wots.rs
```diff
@@ -42,6 +46,63 @@ impl From<Wots256PublicKey> for strata_p2p_types::Wots256PublicKey {
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct Wots256PublicKeyVisitor;
+
+        impl<'de> Visitor<'de> for Wots256PublicKeyVisitor {
+            type Value = Wots256PublicKey;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("a flattened structure of type [[u8; 20]; 68]")
```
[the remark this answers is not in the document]
ProofOfKeags: 8c0177d24d0b216d80fc79982acb8ba6d7d33e90

crates/primitives/src/wots.rs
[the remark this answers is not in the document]
ProofOfKeags: 8c0177d24d0b216d80fc79982acb8ba6d7d33e90

crates/primitives/src/wots.rs
```diff
@@ -128,6 +189,120 @@ impl From<Groth16PublicKeys> for strata_p2p_types::Groth16PublicKeys {
...
+                    "a flattened structure of type ([[[u8; 20]; 68]; NUM_PUBS], [[[u8; 20]; 68]; NUM_U256], [[[u8; 20]; 36]; NUM_HASHES])"
+                )
+            }
+
+            // Handle the case where input is a sequence (e.g., JSON array)
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let mut public_inputs = [[[0u8; 20]; 68]; NUM_PUBS];
```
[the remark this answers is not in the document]
ProofOfKeags: 8c0177d24d0b216d80fc79982acb8ba6d7d33e90

crates/primitives/src/wots.rs
```diff
@@ -128,6 +189,120 @@ impl From<Groth16PublicKeys> for strata_p2p_types::Groth16PublicKeys {
...
+                                return Err(serde::de::Error::invalid_length(
+                                    (input_idx + 1) * (key_idx + 1) * (byte_idx + 1),
+                                    &self,
+                                ));
+                            }
+                        }
+                    }
+                }
+
+                let mut fqs = [[[0u8; 20]; 68]; NUM_U256];
```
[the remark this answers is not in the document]
ProofOfKeags: 8c0177d24d0b216d80fc79982acb8ba6d7d33e90

crates/primitives/src/wots.rs
```diff
@@ -128,6 +189,120 @@ impl From<Groth16PublicKeys> for strata_p2p_types::Groth16PublicKeys {
...
+                                return Err(serde::de::Error::invalid_length(
+                                    (fq_idx + 1) * (key_idx + 1) * (byte_idx + 1),
+                                    &self,
+                                ));
+                            }
+                        }
+                    }
+                }
+
+                let mut hashes = [[[0u8; 20]; 36]; NUM_HASH];
```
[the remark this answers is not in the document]
ProofOfKeags: 8c0177d24d0b216d80fc79982acb8ba6d7d33e90

crates/duty-tracker/src/contract_manager.rs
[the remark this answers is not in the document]
ProofOfKeags: 45840f391e9df3efa84a4ea760d05956d975ddc9

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: 6c3836e10b23372fc0b24cdcb2d7b55a7a5eb770

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: 6c3836e10b23372fc0b24cdcb2d7b55a7a5eb770

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: 6c3836e10b23372fc0b24cdcb2d7b55a7a5eb770

crates/duty-tracker/src/contract_manager.rs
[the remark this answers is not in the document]
storopoli: Nice!

crates/primitives/src/wots.rs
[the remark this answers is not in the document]
storopoli: Yeah that is not a bad unsafe. I'm fine with it. It is very easy to verify.

crates/primitives/src/wots.rs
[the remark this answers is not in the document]
storopoli: https://ethereum.org/en/developers/docs/nodes-and-clients/#execution-clients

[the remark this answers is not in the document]
storopoli: Yeah that is not a bad unsafe. I'm fine with it. It is very easy to verify.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
storopoli: I don't like magic numbers :)
I'm fine with leaving them as it is after I nag PR authors about magic number in reviews.
Let me put it in another way: I don't like magic numbers that were added blindly in a knee-jerk reaction.

storopoli (approved):
ACK b79d5a434f98caf55e43ca410af4356f00731a7b

crates/duty-tracker/src/contract_state_machine.rs
ProofOfKeags: What I'm saying here is that the `load_or_create_verifier_scripts` should probably take a "source" or some other thing that we can hot-swap and provide as an argument. Then the mock one just becomes a source that conforms to that API.

ProofOfKeags: fixed it in an amended rebase commit: c9bb17e33c406e02e73a46edfeb068bb41b92e7c

Rajil1213: I agree. The following is just for added context:

Doing it the right way is not straightforward either because we need to provide sp1 with the private key and the network url (via environment variables) .when we pass in the actual source. All of which happen automatically so any source we provide can only be a stub. In a sense, the source in both cases (mock or actual) is controlled via environment variables.

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
ProofOfKeags: I chose blocks that were post-segwit on mainnet, although now that I think about it I should have chosen ones that were post-taproot. I don't know that it matters too much but yeah I think magic numbers in tests are ok.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2746,3 +2726,230 @@ mod tests {
...
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(0))] // This still does 1 test case. It's weird.
+        #[test]
+        fn state_serialization_invertible(state in arb_machine_state().no_shrink()) {
+            let mut time = Instant::now();
+            println!("serializing machine state");
+            match serde_json::to_string(&state) {
+                Ok(serialized) => {
+                    println!("serialization complete. time taken: {:?}", Instant::now().duration_since(time));
```
Rajil1213, line 2930: Can we not use the logging macros here (and in subsequent calls to `println!`)?

Rajil1213 (changes requested):
ACK https://github.com/alpenlabs/strata-bridge/commit/c9bb17e33c406e02e73a46edfeb068bb41b92e7c.

Minor comment on the use of `println`.

Rajil1213 (approved):
ACK https://github.com/alpenlabs/strata-bridge/pull/147/commits/07d271d9c9800a908cebf6e910faab932dda996c.

The logs can be structured a little bit better with key-value pairs instead of format string. But since it's only for the tests, it's fine. Approving this preemptively while the CI continues to run.

PR #149, opened 2025-05-21
TxDriver idempotence

ProofOfKeags:
## Description

This PR allows us to add duplicate transactions to the tx driver without compromising the ability to wait on their burial. To do this I also introduced a new `algebra` crate for abstract utilities you'd normally find in standard FP libraries. I started by adding Semigroup and Monoid since that's what I needed for the new TxJobHeap.

The commits are structured to put the test first where the test verifies the problem exists. If you check out the branch and rewind to the commit where the test is written (prior to the commit where we fix the bug) you should see it failing. You should then see it succeeding at the tip of this branch.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

If anyone has a better idea for the test design or suggestions about how to extract out common utilities into our common testing infrastructure, those would also be good comments to add. I want to lower the barrier to entry on new tests.

## Checklist


crates/algebra/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "algebra"
```
Rajil1213, line 2: I think this crate has a short-enough name that it should be prefixed with `strata-bridge-`.

crates/algebra/src/semigroup.rs
```diff
@@ -0,0 +1,26 @@
...
+/// folding function.
+pub fn sconcat<T: Semigroup>(xs: impl IntoIterator<Item = T>) -> Option<T> {
+    let mut res: Option<T> = None;
+    for x in xs {
+        res = Some(match res {
+            Some(acc) => acc.merge(x),
+            None => x,
+        })
+    }
+    res
```
Rajil1213, line 25: This can be implemented with a `.fold`, can it not?

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -91,21 +166,12 @@ impl TxDriver {
...
                             }
-                            btc_notify::client::TxStatus::Buried { height, .. } => {
+                            btc_notify::client::TxStatus::Buried { .. } => {
                                 // Since our responsibility ends at block inclusion we will send an
                                 // answer on the response channel now. It is the API caller's
                                 // responsibility for handling reorgs after inclusion.
                                 let txid = event.rawtx.compute_txid();
-                                match active_jobs.remove(&txid) {
-                                    Some(job) => {
-                                        info!(%txid, %height, "transaction confirmed in block");
```
Rajil1213, line 101: I recommend preservingt this log (before the call to `discharge`).

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -144,3 +210,147 @@ impl Drop for TxDriver {
...
+    async fn setup() -> Result<(TxDriver, corepc_node::Node), Box<dyn std::error::Error>> {
+        let mut bitcoin_conf = corepc_node::Conf::default();
+        bitcoin_conf.enable_zmq = true;
+        // TODO(proofofkeags): do dynamic port allocation so these can be run in parallel
+        bitcoin_conf.args.extend(vec![
+            "-zmqpubhashblock=tcp://127.0.0.1:23882",
+            "-zmqpubhashtx=tcp://127.0.0.1:23883",
+            "-zmqpubrawblock=tcp://127.0.0.1:23884",
+            "-zmqpubrawtx=tcp://127.0.0.1:23885",
+            "-zmqpubsequence=tcp://127.0.0.1:23886",
```
Rajil1213, line 240: For the time being, can we add the ports into consts so that this config and the one used later below stay consistent?

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -144,3 +210,147 @@ impl Drop for TxDriver {
...
+        snd_res.expect("second drive succeeds");
+
+        Ok(())
+    }
+
+    // This is disabled because it is merely a testing helper function to ensure tests complete in
+    // a timely manner, so we don't want lack of full coverage in this function to distract from
+    // overall coverage.
+    #[cfg_attr(coverage_nightly, coverage(off))]
+    async fn wait_for_height(
```
Rajil1213, line 343: A more useful function would be one that waits for `n` blocks to be mined instead of mining till a fixed block height. I also recommend moving this to the `test-utils` crate because it'll be useful beyond this crate too.

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -26,11 +30,76 @@ pub enum DriveErr {
...
+
+    /// The channel that we should publish on when the job is done.
     respond_on: oneshot::Sender<Result<(), DriveErr>>,
 }
 
+type TxSubscriberSet = Vec<oneshot::Sender<Result<(), DriveErr>>>;
+
+/// The TxJobHeap is a map from [`Txid`]s to the corresponding [`Transaction`] and a list of
+/// listeners for the results.
+struct TxJobHeap(BTreeMap<Txid, (Transaction, TxSubscriberSet)>);
```
Rajil1213, line 46: Is the `Transaction` actually used anywhere?

crates/algebra/src/monoid.rs
```diff
@@ -0,0 +1,26 @@
...
+    fn empty() -> Self;
+}
+
+/// The free catamorphism (fold) over a Monoidal iterator. If the iterator has no elements, it
+/// returns [`Monoid::empty`], otherwise it folds the iterator using [`Semigroup::merge`].
+pub fn concat<T: Monoid>(xs: impl IntoIterator<Item = T>) -> T {
+    xs.into_iter().fold(T::empty(), <T as Semigroup>::merge)
+}
+
+/// The universal catamorphism over all iterators. If given the iterant type has a morphism into a
```
Rajil1213, line 22:
```suggestion
/// The universal catamorphism over all iterators. If the given iterant type has a morphism into a
```
?

Rajil1213 (changes requested):
This PR is a reminder to resume my FP studies. 😅 

Anyways, I've mostly left some nits and questions. Functionally, I don't see any problems with it.

Rajil1213:
Also wanted to say that it's good to finally see a PR that does not decrease the project coverage. 😌

crates/algebra/src/monoid.rs
ProofOfKeags: This function can be used to express all catamorphisms over any iterator, provided that you can project it to a Monoid.

crates/algebra/src/semigroup.rs
ProofOfKeags: Done.

crates/algebra/Cargo.toml
ProofOfKeags: No. It is not a strata-bridge thing. It is universal. This crate could be exported and used on the Core team without issue.

crates/duty-tracker/src/tx_driver.rs
ProofOfKeags: Not right now. But we would hold onto it in case we need to CPFP anchors or resubmit via package submission etc.

crates/duty-tracker/src/tx_driver.rs
ProofOfKeags: I can also add that, but waiting for a specific block height has fewer footguns so I do want to keep it.

crates/duty-tracker/src/tx_driver.rs
ProofOfKeags: Yep. Good idea.

crates/algebra/src/monoid.rs
Rajil1213: This comment was more about grammar. Is `If given the iterant type has a morphism...` correct?

crates/algebra/src/semigroup.rs
Rajil1213: I'll believe it when I see it lol.

crates/duty-tracker/src/tx_driver.rs
Rajil1213: That is non-obvious to me looking at this impl. Could you add a note there somewhere?

crates/duty-tracker/src/tx_driver.rs
Rajil1213: Fair enough. In this particular use case though, you're not actually waiting for block 101 but rather for 101 blocks to be mined on top.

crates/algebra/src/monoid.rs
ProofOfKeags: Yes. The "iterant type" is the `T` in `Item = T`.

Rajil1213: Okay I think what you meant. I'd still prefer something like: 
```suggestion
/// The universal catamorphism over all iterators. Provided that the iterant type has a morphism into a
```

ProofOfKeags: Oh I missed what you were suggesting before. You were correct in your original suggestion but I have changed it to your newest suggestion.

crates/algebra/src/semigroup.rs
ProofOfKeags: Sorry I had commits in the wings and I entered my comment inside my editor.

crates/duty-tracker/src/tx_driver.rs
ProofOfKeags: Well actually we can get rid of it until we need it. Theres no reason to keep it around right now.

crates/algebra/src/semigroup.rs
Rajil1213: Yeah just saw the change. You could make it even more succince to: `Some(opt.map(acc.merge).unwrap_or(x))` but this is fine too.

crates/duty-tracker/src/tx_driver.rs
Rajil1213: That's fine too.

crates/duty-tracker/src/tx_driver.rs
ProofOfKeags: Why not? I'm trying to wait for the first funds to mature which should happen at block 101, no?

crates/algebra/src/semigroup.rs
ProofOfKeags: No you can't because rust can't reason about linearity properly with this. Merge takes an owned argument.

crates/algebra/Cargo.toml
storopoli: Eventually we'll move it outside of `strata-bridge` into its own repo.

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -144,3 +208,137 @@ impl Drop for TxDriver {
...
+        let stop = Arc::new(std::sync::atomic::AtomicBool::new(true));
+        let stop_thread = stop.clone();
+        let mine_task = tokio::task::spawn_blocking(move || {
+            while stop_thread.load(std::sync::atomic::Ordering::SeqCst) {
+                bitcoind
+                    .client
+                    .generate_to_address(1, &new_address)
+                    .unwrap();
+                std::thread::sleep(std::time::Duration::from_millis(100));
+            }
```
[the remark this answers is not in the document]
Rajil1213: I see that you've already approved this PR as well. But I agree with this comment. We should have the initial value of `stop` be `false` and then, pass `true` when we want the mining task to stop.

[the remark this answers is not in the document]
Rajil1213: As @ProofOfKeags is on a flight, I've made that change here: https://github.com/alpenlabs/strata-bridge/commit/d4306eeec1a37b45719a2fed2659600d7d37baf3

But it will require re-approval.

PR #154, opened 2025-05-28
STR-1429: Restrict magic bytes in bridge to 4 bytes

storopoli:
## Description

OL and the specs dictate the `magic_bytes` header of the `OP_RETURN`s are EXACTLY 4 bytes in length.
This PR fixes it by adding checks to the construction of the `metadata` and also changes everything from `strata` to `alp0` (ALP padded by a `0`).

Also fixes a race condition that is making the tests hangs in the `tx-driver`, see "Note to Reviewers".

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

1. Do we need to change these as well? https://github.com/alpenlabs/strata-bridge/blob/d25e334b13f53b59dc56cd8c8200b1e5a467478a/bin/alpen-bridge/src/params.rs#L157-L158

2. The `tx-driver`, when multiple callers submitted the same transaction concurrently, the TxDriver would:

   1. Add both jobs to active_jobs heap
   2. Check for duplicates after adding jobs
   3. Skip broadcast for duplicates but leave jobs in heap
   4. Only notify the first job when transaction confirmed
   5. Leave subsequent duplicate jobs hanging forever

   **Root Cause**: Duplicate detection happened too late in the processing pipeline - jobs were committed to the tracking heap before validation.
   
   **Fix**: Reordered logic to validate duplicates before committing jobs in 9deed4058a4b09dfe8a37bd3807589105fad1cfd 

## Checklist

## Related Issues

STR-1429


crates/primitives/src/scripts/metadata.rs
```diff
@@ -55,6 +57,38 @@ pub enum DepositMetadata {
...
+        ee_address: &[u8],
+        expected_ee_address_size: usize,
+        metadata: DepositMetadata,
+    ) -> BridgeTxBuilderResult<Self> {
+        // Validate tag size
+        if tag.len() != 4 {
+            return Err(DepositTransactionError::InvalidTagSize(tag.len()).into());
+        }
+
+        // Validate EE address size
```
barakshani, line 78: how about check validity of address sum-check?

Rajil1213 (changes requested):
Are you sure the tag is `alp0` and not `alpe`?

The only problem I have with this PR is that the validation is happening when constructing the deposit but I think we can keep that construction completely agnostic to these arbitrary constraints. Ultimately, the source of the tag is the `params` file passed to the bridge node. It is when parsing this toml file that this tag check should take place so that we can get as early a feedback as possible and crash the node. The way it's written now, we will only be able to identify the issue when the deposit transaction is being constructed.

crates/primitives/src/scripts/metadata.rs
```diff
@@ -55,6 +57,38 @@ pub enum DepositMetadata {
...
+    ///
+    /// Returns the metadata as bytes if validation passes.
+    pub fn validate_and_extract(
+        tag: &'tag [u8],
+        ee_address: &[u8],
+        expected_ee_address_size: usize,
+        metadata: DepositMetadata,
+    ) -> BridgeTxBuilderResult<Self> {
+        // Validate tag size
+        if tag.len() != 4 {
```
barakshani, line 74: I believe that the core side pads with 0 if the bytes, taken from params I suppose, are less then 4. The bridge should comply, or for the least the behavior should be consistent. It's fine if here we enforce it to be exactly 4 -- while I haven't checked the logic for where this function is being called, I assume that so far we haven't formatted the bytes in any specific way.

barakshani: [This](https://github.com/alpenlabs/alpen/pull/821/files#diff-e33e40a055c41e397078372c6f49f7246a07657733afeace44c761a94cb93cefR59-R65), no?

barakshani: oh, I see now that this isn't the right place to make my comment -- this is what we are taking from the tx. It's where we take value from the rollup params that this should be enforced.

crates/params/src/default.rs
```diff
@@ -2,34 +2,37 @@
 
 use bitcoin::{relative, Amount};
 
-/// The default tag for the bridge.
-pub(crate) const BRIDGE_TAG: &str = "strata";
+/// Default tag for the bridge.
+pub(crate) const BRIDGE_TAG: &str = "alp\0";
```
Rajil1213, line 6: I don't think the `0` needs escaping.

Rajil1213 (changes requested):
I see that the `PegOutGraphParams` is being validated now but I think we can go a step further. What we really want is for the tag to always be valid by construction. At no point during the lifecycle of the bridge do we want to crash because we were accidentally using the wrong tag. Since it's just bytes, this cannot be enforced. So, I think what we need is a primitive type for the `tag` that wraps around a `Vec<u8>` or `[u8; TAG_SIZE]` that has a bunch of helper impls like `TryFrom<Vec<u8>>`, `TryFrom<String>`, `TryFrom<AsRef<u8>>`, `Into<Vec<u8>>`, `Into<String>`, etc. The params file will serialize into this newtype and it is passed everywhere where we previously accepted a slice of bytes. Eventually, this is something that can go into a common repo.

crates/primitives/src/scripts/metadata.rs
storopoli: Yes it is done in 407a51f81fdf12de2cb9fed8ef2410addd87725a and the node panics, but we might not do that with a recent suggestion.

Rajil1213:
> can you help me regenerating test-data/blocks.bin with the alp0 tag instead of strata. Otherwise the tests will never pass. That is the only thing left for CI to pass in this PR.

This will take time but let me see if I can generate it by running the actual bridge.

crates/params/src/default.rs
storopoli: Done in 9094d239fa10de08ed1da4a93f74bed0fa988909

storopoli:
> I see that the `PegOutGraphParams` is being validated now but I think we can go a step further. What we really want is for the tag to always be valid by construction. At no point during the lifecycle of the bridge do we want to crash because we were accidentally using the wrong tag. Since it's just bytes, this cannot be enforced. So, I think what we need is a primitive type for the `tag` that wraps around a `Vec<u8>` or `[u8; TAG_SIZE]` that has a bunch of helper impls like `TryFrom<Vec<u8>>`, `TryFrom<String>`, `TryFrom<AsRef<u8>>`, `Into<Vec<u8>>`, `Into<String>`, etc. The params file will serialize into this newtype and it is passed everywhere where we previously accepted a slice of bytes. Eventually, this is something that can go into a common repo.

I think I've achieved this with f837f64d42098d6d3cef0adf192743bc73d86208

Rajil1213 (changes requested):
I think this is almost good to go now (except the decision on what the magic bytes should be).

crates/bridge-proof/protocol/src/tx_info.rs
```diff
@@ -70,7 +71,6 @@ pub(crate) struct WithdrawalInfo {
 
 // TODO: make this standard
 pub(crate) fn extract_withdrawal_info(
-    tag_len: usize,
```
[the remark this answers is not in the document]
storopoli: Done in 0da1ee0ad2fb5f0db57348409cfef9a7904ab62e
I also ignore the test that uses hardcoded binary test data with the old `strata` tag.

crates/params/src/tx_graph.rs
```diff
@@ -1,10 +1,26 @@
...
+    serializer.serialize_str(&tag_string)
+}
+
+fn deserialize_tag<'de, D>(deserializer: D) -> Result<Tag, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let tag_string = String::deserialize(deserializer)?;
+    Tag::try_from(tag_string.as_str()).map_err(|e| D::Error::custom(e.to_string()))
+}
```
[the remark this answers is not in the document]
storopoli: Because it serializes automatically as bytes and we need to be a string.

Rajil1213 (changes requested):
Forgot to actually submit the review. 🤦

crates/bridge-proof/protocol/src/tx_info.rs
```diff
@@ -182,10 +204,13 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "rollup_name is not 4 bytes because of hardcoded test data blobs"]
```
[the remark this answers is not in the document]
storopoli: There are two ignored tests not only this one (I’m in my mobile it’s hard to find it)

crates/bridge-proof/protocol/src/program.rs
```diff
@@ -121,6 +121,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "binary test blobs are broken due to using a tag that is not 4 bytes (strata)"]
```
Rajil1213, line 124: Found the other ignored test. Leaving a comment here so we don't miss this either.

Rajil1213 (changes requested):

bin/dev-cli/src/constants.rs
```diff
@@ -18,7 +19,8 @@ pub(crate) const BTC_TO_WEI: u128 = ETH_TO_WEI;
 
 pub(crate) const SATS_TO_WEI: u128 = BTC_TO_WEI / 100_000_000;
 
-pub(crate) const MAGIC_BYTES: &[u8] = b"strata";
+/// Magic bytes for the deposit request transaction.
+pub(crate) const MAGIC_BYTES: Tag = Tag::new([b'a', b'l', b'p', b'0']);
```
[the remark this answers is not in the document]
storopoli: nice catch, done in 519920946ddddeea0f8ff45e664bd8c6d963eed5

crates/params/src/errors.rs
```diff
@@ -0,0 +1,15 @@
+//! Errors for the bridge parameters.
+
+use thiserror::Error;
+
+/// Error while creating or validating a bridge tag.
+#[derive(Debug, Clone, Error)]
+pub enum TagError {
+    /// Tag size is invalid - must be exactly 4 bytes.
+    #[error("tag size must be exactly 4 bytes, got {0} bytes")]
```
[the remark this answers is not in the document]
storopoli: Nice suggestion! Done in 86760630ac9c527a16820ead8a6372014d6a3c92

crates/bridge-proof/protocol/src/statement.rs
```diff
@@ -71,6 +71,13 @@ pub(crate) fn process_bridge_proof(
     rollup_params: RollupParams,
     peg_out_graph_params: PegOutGraphParams,
 ) -> Result<BridgeProofPublicOutput, BridgeProofError> {
+    let tag = Tag::try_from(rollup_params.checkpoint_tag.clone()).map_err(|_| {
```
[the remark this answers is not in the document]
storopoli: Ooops done in 6fff98947396513d744e39ca3a0d0f636e2fd520

storopoli:
This should be good to go EXCEPT that we need the binary test data blobs with `alpn` tag.

Rajil1213:
> This should be good to go EXCEPT that we need the binary test data blobs with alpn tag.

@storopoli done!

Rajil1213 (approved):

PR #155, opened 2025-05-29
(perf) avoid copying wots keys

Rajil1213:
## Description

This PR adds optimizations to reduce memory usage for the bridge node. These include:

* `PegOutGraphInput` is now passed by ref in calls to `PegOutGraph::generate`.
* Delayed nags are now skipped so that we don't publish duplicate nags if multiple ticks pass before being polled.
* Removes the redundant `PegOutGraphInput` from the `Requested` state. The summaries are still indexed by the claim txid.
* Clears the peg out graph cache after the deposit transaction is confirmed.
* Wraps the primitive wots types in Arc's (but this does not actually reduce the memory usage).

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Changing the wots primitives to `Arc<T>` did not reduce the memory usage but this might be due to how it's been implemented. This may require deeper scrutiny.

## Checklist


crates/duty-tracker/src/contract_manager.rs
```diff
@@ -272,6 +272,8 @@ impl ContractManager {
             }
 
             let mut interval = time::interval(nag_interval);
+            // skip any missed ticks to avoid flooding the network with duplicate nag messages
+            interval.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
```
storopoli, line 276: TIL

storopoli (dismissed):
ACK 9b88d2c24e80ce7b5db5edd72e48753c7ab5664d

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2419,6 +2420,11 @@ impl ContractSM {
         &self.state
     }
 
+    /// Returns a mutable copy of the peg out graph.
```
ProofOfKeags, line 2473: immutable?

crates/tx-graph/src/transactions/payout_optimistic.rs
```diff
@@ -60,9 +60,6 @@ pub struct PayoutOptimisticTx {
     connector_n_of_n: ConnectorNOfN,
 
     connector_p: ConnectorP,
-
-    #[expect(dead_code)] // this field is read when settling this tx via cpfp
-    connector_cpfp: ConnectorCpfp,
```
ProofOfKeags, line 65: I don't think we want to make this change. Don't we want operators to be able to do Anchor-CPFP's?

ProofOfKeags (changes requested):

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: Nice catch.

crates/tx-graph/src/transactions/payout_optimistic.rs
Rajil1213: We do. But there's no value in having the connector as part of the Tx struct. The idea was that we'd expose the connector for cpfp-ing it but I don't know if that is what the cpfp api should be like. I can, of course, leave it in if you insist (it's a NOOP except for the few bytes we save by not having it here).

storopoli (approved):
ACK 0eb3050abdc6c61ca41abc446cca5bc32bc893be

storopoli (approved):
ACK 589ea022fa66e543a1d1ff11d3441b7063ef4d39
My bad!

PR #173, opened 2025-06-16
feat(alpen-bridge): restart low-hanging fruit fixes

storopoli:
## Description

This PR improves the Alpen Bridge node's restart reliability by implementing proper shutdown handling, port conflict resolution, and resource cleanup. Previously, the bridge node had several issues that made it difficult to restart cleanly in production environments.

**Key Problems Solved:**

- Bridge node couldn't handle shutdown signals gracefully (SIGINT/SIGTERM)
- RPC server had a broken shutdown mechanism that never actually stopped
- "Address already in use" errors occurred when attempting to restart
- Panic-based error handling made debugging difficult
- No validation of port availability before starting services

**Key Improvements:**

- **Signal Handling**: Added proper SIGINT and SIGTERM signal handling with graceful shutdown
- **RPC Server Shutdown**: Fixed broken shutdown channel mechanism  
- **Port Availability**: Added startup validation and retry logic with exponential backoff
- **Resource Cleanup**: Improved automatic resource cleanup during task cancellation
- **Error Handling**: Replaced panics with proper exit codes and clear error messages

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

**Implementation Notes:**

- Uses `tokio::select!` to handle signals without blocking main bootstrap logic
- Port validation happens before any services start to fail fast on conflicts  
- Retry logic uses exponential backoff (500ms, 1s, 2s, 4s, 8s) for transient port conflicts
- Task cancellation properly propagates through `try_join!` for coordinated shutdown

## Checklist

## Related Issues

Bridge restartable working force

### DIFF
--- a/bin/alpen-bridge/src/main.rs
+++ b/bin/alpen-bridge/src/main.rs
@@ -1,6 +1,8 @@
 //! The Alpen Bridge is a bridge node for the Alpen BitVM rollup.
 
-use std::{fs, path::Path, thread::sleep};
+#[cfg(not(unix))]
+use std::future::pending;
+use std::{fmt, fs, path::Path, process, thread::sleep};
 
 use args::OperationMode;
 use clap::Parser;
@@ -10,7 +12,10 @@ use mode::{operator, verifier};
 use params::Params;
 use serde::de::DeserializeOwned;
 use strata_bridge_common::{logging, logging::LoggerConfig};
-use tracing::{debug, info};
+#[cfg(unix)]
+use tokio::signal::unix::{signal, SignalKind};
+use tokio::{select, signal::ctrl_c};
+use tracing::{debug, error, info};
 
 mod args;
 mod config;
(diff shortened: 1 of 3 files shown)


Rajil1213 (changes requested):
The changes here look fine but at the same time, I don't see how this adds robustness to the system. The implementation in this PR does not have a shutdown mechanism that is any more graceful than what is already there. It _would_ be graceful if when a shutdown signal is received, we make sure to persist all the contract/stake chain data to disk  or perform any relevant cleanup operations (executing/clearing out duties, etc.) before actually shutting down the node. As it stands right now, afaict all we're doing is adding logs for when the shutdown signal is received. You might want to take a look at the [`strata-tasks`](https://github.com/alpenlabs/alpen/tree/main/crates/tasks) crate which has a more robust mechanism for shutdowns but I'm not sure how directly applicable it will be for our use case.

PR #189, opened 2025-06-27
Fix/Do not regenerate graphs when nagged

Rajil1213:
## Description

This PR exposes the `pog_cache` in the `ContractSM` to the `ContractActor` and uses this cache to retrieve graphs when a node is nagged for nonces and partials instead of regenerating the graph each time.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/duty-tracker/src/contract_state_machine.rs
+++ b/crates/duty-tracker/src/contract_state_machine.rs
@@ -3143,7 +3143,8 @@ impl ContractSM {
         &self.state
     }
 
-    /// Returns an immutable copy of the peg out graph.
+    /// Returns an immutable copy of the [`PegOutGraph`] cache indexed by the corresponding stake
+    /// [`Txid`].
     pub const fn pog(&self) -> &BTreeMap<Txid, PegOutGraph> {
         &self.pog
     }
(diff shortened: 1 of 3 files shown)


crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -3143,7 +3143,8 @@ impl ContractSM {
         &self.state
     }
 
-    /// Returns an immutable copy of the peg out graph.
+    /// Returns an immutable copy of the [`PegOutGraph`] cache indexed by the corresponding stake
+    /// [`Txid`].
     pub const fn pog(&self) -> &BTreeMap<Txid, PegOutGraph> {
```
storopoli, line 3148: Isn't this the Claim Txid instead of the Stake Txid?

Rajil1213: No, the cache is referenced by the stake txid because there is circular problem here...we want to check whether we need to generate the graph, if we use `claim_txid`, we'll need to generate the graph to know whether we need to generate the graph.

PR #190, opened 2025-06-27
Hack/support delta zero in stake chain

Rajil1213:
## Description

This PR supports setting a 0-block timelock between consecutive stake transactions in the stake chain. To enable this, it adds the following features:

 - Adds a mechanism for the `tx-driver` to provide immediate feedback if the transaction could not be published.
 - Adds a retry mechanism when publishing stake transactions to account for the case where the parent stake may not have been broadcasted.
 
This also fixes a bug in reassignments where the active graph was not updated upon reassignment.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The retry logic for broadcasting stake transactions is a temporary hack but supporting of `0-blocks` timelock is not as that is something the `stake-chain` crate must support.

## Checklist

### DIFF
--- a/crates/connectors/src/connector_s.rs
+++ b/crates/connectors/src/connector_s.rs
@@ -4,9 +4,12 @@ use std::slice;
 
 use bitcoin::{
     hashes::{sha256, Hash},
-    opcodes::all::{OP_CHECKSIGVERIFY, OP_CSV, OP_EQUALVERIFY, OP_SHA256, OP_SIZE},
+    opcodes::{
+        all::{OP_CHECKSIGVERIFY, OP_CSV, OP_EQUALVERIFY, OP_SHA256, OP_SIZE},
+        OP_TRUE,
+    },
     psbt::Input,
-    relative,
+    relative::{self, LockTime},
     taproot::{ControlBlock, LeafVersion},
     Address, Network, ScriptBuf, TapNodeHash,
 };
@@ -106,7 +109,7 @@ impl ConnectorStake {
     /// <stake_preimage> OP_EQUALVERIFY <ΔS> OP_CHECKSEQUENCEVERIFY
     /// ```
     pub fn generate_script(&self) -> ScriptBuf {
-        ScriptBuf::builder()
+        let mut locking_script = ScriptBuf::builder()
             .push_slice(self.operator_pubkey.serialize())
             .push_opcode(OP_CHECKSIGVERIFY)
             .push_opcode(OP_SIZE)
@@ -114,10 +117,19 @@ impl ConnectorStake {
             .push_opcode(OP_EQUALVERIFY)
             .push_opcode(OP_SHA256)
             .push_slice(self.stake_hash.to_byte_array())
-            .push_opcode(OP_EQUALVERIFY)
-            .push_sequence(self.delta.into())
-            .push_opcode(OP_CSV)
-            .into_script()
+            .push_opcode(OP_EQUALVERIFY);
+
+        // handle `0`-locktime differently as pushing `0` sequence means no element is pushed which
+        // results in the stack being empty when it is executed when spending.
+        if self.delta != LockTime::ZERO {
+            locking_script = locking_script
+                .push_sequence(self.delta.into())
+                .push_opcode(OP_CSV)
+        } else {
+            locking_script = locking_script.push_opcode(OP_TRUE);
+        }
+
+        locking_script.into_script()
     }
 
     /// Creates a P2TR address with key spend path for the given operator set and a single script
--- a/crates/duty-tracker/src/contract_state_machine.rs
+++ b/crates/duty-tracker/src/contract_state_machine.rs
@@ -2373,6 +2373,7 @@ impl ContractSM {
             )));
         }
 
+        let operator_table = self.cfg().operator_table.clone();
         match assignment.deposit_state() {
             DepositState::Dispatched(dispatched_state) => {
                 let assignee = dispatched_state.assignee();
@@ -2446,6 +2447,9 @@ impl ContractSM {
                     ContractState::Assigned {
                         fulfiller,
                         deadline,
+                        claim_txids,
+                        active_graph,
+                        peg_out_graphs,
                         ..
                     } => {
                         if *fulfiller != assignee {
@@ -2453,6 +2457,23 @@ impl ContractSM {
 
                             *fulfiller = assignee;
                             *deadline = dispatched_state.exec_deadline();
+                            let p2p_key =
+                                operator_table
+                                    .idx_to_op_key(&assignee)
+                                    .ok_or(TransitionErr(format!(
+                                    "could not convert operator index {assignee} to operator key"
+                                )))?;
+
+                            let claim_txid = claim_txids.get(p2p_key).ok_or(TransitionErr(format!(
+                                "could not find claim_txid for operator {p2p_key} in csm {deposit_txid}",
+                            )))?;
+
+                            *active_graph = peg_out_graphs
+                                .get(claim_txid)
+                                .ok_or(TransitionErr(format!(
+                                    "could not find peg out graph for claim txid {claim_txid} in csm {deposit_txid}",
+                                )))?
+                                .to_owned();
                         }
 
                         Ok(Some(OperatorDuty::FulfillerDuty(
@@ -2466,6 +2487,9 @@ impl ContractSM {
                     ContractState::StakeTxReady {
                         fulfiller,
                         deadline,
+                        claim_txids,
+                        active_graph,
+                        peg_out_graphs,
                         ..
                     } => {
                         if *fulfiller != assignee {
@@ -2473,6 +2497,24 @@ impl ContractSM {
 
                             *fulfiller = assignee;
                             *deadline = dispatched_state.exec_deadline();
+
+                            let p2p_key =
+                                operator_table
+                                    .idx_to_op_key(&assignee)
+                                    .ok_or(TransitionErr(format!(
+                                    "could not convert operator index {assignee} to operator key"
+                                )))?;
+
+                            let claim_txid = claim_txids.get(p2p_key).ok_or(TransitionErr(format!(
+                                "could not find claim_txid for operator {p2p_key} in csm {deposit_txid}",
+                            )))?;
+
+                            *active_graph = peg_out_graphs
+                                .get(claim_txid)
+                                .ok_or(TransitionErr(format!(
+                                    "could not find peg out graph for claim txid {claim_txid} in csm {deposit_txid}",
+                                )))?
+                                .to_owned();
                         }
 
                         let is_assigned_to_me = *fulfiller == pov_idx;
@@ -2613,6 +2655,7 @@ impl ContractSM {
                 l1_start_height,
                 ..
             } => {
+                let txid = tx.compute_txid();
                 if !is_fulfillment_tx(
                     network,
                     &peg_out_graph_params,
@@ -2625,6 +2668,7 @@ impl ContractSM {
                     // might get somebody else's stake transaction here.
                     // this can happen if this node's stake transaction is settled before other
                     // nodes'.
+                    warn!(%txid, "received a non-fulfillment tx in process_fulfillment_confirmation");
 
                     // FIXME: (@Rajil1213) this should be an error case.
                     return Ok(None);
@@ -2635,7 +2679,7 @@ impl ContractSM {
                     let stake_txid = active_graph.1.stake_txid;
 
                     Some(OperatorDuty::FulfillerDuty(FulfillerDuty::PublishClaim {
-                        withdrawal_fulfillment_txid: tx.compute_txid(),
+                        withdrawal_fulfillment_txid: txid,
                         stake_txid,
                         deposit_txid,
                     }))
--- a/crates/duty-tracker/src/executors/optimistic_withdrawal.rs
+++ b/crates/duty-tracker/src/executors/optimistic_withdrawal.rs
@@ -42,7 +42,7 @@ use crate::{
         wots_handler::get_withdrawal_fulfillment_wots_pk,
     },
     s2_session_manager::MusigSessionManager,
-    tx_driver::TxDriver,
+    tx_driver::{DriveErr, TxDriver},
 };
 
 pub(crate) async fn handle_publish_first_stake(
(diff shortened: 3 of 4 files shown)


crates/connectors/src/connector_s.rs
```diff
@@ -106,18 +109,27 @@ impl ConnectorStake {
...
             .push_slice(self.stake_hash.to_byte_array())
-            .push_opcode(OP_EQUALVERIFY)
-            .push_sequence(self.delta.into())
-            .push_opcode(OP_CSV)
-            .into_script()
+            .push_opcode(OP_EQUALVERIFY);
+
+        // handle `0`-locktime differently as pushing `0` sequence means no element is pushed which
+        // results in the stack being empty when it is executed when spending.
+        if self.delta != LockTime::ZERO {
```
storopoli, line 124: You should not import the `LockTime` from the `relative` module in rust-bitcoin.
This may bite you in the future because we have two different `LockTime`s:

- `absolute::LockTime`
- `relative::LockTime`

For best practices, I always adhere to using the module-level qualification.
In other words instead of `LockTime` use `relative::LockTime`.

storopoli: And then you can undo the import changes in the top of the file.

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
```diff
@@ -158,12 +155,71 @@ pub(crate) async fn handle_advance_stake_chain(
...
+                    %num_retries,
+                    "failed to broadcast first stake transaction, retrying..."
+                );
+
+                num_retries += 1;
+            }
+        }
+
+        debug!(%stake_txid, %stake_index, %num_retries, "waiting for {} seconds before retrying", RETRY_DELAY.as_secs());
+        tokio::time::sleep(RETRY_DELAY).await;
```
storopoli, line 231: Full import here?
This module does not have a `use tokio::Foo` in the top of the file.
Hence this `sleep` usage is "flying under the radar", which I don't like because can bite us in the future.

storopoli (changes requested):

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
Rajil1213: This is a deliberate choice. There are two `time` modules `std::time` and `tokio::time`. We should avoid mixing sync and async modules. This helps disambiguate.

storopoli: Fair enough

storopoli (approved):
ACK 32c86043afd823e8d7b1384d59916f5a56cc10c4

But I'd suggest we wait on @ProofOfKeags review on this before we merge.

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
```diff
@@ -158,12 +155,71 @@ pub(crate) async fn handle_advance_stake_chain(
...
+                    ?e,
+                    %stake_txid,
+                    %stake_index,
+                    %num_retries,
+                    "failed to broadcast first stake transaction, retrying..."
+                );
+
+                num_retries += 1;
+            }
+        }
```
ProofOfKeags: I'm going to send you on a side quest. This isn't a condition for approving this PR but if you don't do the sidequest, I will. I think you may enjoy the exercise.

Write me a retry combinator in the `algebra` crate:

```rust
type Strategy<E> = ... // Define the schedule and error continuations.
async fn retry<A, E, Fut: Future<Output = Result<A, E>, Gen: FnMut() -> Fut>(strategy: Strategy<E>) -> impl Fn(Gen) -> Result<A, E> {...}
```

Then define this PR in terms of the retry combinator.

This is neither the first nor the last instance of retrying failed actions. If you want to consult canon for other instances of this being defined, I give you two examples:

[Simple](https://hackage-content.haskell.org/package/extra-1.8/docs/Control-Exception-Extra.html#v:retry) and [Sophisticated](https://hackage.haskell.org/package/retry-0.9.3.1/docs/UnliftIO-Retry.html)

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -120,23 +125,40 @@ impl TxDriver {
...
                             move |tx| tx == &rawtx_filter
                         ).await;
-                        active_tx_subs.push(tx_sub);
-                        active_jobs = active_jobs.merge(job.into());
 
                         if rpc_client.get_raw_transaction_verbosity_zero(&txid).await.is_ok() {
-                            debug!(%txid, "transaction already broadcasted, skipping resubmission.");
+                            debug!(%txid, "transaction already broadcasted, skipping resubmission but subscribing to events");
+                            active_tx_subs.push(tx_sub);
+                            active_jobs = active_jobs.merge(job.into());
```
ProofOfKeags, line 132: if the transaction is already buried for some reason, this will hang indefinitely.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -2373,6 +2373,7 @@ impl ContractSM {
             )));
         }
 
+        let operator_table = self.cfg().operator_table.clone();
```
ProofOfKeags, line 2376: Commentary: I'm increasingly of the opinion that we should Arc the operator table and make it globally available via reference.

crates/connectors/src/connector_s.rs
```diff
@@ -121,7 +121,7 @@ impl ConnectorStake {
 
         // handle `0`-locktime differently as pushing `0` sequence means no element is pushed which
         // results in the stack being empty when it is executed when spending.
-        if self.delta != LockTime::ZERO {
+        if self.delta != relative::LockTime::ZERO {
```
ProofOfKeags, line 124: 😳

crates/duty-tracker/src/contract_state_machine.rs
storopoli: Make the damn thing a static ref: https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html

We don't need ref counting for something that we know will live for the whole duration of the program.

crates/duty-tracker/src/tx_driver.rs
Rajil1213: Yes, but that behavior is already present in `main`. This change just moves the logic so that we don't subscribe to events for transactions we couldn't even submit in the first place. Do you recommend that I make that change (meaning ability to handle transactions buried in the past)?

crates/connectors/src/connector_s.rs
```diff
@@ -106,18 +109,27 @@ impl ConnectorStake {
     /// <stake_preimage> OP_EQUALVERIFY <ΔS> OP_CHECKSEQUENCEVERIFY
     /// ```
     pub fn generate_script(&self) -> ScriptBuf {
-        ScriptBuf::builder()
+        let mut locking_script = ScriptBuf::builder()
```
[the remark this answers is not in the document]
Rajil1213: I'm ambivalent to this but sure.

PS: FWIW, I saw this nit coming the moment I wrote this.

crates/duty-tracker/src/contract_state_machine.rs
Rajil1213: I think we should go a little bit further. The `ContractCfg` should be split into `StaticCfg` (those that apply to all contracts such as the params) and the actual `ContractCfg` that applies to the contract under consideration (the `deposit_idx` and `deposit_tx`).

Rajil1213: I do consider this out of scope for this PR though. So I'll just create a ticket.

crates/duty-tracker/src/tx_driver.rs
Rajil1213: I think it is strictly necessary for this ticket: https://alpenlabs.atlassian.net/browse/STR-1544 (which is about merging the `StakeTxReady` and `Fulfilled` states). So, I'll add this feature when I work on that.

crates/connectors/src/connector_s.rs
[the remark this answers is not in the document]
Rajil1213: No, I'm fine with making this change.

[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/e3944eabcfcffa6c30a669fed4d531d555a6986c.

crates/duty-tracker/src/executors/wots_handler.rs
```diff
@@ -133,14 +133,11 @@ pub(super) async fn sign_assertions(
...
-                "public params signatures must have the right size, expected: {}, got: {}",
-                NUM_PUBS,
-                public_params_sigs.len()
-            )
-            .into(),
-        ));
+        return Err(ContractManagerErr::FatalErr(format!(
+            "public params signatures must have the right size, expected: {}, got: {}",
+            NUM_PUBS,
+            public_params_sigs.len()
```
storopoli, line 139:
```suggestion
        return Err(ContractManagerErr::FatalErr(format!(
            "public params signatures must have the right size, expected: {NUM_PUBS}, got: {}",
            public_params_sigs.len()
```

crates/duty-tracker/src/executors/wots_handler.rs
```diff
@@ -152,14 +149,11 @@ pub(super) async fn sign_assertions(
...
-                NUM_U256,
-                field_elems_sigs.len()
-            )
-            .into(),
-        ));
+        return Err(ContractManagerErr::FatalErr(format!(
+            "field element signatures must have the right size, expected: {}, got: {}",
+            NUM_U256,
+            field_elems_sigs.len()
+        )));
```
storopoli, line 156:
```suggestion
        return Err(ContractManagerErr::FatalErr(format!(
            "field element signatures must have the right size, expected: {NUM_U256}, got: {}",
            field_elems_sigs.len()
        )));
```

crates/duty-tracker/src/executors/wots_handler.rs
```diff
@@ -169,14 +163,11 @@ pub(super) async fn sign_assertions(
...
-                NUM_HASH,
-                hash_sigs.len()
-            )
-            .into(),
-        ));
+        return Err(ContractManagerErr::FatalErr(format!(
+            "hash signatures must have the right size, expected: {}, got: {}",
+            NUM_HASH,
+            hash_sigs.len()
+        )));
```
storopoli, line 170:
```suggestion
        return Err(ContractManagerErr::FatalErr(format!(
            "hash signatures must have the right size, expected: {NUM_HASH}, got: {}",
            hash_sigs.len()
        )));
```

storopoli (changes requested):

crates/algebra/src/retry.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Continue retrying with the specified delay.
+    Retry(Duration),
+
+    /// Stop retrying and return the error.
+    Stop,
+}
+
+/// A retry strategy that determines how to handle errors and when to retry.
+///
+/// This is similar to Haskell's [RetryPolicyM](https://hackage.haskell.org/package/retry-0.9.3.1/docs/UnliftIO-Retry.html), providing a way to configure
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/4900c9383724e5ee3d31117180e27a1369b798cc.

crates/algebra/src/retry.rs
```diff
@@ -0,0 +1,405 @@
...
+/// retry behavior through error classification and scheduling.
+#[derive(Clone)]
+pub struct Strategy<E> {
+    /// Determines the action to take for a given error and retry attempt number.
+    error_handler: ErrorHandler<E>,
+
+    /// Maximum number of retry attempts (None for unlimited).
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/4900c9383724e5ee3d31117180e27a1369b798cc.

crates/algebra/src/retry.rs
```diff
@@ -0,0 +1,405 @@
...
+}
+
+impl<E> Strategy<E> {
+    /// Creates a new retry strategy with the given error handler.
+    ///
+    /// This will retry indefinitely unless a maximum number of retries is set later with
+    /// [`Self::with_max_retries`].
+    pub fn new<F>(error_handler: F) -> Self
+    where
+        F: Fn(&E, usize) -> RetryAction + Send + Sync + 'static,
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/4900c9383724e5ee3d31117180e27a1369b798cc.

crates/algebra/src/retry.rs
```diff
@@ -0,0 +1,405 @@
...
+            }),
+            max_retries: self.max_retries,
+        }
+    }
+}
+
+/// Retry function that applies a strategy to a future-generating function.
+///
+/// This function takes a retry strategy and returns a function that can be applied
+/// to a future generator to add retry behavior.
```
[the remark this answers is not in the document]
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/4900c9383724e5ee3d31117180e27a1369b798cc.

crates/duty-tracker/src/executors/wots_handler.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/5f0ce93c8c5a93fbd25fa6e198bf4f3314f48942.

crates/duty-tracker/src/executors/wots_handler.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/5f0ce93c8c5a93fbd25fa6e198bf4f3314f48942.

storopoli (approved):
ACK 5f0ce93c8c5a93fbd25fa6e198bf4f3314f48942

storopoli (dismissed):
ACK 48b3abe45a19cfd0650fc77b3f5e38695e6e0c79

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
```diff
@@ -187,49 +187,54 @@ async fn try_publish_stake_tx(
...
-        {
-            Ok(_) => {
-                debug!(%stake_txid, %stake_index, "successfully broadcasted stake transaction");
-                return Ok(());
+    // Create a retry strategy that handles different error types appropriately
+    let strategy = Strategy::new(move |error: &ContractManagerErr, attempt| {
+        match error {
+            ContractManagerErr::TxDriverErr(DriveErr::DriverAborted) => {
+                // Fatal error - don't retry
+                error!(?error, %stake_txid, %stake_index, "fatal error: driver aborted, not retrying");
```
ProofOfKeags, line 197: Does it make sense to put the logs about whether or not something is being retried *in* the retry combinator? I think the error handler itself would benefit from "ignoring" the retry semantics if possible, but I can be convinced otherwise.

ProofOfKeags (dismissed):
Overall I think this is great and is mergeable. I do have two lines of questioning I'd like you to consider before merging though. The main one that I think needs to be sorted out earlier than later is whether or not `attempt` (the second argument to your error handler is useful enough to keep. You can control the number of attempts with the strategy itself, so the only reason to put it there is that if you want to make your decision based off of the *combo* of what error you get back and what attempt you're on. This, to me, doesn't seem to be that useful mostly on the grounds that I think that aborting a retry seems like it would *either* entirely depend on the error, *or* entirely depend on the attempt. WDYT?

crates/algebra/src/retry.rs
```diff
@@ -0,0 +1,405 @@
...
+        let first_handler = self.error_handler;
+        Strategy {
+            error_handler: Arc::new(move |error, attempt| match first_handler(error, attempt) {
+                RetryAction::Stop => other_handler(error, attempt),
+                retry_action => retry_action,
+            }),
+            max_retries: self.max_retries,
+        }
+    }
+}
```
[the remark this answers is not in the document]
Rajil1213: > My gut instinct is no,

I'd say the same. More concretely, having an `and` would lead to the following semantics:

> if the first error handler says retry again, run the error through another error handler as well.

I can't think of a case where this would be useful since we can just `match` on both cases in the first handler instead of passing it to another one.

crates/duty-tracker/src/executors/optimistic_withdrawal.rs
Rajil1213: I think the error handling logic (which the logs are a part of) should live outside the combinator. The reason being that the consumer has more context on how to handle the error than the conbinator which only know how to retry and action. This is similar to a master-slave architecture where the consumer controls the logic while the conbinator "dumbly" performs the action. The error handler is also capturing useful contextual information (such as the stake txid, index).

Rajil1213:
> whether or not attempt (the second argument to your error handler is useful enough to keep

There may be a scenario where the error handler decides whether to retry or not depending upon both the error type and the number of attempts that have alresdy been made. For example, there may be some actions we can keep retrying given a certain error while there may be some errors which when encountered, we'd want to stop after a finite number of attempts. That's the only reason this has been kept but I can remove it if this is not a convincing enough argument.

PR #193, opened 2025-06-27
Fix/update contract indexing to handle actor shutdown

Rajil1213:
## Description

This PR uses the height of the current operator's stake chain to compute the next deposit index offset when processing a block. In the current implementation, this value is computed using the number of active contract actors. This is problematic since the deposit index should always increase monotonically but the number of contract actors can decrease when the associated contract is resolved and the actor is terminated.

This also renames `shutdown` to `terminate` to soften the phrasing. CC: @barakshani 

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

+1 in the "we need a proper test-suite" jar.

## Checklist

### DIFF
--- a/crates/duty-tracker/src/contract_actor.rs
+++ b/crates/duty-tracker/src/contract_actor.rs
@@ -69,8 +69,8 @@ pub enum ContractActorMessage {
     /// Clears the [`PegOutGraph`] cache.
     ClearPogCache,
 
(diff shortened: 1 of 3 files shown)


storopoli (approved):
ACK 737b7b6cc3c93349bd9ba3e99cb641c3bf5da1f4

PR #204, opened 2025-07-03
Algebra: new gadgets

ProofOfKeags:
## Description

Here is a set of new gadgets I've worked on as I've been inspired to do so. None of this is strictly necessary but I think absorbing it will be nice as it'll give us nice primitives to work with.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Each commit is reviewable separately and there shouldn't be any dependencies between them. As such if one of these commits needs more work we can cherry pick it out and merge the rest while that one is polished.

## Checklist

### DIFF
--- a/crates/algebra/src/bijection.rs
+++ b/crates/algebra/src/bijection.rs
@@ -1,0 +1,94 @@
+//! Bijections between sets of unique elements.
+use std::{
+    collections::{BTreeMap, HashMap},
+    hash::Hash,
+    sync::Arc,
+};
+
+use crate::category;
+
+type Lookup<A, B> = Arc<dyn Fn(&A) -> Option<B>>;
+
+/// A Bijection is a two way lookup.
+#[allow(missing_debug_implementations)]
+pub struct Bijection<A: Clone + 'static, B: Clone + 'static> {
+    forward: Lookup<A, B>,
+    reverse: Lookup<B, A>,
+}
+impl<A: Clone, B: Clone> Bijection<A, B> {
+    /// Finds the unique corresponding value B for a given A
+    pub fn lookup(&self, a: &A) -> Option<B> {
+        (self.forward)(a)
+    }
+
+    /// Inverts the Bijection
+    pub fn invert(self) -> Bijection<B, A> {
+        Bijection {
+            forward: self.reverse,
+            reverse: self.forward,
+        }
+    }
+
+    /// Composes two Bijections. Note: If the codomain of self doesn't perfectly match the domain of
+    /// other, then the composed Bijection won't behave as expected.
+    pub fn comp<C: Clone>(&self, other: &Bijection<B, C>) -> Bijection<A, C> {
+        let forward_ab = self.forward.clone();
+        let forward_bc = other.forward.clone();
+        let reverse_cb = other.reverse.clone();
+        let reverse_ba = self.reverse.clone();
+
+        let forward = Arc::new(move |a: &A| forward_ab(a).and_then(category::moved(&*forward_bc)));
+        let reverse = Arc::new(move |c: &C| reverse_cb(c).and_then(category::moved(&*reverse_ba)));
+        Bijection { forward, reverse }
+    }
+}
+
+impl<A: Ord + Clone, B: Ord + Clone> TryFrom<BTreeMap<A, B>> for Bijection<A, B> {
+    type Error = B;
+
+    fn try_from(value: BTreeMap<A, B>) -> Result<Self, Self::Error> {
+        let mut reverse_map = BTreeMap::new();
+        for (a, b) in value {
+            if reverse_map.contains_key(&b) {
+                return Err(b);
+            } else {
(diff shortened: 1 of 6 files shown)


ProofOfKeags:
I'm not fundamentally opposed to moving to a new repo. I just don't want to yet. There's a whole new level of procedure required when you do with all the versioning and tagging and stuff like that. I want to be able to make updates to the API and understand in real time how they break the type/borrow checker inference rules. It's not just about writing code that works. It's also about writing code that is easy to call. Lifetime and type inference in rust is still not something I feel super confident in in all situations and so I will be making frequent updates to things as I learn about their finer details at the call sites.

crates/algebra/src/state_machine.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        let inputs = stream::StreamExt::cycle(stream::iter(vec![false, true])).take(10);
+
+        sm.attach(inputs);
+
+        let transcript = sm
+            .filter_map(|x| future::ready(x.ok()))
+            .collect::<Vec<usize>>()
+            .await;
+        println!("{transcript:?}");
```
storopoli, line 171: Use logging and not `println` statements in tests.
Also, can you add an `assert!`/`assert_eq` here?

crates/algebra/src/bijection.rs
```diff
@@ -0,0 +1,94 @@
...
+    hash::Hash,
+    sync::Arc,
+};
+
+use crate::category;
+
+type Lookup<A, B> = Arc<dyn Fn(&A) -> Option<B>>;
+
+/// A Bijection is a two way lookup.
+#[allow(missing_debug_implementations)]
```
storopoli, line 13: add `reason = "foo"`

crates/algebra/src/bijection.rs
```diff
@@ -0,0 +1,94 @@
+//! Bijections between sets of unique elements.
+use std::{
+    collections::{BTreeMap, HashMap},
+    hash::Hash,
+    sync::Arc,
+};
+
+use crate::category;
+
+type Lookup<A, B> = Arc<dyn Fn(&A) -> Option<B>>;
```
storopoli, line 10: `Arc` ing here does not make sense. What if the user does not want to pay the "atomic" overhead? Can we switch this a `Box` or any other more generic thing that can be supersed by an `Arc` or any other primitive that the user might bring?

crates/algebra/src/category.rs
```diff
@@ -56,22 +59,58 @@ pub const fn iden<A>(a: A) -> A {
 
 /// Lifts an `FnOnce` that takes a borrowed argument into one that consumes that argument. This is
 /// useful because there is no way to build a function of type `f : A -> &A`
+#[inline(always)]
 pub fn moved_once<A, B>(f: impl FnOnce(&A) -> B) -> impl FnOnce(A) -> B {
```
storopoli, line 63: why this and `index` are the ones that is only `inline` without the always?
Ref: https://nnethercote.github.io/perf-book/inlining.html

storopoli (changes requested):
some comments

crates/algebra/src/state_machine.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /// Attaches this StateMachine to an Input Stream which repeatedly feeds it.
+    pub fn attach(&mut self, input_stream: impl Stream<Item = Input> + Unpin + 'static) {
+        self.input_stream = Box::new(input_stream);
+    }
+
+    /// Registers an async State continuation to be called every time an Input event is accepted by
+    /// the StateMachine.
+    pub fn state_hook<R, Fut: Sync + Send + Future<Output = ()> + 'static>(
```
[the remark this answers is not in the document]
Rajil1213: Do you think it's worth having a pre-state hook as well?

crates/algebra/src/bijection.rs
ProofOfKeags: The reason for this is to make Bijections easily clonable which they need to be for bijection composition if we want that. What overhead are you talking about, is it more expensive to deref an Arc than a Box?

crates/algebra/src/category.rs
ProofOfKeags: Because I'm less confident that the more complicated functions should always be inlined. The "core" gadgets, particularly the ones that do closure allocation should be safe to inline but that doesn't mean that performance wise the inlining is self-justifying. The trouble in rust is that inlining rules are a little tricky. In pure functional (referentially transparent) code, functions are *always safe to inline*. Inlining is worth it for functions with small bodies where the allocation of the stack frame is more overhead than just puking the RHS of the function definition at the callsite.

The reason inlining is so prevalent in pure functional languages is because very often inlining *exposes other optimizations*, however, rust does not have a user defined compiler rewrite rule system the way Haskell does so inlining is less essential to performance optimization because we can't define rewrite rules for the simplifier.

As such, I only went for inlining always for the core utilities.

PR #209, opened 2025-07-09
feat!: add claim and withdraw data to `ContractState`s past `Assigned`

storopoli:
## Description

The `ContractState::Resolved` does not have a `claim_txid` field and this is needed for the bridge RPC monitoring API.

Also none of the post-`Assigned` states have the `withdraw_request_txid` which is also needed for the bridge RPC monitoring API.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Light breaking since it adds a new field to enum variants.

## Checklist

## Related Issues

STR-1560

### DIFF
--- a/bin/alpen-bridge/src/rpc_server.rs
+++ b/bin/alpen-bridge/src/rpc_server.rs
@@ -568,6 +568,7 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
 
         let mut withdrawals = Vec::new();
         for entry in all_entries {
+            // NOTE: this is a source of bugs, don't use the `_` to match all.
             match &entry.0.state.state {
                 ContractState::Assigned {
                     withdrawal_request_txid,
@@ -580,10 +581,37 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
                 | ContractState::Fulfilled {
                     withdrawal_request_txid,
                     ..
+                }
+                | ContractState::Claimed {
+                    withdrawal_request_txid,
+                    ..
+                }
+                | ContractState::Challenged {
+                    withdrawal_request_txid,
+                    ..
+                }
+                | ContractState::PreAssertConfirmed {
+                    withdrawal_request_txid,
+                    ..
+                }
+                | ContractState::AssertDataConfirmed {
+                    withdrawal_request_txid,
+                    ..
+                }
+                | ContractState::Asserted {
+                    withdrawal_request_txid,
+                    ..
                 } => {
                     withdrawals.push(*withdrawal_request_txid);
                 }
-                _ => {}
+
+                ContractState::Requested { .. }
+                | ContractState::Deposited { .. }
+                // NOTE: Resolved contracts have no *current* withdrawals and will pollute the return array.
+                | ContractState::Resolved { .. }
+                | ContractState::Disproved { .. } => {
+                    continue;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -603,6 +603,10 @@ pub enum ContractState {
     /// This state describes the state after either the optimistic or defended payout transactions
     /// confirm.
     Resolved {
+        /// This is an index so we can look up the claim txid that is owned by the specified key.
+        /// This is primarily used to process assignments.
```
Rajil1213, line 607: This is not the purpose of this field in the `Resolved` state. It's used to associate resolved contracts with the corresponding claim txs.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -603,6 +603,10 @@ pub enum ContractState {
...
+        claim_txids: BTreeMap<P2POperatorPubKey, Txid>,
```
Rajil1213, line 608: I think having a single `Txid` is enough here. Once the contract is resolved, there is no point in keeping around claim txids that will never be identified by the CSM.

In the RPC impls, you can just create a `BTreeMap` with a bogus key since we only care about the values in the `claim_txids` method.

Rajil1213 (commented):
This looks good in the current state of things but it will probably undergo changes once we change the RPC to only serve claim transactions that actually make it on chain. More concretely, in the `Resolved` state, we only need the `claim_txid` that is actually submitted on-chain instead of the entire `BTreeMap` of `claim_txid`'s.

The only other nit is regarding the docstring on the `claim_txids` field.

crates/duty-tracker/src/contract_state_machine.rs
storopoli: Check d49bf40d93909e5086c746656aa0c6953ace8627

crates/duty-tracker/src/contract_state_machine.rs
storopoli: Check d49bf40d93909e5086c746656aa0c6953ace8627

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -580,10 +581,37 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
...
+                    withdrawal_request_txid,
+                    ..
                 } => {
                     withdrawals.push(*withdrawal_request_txid);
                 }
-                _ => {}
+
+                ContractState::Requested { .. }
+                | ContractState::Deposited { .. }
+                | ContractState::Resolved { .. }
```
Rajil1213, line 610: Looks like this does have `withdrawal_request_txid` since the `get_withdrawal_info` filters against it?

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -603,6 +633,15 @@ pub enum ContractState {
     /// This state describes the state after either the optimistic or defended payout transactions
     /// confirm.
     Resolved {
+        /// The spent claim transaction ID that resolved this contract.
```
Rajil1213, line 636:
```suggestion
        /// The spent claim transaction ID that led to this contract being resolved
```

Rajil1213 (changes requested):
Just a few final nits and this should be good to go if the plan is to get this merged first.

crates/duty-tracker/src/contract_state_machine.rs
```diff
@@ -764,7 +770,11 @@ impl ContractState {
             | ContractState::PreAssertConfirmed { claim_txids, .. }
             | ContractState::AssertDataConfirmed { claim_txids, .. }
             | ContractState::Asserted { claim_txids, .. } => claim_txids,
-            ContractState::Disproved { .. } | ContractState::Resolved { .. } => &BTreeMap::new(),
+            ContractState::Resolved { claim_txid, .. } => {
+                return HashSet::from([*claim_txid]);
+            }
```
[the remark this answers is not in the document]
storopoli: > we collapse the number of valid claims to just the one that resolves the contract at a certain point.

isn’t that what `Resolved` means? That there was a successful claim that spent the deposit UTXO into a successful payout (happy or unhappy)?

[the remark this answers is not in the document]
storopoli: Oh you mean that the `claim_txids` method for `ContractState` will always return a single one for the `Resolved` variant and the plural there is confusing?

bin/alpen-bridge/src/rpc_server.rs
Rajil1213: Oh okay. That makes sense. Could you add a comment next to the `Resolved` variant that clarifies this?

crates/duty-tracker/src/contract_state_machine.rs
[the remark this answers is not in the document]
Rajil1213: > I think it's probably fine but it looks like we collapse the number of valid claims to just the one that resolves the contract at a certain point.

Yes, once a contract is resolved meaning that the deposit utxo associated with that contract has been spent, we can "gc" the claim txids that did not make it on chain. This is information that is not important for historical book-keeping purposes.

PR #211, opened 2025-07-10
feat!: RPC rich claims information

storopoli:
## Description

- Adds `TxInfo` from the `bc-notify` crate to the `claims` RPC method.
- Filters for claims transactions that are either in the mempool or in a block before returning them in the `claims` RPC method.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The commits are linear and self-contained.

Should preferably be merged after #209 for easy of rebase.

## Checklist

## Related Issues

STR-1556

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9805,6 +9805,7 @@ name = "strata-bridge-rpc"
 version = "0.1.0"
 dependencies = [
  "bitcoin",
+ "btc-notify",
  "jsonrpsee",
  "serde",
  "strata-bridge-primitives",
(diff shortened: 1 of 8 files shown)


Rajil1213 (changes requested):
This is one way to do it but I don't think it is the _right_ way to do it.

Right now, this implementation makes the bridge node act like a proxy for a bitcoin full node to query for the status of arbitrary transactions. A much simpler way is to just check the contract status. If the contract is in the `Claimed` state, we know that the claim transaction has been buried. In fact, I think the `getClaims` RPC should just return the `claim_txid` of the `active_graph` if the contract is in the `Claimed` state or higher. This should achieve the goal of only showing claim transactions in the dashboard that are actually on chain (and buried). But do let me know if this is not what is required or if I'm missing something here.

storopoli:
Replaced by #213.

PR #215, opened 2025-07-12
STR-1561 execute withdrawal fulfillments in parallel

Rajil1213:
## Description

This PR fixes an issue with withdrawal execution where if two withdrawal fufillment duties were created together, the second one would not be executed until the first withdrawal fulfillment transaction had been buried. This can also cause delays in publishing the deposit setup message. The cause of this issue is that in both cases, we first try to sync the operator wallet which requires that we obtain a `WriteLock` on the operator wallet. This lock is held until the transaction is buried. This PR fixes this by releasing the lock on the wallet as soon as possible and tracking the used funding utxos in memory so that two withdrawal fulfillment execution threads running in parallel do not end up using the same funding utxo.

**Auxiliary Fixes**

- Fixes the bridge binary to actually use the `shutdown_timeout` configuration option instead of using the hardcoded value of 30 seconds.
- Updates the `config.toml` file to include the `shutdown_timeout` configuration option when running docker locally.
- Updates the sequencer xpriv value in the `.justfile` to use the one corresponding to the `SchnorrKey` in the `params.toml` files.
- Updates the `README.md` in `mock-checkpoint` to specify `deposit-entries`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

Closes STR-1561

### DIFF
--- a/bin/alpen-bridge/src/main.rs
+++ b/bin/alpen-bridge/src/main.rs
@@ -5,7 +5,7 @@
 )]
 #![feature(generic_const_exprs)] //strata-p2p
 
-use std::{fs, path::Path, thread::sleep, time::Duration};
+use std::{fs, path::Path, thread::sleep};
 
 use args::OperationMode;
 use clap::Parser;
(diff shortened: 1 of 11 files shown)


bin/alpen-bridge/src/main.rs
```diff
@@ -59,6 +59,7 @@ fn main() {
 
     let params = parse_toml::<Params>(cli.params);
     let config = parse_toml::<Config>(cli.config);
+    let shutdown_timeout = config.shutdown_timeout;
```
storopoli, line 62: My bad, this was my mistake. Thanks for fixing it.

storopoli (approved):
ACK 55f25941c85835100c4ca4d51aefd44c3d103c42

PR #218, opened 2025-07-15
feat: improve restart resilience of the bridge node

Rajil1213:
## Description

This PR improves the restart resilience of the bridge node in the following ways:

1. **Atomic Updates into Persistence**
   Previously, the bridge node persisted current machine state after processing of a block in two phases -- old contracts were persisted first and then, contracts from the new block were persisted one at a time. This PR makes it so that all contracts (old and new) are persisted over a single database transaction ensuring atomicity.

2. **Deposit Index Offset**
    Previously, when processing a new block, the deposit index offset (used as the starting index for new contracts in the block) was computed by computing the current height of the stake chain. This is problematic because the stake chain database table (`operator_stake_data`) is not updated in sync with the contract state. So, in the event of a restart, it may be possible that the stake table has not been updated whereas the contracts table has. This causes an issue with deposit indexing as the deposit index offset will lag and so, violates the UNIQUE constraint in the `contracts` table. This PR fixes this issue by computing the deposit index offset as the maximum value of the deposit index among all active contracts.

3. **Self Nagging**
    If the node restarts after having published a message (say, `GraphNonces`) but before it has been able to persist into its database (via `ouroboros`), it is possible that the node loses its own data. Since other nodes already have this data, they do not nag this node and so there is no way for it to recover its state and maintain consistency with the rest of the network. This PR adds an `ouroboros_req` channel to the `MessageHandler` and updates the nagging mechanism to use the `MessageHandler` instead of the p2p handle directly.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Don't trust, verify.

## Checklist

## Related Issues

Closes STR-1551 (this PR removes the need for a separate table for leased funding utxos).

### DIFF
--- a/crates/duty-tracker/src/contract_manager.rs
+++ b/crates/duty-tracker/src/contract_manager.rs
@@ -291,7 +291,7 @@ impl ContractManager {
                                     if !ouroboros_duties.is_empty() {
                                         info!(num_duties=ouroboros_duties.len(), "queueing duties generated via ouroboros");
                                         for duty in ouroboros_duties.iter() {
-                                            debug!(?duty);
+                                            trace!(?duty, "received ouroboros message duty");
                                         }
 
                                         duties.extend(ouroboros_duties);
@@ -314,7 +314,7 @@ impl ContractManager {
                         },
                     },
 
-                    // And similarly, prioritize self-nags next
+                    // And similarly, prioritize nags next
                     ouroboros_req = ouroboros_req_receiver.recv() => match ouroboros_req {
                         Ok(req) => {
                             if req.operator_pubkey() == cfg.operator_table.pov_p2p_key() {
--- a/crates/p2p-service/src/message_handler.rs
+++ b/crates/p2p-service/src/message_handler.rs
@@ -8,10 +8,9 @@ use strata_p2p_wire::p2p::v1::GetMessageRequest;
 use tokio::sync::broadcast;
 use tracing::{debug, error, info, trace};
 
-/// Message handler for the P2P client.
+/// Message handler for the bridge node for relaying p2p messages.
 ///
-/// This exposes an interface that allows publishing messages but not reading messages from the p2p
-/// network.
+/// This exposes an interface that allows publishing messages to the node itself as [`libbp2p`](https://docs.rs/libp2p/latest/libp2p/) does not support self-publishing.
 // TODO: (@Rajil1213) rename this to `Outbox` and create a newtype for `P2PHandle` that exposes the
 // interface to read messages off of the p2p network (aka the `Inbox`).
 #[derive(Debug, Clone)]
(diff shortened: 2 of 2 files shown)


crates/duty-tracker/src/contract_manager.rs
```diff
@@ -522,8 +559,15 @@ impl ContractManagerCtx {
...
-        // The next contract will have its index at the tip of the current stake chain.
-        let deposit_idx_offset = self.state.stake_chains.height();
+        // The next contract will have its index at a value that is 1 more than the max deposit
+        // index that has already been processed.
+        let deposit_idx_offset = self
+            .state
+            .active_contracts
+            .values()
+            .map(|contract| contract.deposit_idx() + 1) // start at the max index + 1
+            .max()
```
ProofOfKeags, line 569: I think this is a more "structurally correct computation". You don't have to incorporate this change if you don't want to but it makes the provenance of this computation a little clearer

```suggestion
            .map(|contract| contract.deposit_idx())
            .max()
            .map(|max| max + 1) // start at the max index + 1
```

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -670,28 +670,21 @@ impl ContractManagerCtx {
...
-                .await?;
-
             self.state.active_contracts.insert(sm.deposit_txid(), sm);
         }
 
+        debug!(%num_active_contracts, %num_new_contracts, %height, "finished processing block, committing all contract states to disk");
+        self.state_handles
+            .contract_persister
+            .commit_all(self.state.active_contracts.iter())
+            .await?;
```
ProofOfKeags, line 723: 🧼

ProofOfKeags (changes requested):
> Self Nagging
If the node restarts after having published a message (say, GraphNonces) but before it has been able to persist into its database (via ouroboros), it is possible that the node loses its own data. Since other nodes already have this data, they do not nag this node and so there is no way for it to recover its state and maintain consistency with the rest of the network. This PR adds an ouroboros_req channel to the MessageHandler and updates the nagging mechanism to use the MessageHandler instead of the p2p handle directly.

Ok so this and the call we had earlier made me realize something. You mentioned on the call 3 notions of state:

1. in-memory state
2. persistent state
3. network state

It made me realize that there is a latent requirement that isn't being satisfied in the bridge right now and I think it can cause us problems if we don't address it:

**we must persist a *consistent* state resulting from an outbound message before the rest of the network becomes aware of the message**

To meet this requirement I suggest the following change. Rather than bundling both the OB channel and the p2p handle inside of the message outbox, I think we should only forward the message to the p2p network after we have processed its effects from the OB channel. This will ensure that we always fully synchronize our internal state before publishing to the network.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: Agreed. I've ammended the commit: https://github.com/alpenlabs/strata-bridge/commit/95fca683f7423ca12343663a9ebec064364ad45d.

Rajil1213 (commented):
This was a much simpler change than I anticipated.. Thanks @ProofOfKeags.

However, all this does is maintain consistency between the in-memory state and the network state. Since we do not persist state upon every p2p message (we only do it for some critical ones), it can still be the case that the node does not have its own data after a restart. So we still need self-nagging. Do you think we should also persist state after every p2p message is processed? I am ambivalent to negative on this given that the system does still work and increasing number of db writes have been known to cause issues in our infra.

ProofOfKeags (approved):

storopoli (commented):
Coming late to the party.

Forwarding all `GetMessageRequest`s through ouroboros is a simple solution. I like it!

Thanks for the PR!

PR #221, opened 2025-07-17
fix!: bridge does not crash when contract manager loop exits

storopoli:
## Description

This PR fixes a bug in the graceful shutdown mechanism where the bridge node was not crashing when the contract manager event loop exited unexpectedly.

The issue was that the contract manager task always returned `Ok(())` even when the event loop exited due to errors, preventing the `TaskManager` from detecting the failure and crashing the node as expected.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Thanks to Rajil for filling the bug report!

The key changes are:

1. `ContractManager` Event Loop: Modified the contract manager's event loop to return `Result<(), ContractManagerErr>` instead of `()`, and track whether it exited due to shutdown or error. Also removes the `crash` closure.

2. Error Handling: Replaced `break` statements with return `Err(...)` for error conditions in the event loop. Also, previous to the "graceful shutdown" changes, these would `break` the loop and panic because of: https://github.com/alpenlabs/strata-bridge/blob/49984a53abdaef4f4ca2a612c90760e441458d47/crates/duty-tracker/src/contract_manager.rs#L373

3. Shutdown Tracking: Added a `shutdown_requested` flag to distinguish between normal shutdown and unexpected exit

4. Operator mode Integration in `alpen-bridge`: Updated the operator `bootstrap` to properly handle both `shutdown` and `error` cases using `tokio::select!`

## Checklist

## Related Issues

STR-1586

### DIFF
--- a/bin/alpen-bridge/src/mode/operator.rs
+++ b/bin/alpen-bridge/src/mode/operator.rs
@@ -54,7 +54,7 @@ use strata_bridge_stake_chain::prelude::OPERATOR_FUNDS;
 use strata_p2p::swarm::handle::P2PHandle;
 use strata_p2p_types::{P2POperatorPubKey, StakeChainId};
 use strata_tasks::TaskExecutor;
-use tokio::{net::lookup_host, sync::broadcast, task::JoinHandle};
+use tokio::{net::lookup_host, select, sync::broadcast, task::JoinHandle};
 use tracing::{debug, error, info};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


crates/duty-tracker/src/contract_manager.rs
```diff
@@ -97,10 +97,11 @@ impl ContractManager {
...
             let mut shutdown_receiver = Some(shutdown_receiver);
-            let crash = |e: ContractManagerErr| {
-                error!(?e, "crashing");
-                panic!("{e}");
-            };
+            #[allow(
+                unused_assignments,
+                reason = "It is being used in the control flow statements in this function"
+            )]
+            let mut shutdown_requested = false;
```
Rajil1213, line 104: I'm not sure why `clippy` thinks this is not being used but there is a simpler solution here than to silence clippy. Just `return Ok(())` instead of `break`-ing from the event loop.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -462,7 +467,15 @@ impl ContractManager {
                 }
             }
 
-            info!("contract manager event loop exited");
+            if shutdown_requested {
```
Rajil1213, line 470: Once all the `break`'s have been removed, this code should be `unreachable!` but just to be extra cautious, return a `FatalErr` here. And then, we should be able to remove the `wait_for_completion` function entirely because the event loop will always return an error. In fact, we should change the signature for the `thread_handle` from returning a result type to just returning a `ContractManagerErr` because that is what the event loop should return if it ever does return anything.

Rajil1213 (changes requested):
I'm recommending some refactors in this PR to reduce the amount of code that this fix is adding to the shutdown manager which I think is already pretty complex to reason about. (looking at you `oneshot::channel::<oneshot::Sender>(...)`).

crates/duty-tracker/src/contract_manager.rs
storopoli: > In fact, we should change the signature for the `thread_handle` from returning a result type to just returning a `ContractManagerErr` because that is what the event loop should return if it ever does return anything.

That is so clever!
Yeah I like this suggestion!

Amended in 21a2e37a149a984a5552916fa625d4c0f48fa21d

Rajil1213 (dismissed):
A minor nit but this looks much better. Also, I'm assuming that this has been tested out. If not, I strongly recommend testing it out (easiest way is to add an if condition somewhere like crash if the block count > n).

storopoli:
> A minor nit but this looks much better. Also, I'm assuming that this has been tested out. If not, I strongly recommend testing it out (easiest way is to add an if condition somewhere like crash if the block count > n).

Not yet will test soon.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: @barakshani this PR removes that log entirely and instead, it should just crash the entire bridge node if this ever happens.

PR #222, opened 2025-07-18
duty-tracker: actually persist deposit setup duties on time

ProofOfKeags:
turns out this code would *never* persist contract state after deposit setup processing. This means that we would lose this state if a node went down.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/crates/duty-tracker/src/contract_manager.rs
+++ b/crates/duty-tracker/src/contract_manager.rs
@@ -954,28 +954,25 @@ impl ContractManagerCtx {
                             ))
                         })?);
 
-                    let deposit_setup_duties = contract
-                        .process_contract_event(ContractEvent::DepositSetup {
+                    let deposit_setup_duties =
+                        contract.process_contract_event(ContractEvent::DepositSetup {
                             operator_p2p_key: key.clone(),
                             operator_btc_key: operator_pk,
                             stake_hash: hash,
                             stake_txid: stake_tx.compute_txid(),
                             wots_keys,
-                        })?
-                        .into_iter()
-                        .map(|duty| {
-                            // we need a way to feed the claim txids back into the manager's index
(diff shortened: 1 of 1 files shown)


barakshani:
What are the implications of this? I'm pretty sure we have done withdrawals after a restart on our envs (and I'm very hopeful that it was tested locally before making the restart PRs).

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -954,28 +954,25 @@ impl ContractManagerCtx {
...
+                        // we need a way to feed the claim txids back into the manager's index so we
+                        // skim it off of the publish graph nonces duty.
+                        if let OperatorDuty::PublishGraphNonces { claim_txid, .. } = duty {
+                            self.state.claim_txids.insert(*claim_txid, deposit_txid);
+                        }
+                    }
 
                     // immediately commit the state to disk as this is a critical state transition.
-                    if !duties.is_empty() {
+                    if !deposit_setup_duties.is_empty() {
```
storopoli, line 975: This whole PR can be a 1-liner change:

```diff
- if !duties.is_empty() {
+ if !duties.is_empty() {
```

Why we need all the extras changes? What I am missing?

Rajil1213:
> I'm pretty sure we have done withdrawals after a restart on our envs (and I'm very hopeful that it was tested locally before making the restart PRs).

@barakshani with self-nagging, even without persisting this state, it isn't exactly an issue. So, it should not cause any observable effects on the system.

Rajil1213 (approved):
Nice catch.

ProofOfKeags:
Yeah @barakshani this is just to cover an edge case where we can lose state

PR #223, opened 2025-07-21
fix(operator_wallet) add retry mechanism to op wallet sync

Zk2u:
## Description

Adds a simple exponential backoff to the operator wallet sync process

This current configuration does:
100ms, 300ms, 900ms, 2.7s, 8.1sec, error

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

STR-1585

### DIFF
--- a/crates/operator-wallet/src/lib.rs
+++ b/crates/operator-wallet/src/lib.rs
@@ -18,9 +18,9 @@ use tracing::{error, info};
 
 /// How many times we should reattempt after an error during a wallet sync
 const SYNC_RETRIES: u32 = 5;
-/// The wallet will delay a retry by SYNC_BASE_DELAY*2^current_retry,
+/// The wallet will delay a retry by SYNC_BASE_DELAY*SYNC_BACKOFF^current_retry,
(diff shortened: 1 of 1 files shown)


Rajil1213 (dismissed):
You should use the retry strategy from the algebra crate which was created precisely for future use cases like this one. It has tests and examples to guide you through the api. It also has an exponential backoff strategy that you can use.

Rajil1213:
> I cannot use the retry function as the generator it needs uses a static lifetime but we are limited to having self's lifetime because we need to mutably access our own fields.

That's unfortunate but I think the solution is to add a combinator in the `retry` module that allows mutable state access. That and `Box::pin` should get it working.

storopoli (approved):
ACK 165ce484a7df92e8226fb280553a1242e3fd40a2

ProofOfKeags:
> As discussed in meeting, this works and works is better than not working.

Dropping knowledge.

PR #224, opened 2025-07-21
Bugfix/unbounded ouroboros

ProofOfKeags:
## Description

This makes it such that the ouroboros system uses unbounded mpsc channels instead of bounded broadcast channels.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/bin/alpen-bridge/src/mode/operator.rs
+++ b/bin/alpen-bridge/src/mode/operator.rs
@@ -54,7 +54,7 @@ use strata_bridge_stake_chain::prelude::OPERATOR_FUNDS;
 use strata_p2p::swarm::handle::P2PHandle;
 use strata_p2p_types::{P2POperatorPubKey, StakeChainId};
 use strata_tasks::TaskExecutor;
-use tokio::{net::lookup_host, select, sync::broadcast, task::JoinHandle};
+use tokio::{net::lookup_host, select, sync::mpsc, task::JoinHandle};
 use tracing::{debug, error, info};
 
 use crate::{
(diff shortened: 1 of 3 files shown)


storopoli (changes requested):
This changes the API.
`mpsc` is multi producer and single consumer.
Whereas [`broadcast`](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) is multi producer multi consumer.

Maybe instead of
```rs
 broadcast::channel(1);
```

why not?

```rs
 broadcast::channel(usize::MAX);
```

This can hugely increase the memory usage.

Maybe?

```rs
 broadcast::channel(8_192);
```

or other power of 2.

ProofOfKeags:
It should never have been a broadcast channel in the first place. We only want to read ouroboros events in a single place.

storopoli (approved):
> It should never have been a broadcast channel in the first place. We only want to read ouroboros events in a single place.

Given that we have a single consumer,

ACK 009e9237e4af42a75765e2525c1497e6474b9fd9

Rajil1213 (approved):
The caveat to this change is that now, the ouroboros mechanism _must_ have a single consumer so we cannot have any subsystems that can subscribe to self-published messages. I don't think we'll ever need it though so given the current issues, I think it's worth the sacrifice. There are also ways to have a single ouroboros receiver that can then broadcast messages to subsystems if we ever need that functionality.

PR #225, opened 2025-07-21
fix: move all transactions to V3

storopoli:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist


Rajil1213 (changes requested):
I probably should have been more precise with the ticket description, but this is not going to fix the issue that spawned that ticket.

The requirement for the ticket is that we change all transactions created by the `operator_wallet` to v3 instead of v2. It does not require changing anything in any of the crates currently being changed in this PR.

The only transaction created by the `operator_wallet` that may be affected by this change is the claim funding tx which can have multiple outputs of 27720 sats each. It'd be nice to figure out how many such outputs we can add to this transaction before we hit the lower vsize bound on v3 transactions.

PR #236, opened 2025-07-29
fix: wrong cd path in certs loop

mdqst:


storopoli:
@Zk2u I see that `ca-and-tls.sh` is not called anywhere in the codebase? Does it make sense to keep it? Should we also add it to `.justfile`?

Rajil1213:
> I see that ca-and-tls.sh is not called anywhere in the codebase

Yes, the `gen_s2_tls.sh` script largely replaces this file and we should just remove it at this point.

@mdqst how did you come across this bug/issue? Is it something mentioned in any docs?

PR #247, opened 2025-08-01
perf(alpen-bridge): don't clone the `cache`!

storopoli:
## Description

Avoid cloning the `cache` in the RPC server on `alpen-bridge`.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

`just pr` got some `taplo` formatting in one forsaken `Cargo.toml` so I'm adding this in a different commit.

## Checklist

## Related Issues

STR-1625

### DIFF
--- a/bin/alpen-bridge/src/rpc_server.rs
+++ b/bin/alpen-bridge/src/rpc_server.rs
@@ -432,9 +432,9 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
     }
 
     async fn get_deposit_requests(&self) -> RpcResult<Vec<Txid>> {
-        let all_entries = self.cached_contracts.read().await.clone();
+        let cached_contracts = self.cached_contracts.read().await;
 
-        let deposit_requests = all_entries
+        let deposit_requests = cached_contracts
             .iter()
             .map(|entry| entry.1.deposit_request_txid())
             .collect();
@@ -446,10 +446,9 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
         &self,
         deposit_request_txid: Txid,
     ) -> RpcResult<RpcDepositInfo> {
-        // Use the cached contracts
-        let all_entries = self.cached_contracts.read().await.clone();
+        let cached_contracts = self.cached_contracts.read().await;
 
-        for entry in all_entries {
+        for entry in cached_contracts.iter() {
             let entry_deposit_request_txid = entry.1.deposit_request_txid();
             if deposit_request_txid == entry_deposit_request_txid {
                 let status = match &entry.0.state.state {
@@ -474,10 +473,9 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
     }
 
     async fn get_bridge_duties(&self) -> RpcResult<Vec<RpcBridgeDutyStatus>> {
-        // we don't care about the hot state here, we only care about the database
-        let all_entries = self.cached_contracts.read().await;
+        let cached_contracts = self.cached_contracts.read().await;
 
-        let duties = all_entries
+        let duties = cached_contracts
             .iter()
             .filter_map(|entry| {
                 match entry.0.state.state {
@@ -505,12 +503,11 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
         &self,
         operator_pk: PublicKey,
     ) -> RpcResult<Vec<RpcBridgeDutyStatus>> {
-        // Use the cached contracts
-        let all_entries = self.cached_contracts.read().await;
+        let cached_contracts = self.cached_contracts.read().await;
 
         // NOTE: duties by operator pk is only for withdrawal duties,
         //       it does not make sense for deposit duties
-        let duties = all_entries
+        let duties = cached_contracts
             .iter()
             .filter_map(|entry| {
                 // Get the operator p2p key from the operator table
@@ -544,10 +541,10 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
     }
 
     async fn get_withdrawals(&self) -> RpcResult<Vec<Buf32>> {
-        let all_entries = self.cached_contracts.read().await.clone();
+        let cached_contracts = self.cached_contracts.read().await;
 
         let mut withdrawals = Vec::new();
-        for entry in all_entries {
+        for entry in cached_contracts.iter() {
             // NOTE: this is a source of bugs, don't use the `_` to match all.
             match &entry.0.state.state {
                 ContractState::Assigned {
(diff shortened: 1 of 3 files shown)


krsnapaudel (approved):

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -685,47 +714,22 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
 
     async fn get_claims(&self) -> RpcResult<Vec<Txid>> {
         // Use the cached contracts
-        let all_entries = self.cached_contracts.read().await.clone();
+        let cache = self.cached_contracts.read().await.clone();
```
ProofOfKeags, line 717: I very strongly question whether cloning the entire cache in stack memory is a good idea.

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -685,47 +714,22 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
...
-            .find(|entry| {
-                let claim_txids = entry.0.state.state.claim_txids();
+        let cache = self.cached_contracts.read().await.clone();
 
-                claim_txids.contains(&claim_txid)
-            })
-            .map(|entry| contract_state_to_reimbursement_status(&entry.0.state.state))
+        let claim_info = Self::get_contracts_with_claims(&cache)
+            .find(|(_, entry_claim_txid)| *entry_claim_txid == claim_txid)
+            .map(|(entry, _)| contract_state_to_reimbursement_status(&entry.0.state.state))
```
ProofOfKeags, line 732: by the functor composition law: `map g . map f == map (g . f)`

ProofOfKeags (commented):
Good PR. Unification is 🎩

Just some nits about some code that isn't technically in this diff but I noticed because of this diff.

Rajil1213:
> I have no idea why. I went over the logic a lot of times.

Is the issue reproducible, @storopoli?

storopoli:
> > I have no idea why. I went over the logic a lot of times.
> 
> Is the issue reproducible, @storopoli?

I don't know. I had several interactions with @krsnapaudel and asked him if the issue is really in the server. It was guaranteed that it was. This is the only thing that I find that could be the issue. Even if we don't fix this in this PR, unifying the logic for `claims` and `claimInfo` is a good idea anyways.

Rajil1213:
No, I meant did you try running it all locally to see if you get the same behavior? With the mock checkpoint, this should be relatively straightforward to check.

bin/alpen-bridge/src/rpc_server.rs
storopoli: Right! Amended in fc78f9b51e347b27576268060c630ae5f4ec0103

ProofOfKeags (approved):

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -447,9 +476,9 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
         deposit_request_txid: Txid,
     ) -> RpcResult<RpcDepositInfo> {
         // Use the cached contracts
-        let all_entries = self.cached_contracts.read().await.clone();
+        let cache = self.cached_contracts.read().await;
 
-        for entry in all_entries {
+        for entry in &*cache {
```
Rajil1213, line 481: Does this not work?
```suggestion
        for entry in cache.iter() {
```

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -600,133 +629,108 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
...
-        let claim_info = all_entries
-            .iter()
-            .find(|entry| {
-                let claim_txids = entry.0.state.state.claim_txids();
-
-                claim_txids.contains(&claim_txid)
-            })
-            .map(|entry| contract_state_to_reimbursement_status(&entry.0.state.state))
-            .map(|status| RpcClaimInfo { claim_txid, status });
+        // Use the cached contracts without cloning
```
Rajil1213, line 725: This comment is redundant. If you insist on adding `without cloning` then we should be doing it everywhere we're read-locking the `cached_contracts` to maintain consistency.

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -544,10 +573,10 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
     }
 
     async fn get_withdrawals(&self) -> RpcResult<Vec<Buf32>> {
-        let all_entries = self.cached_contracts.read().await.clone();
+        let cache = self.cached_contracts.read().await;
 
         let mut withdrawals = Vec::new();
-        for entry in all_entries {
+        for entry in &*cache {
```
Rajil1213, line 579: Same here.

```suggestion
        for entry in cache.iter() {
```

bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -685,48 +714,25 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
...
-        let claim_info = all_entries
-            .iter()
-            .find(|entry| {
-                let claim_txids = entry.0.state.state.claim_txids();
-
-                claim_txids.contains(&claim_txid)
-            })
-            .map(|entry| contract_state_to_reimbursement_status(&entry.0.state.state))
-            .map(|status| RpcClaimInfo { claim_txid, status });
+        let claim_info = Self::get_contracts_with_claims(&cache)
```
Rajil1213: This changes how this RPC works. In the current implementation, this RPC returns information on any claim regardless of whether it is part of an active graph whereas in this new implementation, it only returns information on claims that are associated with an active graph. This is a breaking change and not just a simple refactor.

Rajil1213 (changes requested):
I understand where @barakshani  is coming from.

But I have a slightly different issue with this PR. First, this behavior needs to be tested i.e., we must make sure this refactor hasn't broken any existing functionality. It would also be nice to check if the actual bug that birthed this PR is reproducible locally. If not, we may be looking for issues in the wrong place entirely. If it _is_ reproducible then we should get to the bottom of it. If it is an issue with the RPC implementation, I am okay patching this later but if it is an with the CSM logic or the database implementation itself, then this is a release blocker.

As it stands right now, I fully disagree with the `P1 - Blocks Release` and `needs backport` labels on this PR. Any PR that is not fixing an explicit bug is not a strict requirement for release. This is a `P2 - Nice to Have`.

bin/alpen-bridge/src/rpc_server.rs
barakshani: > In the current implementation, this RPC returns information on any claim regardless of whether it is part of an active graph

I think this isn't the case. As far as I can see, this PR doesn't change the logic for `get_claims`, only it moves it somewhere else. Actually previously we didn't return `Resolved` contract, which was changed in https://github.com/alpenlabs/strata-bridge/pull/242, and if I understand correctly, we still don't return claims in some states.

Rajil1213: > I think this isn't the case.

It is. The [current implementation](https://github.com/alpenlabs/strata-bridge/blob/b54f5b4325603b8f9b2631ecc9039ae0793067e4/bin/alpen-bridge/src/rpc_server.rs#L724) uses the [`claim_txids`](https://github.com/alpenlabs/strata-bridge/blob/b54f5b4325603b8f9b2631ecc9039ae0793067e4/crates/duty-tracker/src/contract_state_machine.rs#L3194) method defined on the CSM. This method gets all the claim txids for a given contract (regardless of whether it is active). Then, it only selects the contract which has the claim txid supplied in the argument to the RPC call. This PR changes it so that we instead use the new `get_contracts_with_claims` function which only selects the claim txids that are active.

The `claims_txids` RPC, however, only returns the active claims. And these claim txids are then used by the dashboard to fetch the claim info (afaict). The result is that we only see claim info for active claims but the claim info RPC itself is capable of returning the status even if the claim is not actually active (in the current impl). So, this PR "unifies" that behavior such that we only filter over active claims.

bin/alpen-bridge/src/rpc_server.rs
storopoli: Done in 5f75183dc5a78e24933a62b806fa0326aa26a1c3

storopoli:
Thanks for the review folks and discussion. I'll refactor this PR to only avoid the expensive clones in the `cache`.
I think we need further insights from the bridge monitoring dashboard that consumes the `claims`/`claimInfo` to understand why

> we still don't return claims in some states.

for the `claimInfo` RPC method.

barakshani:
Can I ask to make a change to the text [here](https://github.com/alpenlabs/strata-bridge/blob/main/crates/rpc/src/types.rs#L72)? probably just add "onchain" in the end. We can do better wording, but then we'll have to change subsequent text as well.

PR #250, opened 2025-08-06
fix: STR-1627 withdrawal data disappears after ~42 confirmations

krsnapaudel:
## Description

This fix is meant to address [STR-1627](https://alpenlabs.atlassian.net/browse/STR-1627).

This is likely not needed if monitoring dashboard caches withdrawal requests queried previously. See [this PR](https://github.com/alpenlabs/alpen-dashboards/pull/35).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This fix is similar to #242.

## Checklist


bin/alpen-bridge/src/rpc_server.rs
```diff
@@ -574,14 +574,20 @@ impl StrataBridgeMonitoringApiServer for BridgeRpc {
                 | ContractState::Asserted {
                     withdrawal_request_txid,
                     ..
+                }
+                // For internal monitoring, we may track withdrawals longer than
+                // when a contract gets resolved (e.g. 144 block confirmations).
+                // Without this change, withdrawals cannot be tracked after about 42 confirmations.
+                // This can be removed when monitoring backend saves txids from previous queries.
```
storopoli, line 581: We can remove this comment.

storopoli (changes requested):
I think this change is in-line with #242.
However:

1. We don't need the comment
2. Why 42 confirmations?
3. And why we need to include 42 in the commit message? (Despite being a number that I really like for obvious reasons: it's the meaning of everything).

barakshani (changes requested):
Are you sure this is the source of the issue? If so, as stated in the previous code comment, adding withdrawals is undesired, so we should live with that on the internal dashboard, and find an alternative fix (eg. the cache).

barakshani:
Thanks Dilli. I recommend not to change this RPC result, hence closing this PR.

krsnapaudel:
Don't need this fix.

PR #265, opened 2025-09-04
btc-notify: unify cursor sync and maintain continuity

Rajil1213:
## Description

This PR adds an additional property to the `btc-notify` crate such that it always emits blocks in a continuous stream (i.e., no gaps in block height). It also unifies the cursor management such that an initial height can be specified from which to start emitting blocks. This allows us to unify the cursor management for lagged blocks and recent blocks (after restarts). This is achieved by passing in a `fetcher` whose responsibility is to fetch blocks for a given height.

Possible additions:

- [x] Rename `BtcZmqSM` to `BtcNotifySM` since it is no longer tied solely to ZMQ.
- [ ] Add a global block height cursor with which to initialize the `btc-notify` client in case there are no contracts.
- [ ] Add crate-level documentation for the new `btc-tracker` crate along with its working and the invariants it maintains.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

The following manual tests were performed:

- Check that after a restart, the `btc-notify` client continues emitting blocks starting from the supplied `start_height`.
- Check that if the node misses one or more blocks that contain a deposit request due to some downtime, it is still able to recognize and process the deposit request when it comes back up.

## Checklist

## Related Issues

Related to STR-1606


Rajil1213:
Setting to draft until the regression is fixed.

crates/btc-notify/src/client.rs
ProofOfKeags, line 1: This is a legendary commit. Seriously well done type-theory flex.

crates/btc-notify/src/client.rs
```diff
@@ -181,33 +183,66 @@ impl BtcZmqClient<Disconnected> {
...
-                                    let (tx_events, block_event) = sm.process_block(block);
+                                    let mut blocks = Vec::new();
+
+                                    debug!("received block at height {received_height}, expected {cursor}");
+
+                                    // check if the height of this block is the one that we expect.
+                                    // and then, backfill any skipped blocks.
+                                    let mut blocks = Vec::new();
+                                    while cursor < received_height {
+                                        match fetcher.fetch_block(cursor).await {
```
ProofOfKeags, line 199: This can be done concurrently.

ProofOfKeags: In fact if you use a future join_all you'll both speed up the concurrency aspect as well as having a shorter expression and no mutation

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -161,26 +161,28 @@ impl ContractManager {
...
-            let mut cursor = active_contracts
+            let cursor = active_contracts
                 .values()
                 .min_by(|sm1, sm2| sm1.state().block_height.cmp(&sm2.state().block_height))
                 .map(|sm| sm.state().block_height)
                 .unwrap_or(current);
 
+            let fetcher = BtcFetcher(rpc_client.clone());
+            let zmq_client = zmq_client
+                .connect(cursor, fetcher)
```
ProofOfKeags, line 182: I actually think the cursor should be a property of the subscription and not the client overall. However, if you're finding you don't need that design for the current moment that's fine. However, with this design you now require reconstructing the client if you want multiple cursors. Use your judgement as to whether or not you think it's worth it to do now.

crates/duty-tracker/src/tx_driver.rs
```diff
@@ -314,6 +314,10 @@ mod e2e_tests {
 
     use super::TxDriver;
 
+    // TODO(proofofkeags): once rust-bitcoin@0.33.x lands this isn't necessary anymore. This is
+    // due to a bug in rust-bitcoin.
+    pub(crate) const BIP34_MIN_BLOCKS: usize = 17;
```
ProofOfKeags, line 319: TxDriver should be moved into btc-notify. It's really an extension of the subscription system, not a component of the "duty tracker"

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -167,14 +167,17 @@ impl ContractManager {
...
-            // re-feed chain events that they might have missed.
+            // re-feed chain events that they might have missed. We then start processing from the
+            // next block.
             let cursor = active_contracts
                 .values()
                 .min_by(|sm1, sm2| sm1.state().block_height.cmp(&sm2.state().block_height))
                 .map(|sm| sm.state().block_height)
-                .unwrap_or(current);
+                .unwrap_or(current.saturating_sub(1)) // -1 to nullify the +1 below
+                .saturating_add(1); // start at the next block
```
ProofOfKeags, line 177: It's not clear to me why this is needed.

bin/alpen-bridge/src/mode/operator.rs
```diff
@@ -166,15 +166,8 @@ pub(crate) async fn bootstrap(
...
-    let bury_height = current.saturating_sub(config.btc_zmq.bury_depth() as u64);
-
-    // we grab every block starting with the block after the bury_height all the way up to the
-    // current height and place it in the unburied blocks queue.
-    let mut unburied_blocks = VecDeque::new();
-    for height in bury_height + 1..=current {
-        unburied_blocks.push_front(bitcoin_rpc_client.get_block_at(height).await?);
-    }
+    // We have no awareness of what blocks are unburied at startup, so we start with an empty list.
+    let unburied_blocks = VecDeque::new();
```
ProofOfKeags, line 170: This is a serious semantics change, what was the rationale?

crates/btc-notify/src/client.rs
Rajil1213: Nice observation. Yeah I think we can do that. I'll make the change. Thanks!

crates/duty-tracker/src/contract_manager.rs
Rajil1213: The semantics of the `start_cursor` is that it is the height of the first block that is emitted by the client. So if the contracts have been synced till at least height `h`, we would initialize the client with `start_cursor = h + 1` to start receiving blocks from height `h + 1` onwards. If the node has been down for a while and the current height is `H`, the client would have to start at `H` (and not `H + 1`). But I think there is a better way to express this than what's currently in the code so that it's not as confusing. I'll see if I can come up with something better.

bin/alpen-bridge/src/mode/operator.rs
Rajil1213: There are a couple of reasons.

 With no concept of a `start_height`, starting from the first buried height (i.e., `current_height - bury_depth`) makes complete sense. However, when we introduce the `start_height`, I find it confusing that the first block emitted by the notifier is not the one at height `start_height` but rather at `start_height - bury_depth`.

 The second reason is that with the previous design, if the current height is at $h$, the first block delivered to the state machine is at height $h - \\text{bury_depth}$. When we unify the cursor, if the contracts have been synced till height $h'$ (s.t. $h > h'$), then the the blocks in the range $[h - \\text{bury_depth}, h']$ will be delivered to the state machine twice: once when the client tries to catch up to the tip and the next time by popping the unburied blocks queue.

There is also one more problem. Once a block is delivered to the state machine, the first thing it does when processing the block is to check if the previous block hash (of the current block) matches the last seen block hash. If it doesn't, it means there was a reorg and right now, the state machine just rejects that block. This wasn't a problem in the previous design because the lagged blocks were synced out-of-band. Once the cursor is unified, the blocks in the range $[\\text{start_height}, h]$ will all be rejected because their previous block hash does not match the hash at the front of the unburied blocks queue. This wasn't a problem in the previous design because the lagged blocks were synced out-of-band. Once the cursor is unified, the blocks in the range $[\\text{start_height}, h]$ will all be rejected because their previous block hash does not match the hash at the front of the unburied blocks queue.

crates/duty-tracker/src/contract_manager.rs
Rajil1213: See if this is better https://github.com/alpenlabs/strata-bridge/commit/9648ab9831b736ee8e56d9655f3cbb4ccaac2b09.

crates/btc-notify/src/client.rs
```diff
@@ -241,11 +252,10 @@ impl BtcZmqClient<Disconnected> {
                                         }
 
                                         all_tx_events.extend(tx_events);
-                                    }
 
-                                    // finally, increment the cursor after processing the latest
-                                    // block.
-                                    cursor += 1;
+                                        // increment the cursor with every block that is processed
+                                        cursor += 1;
```
ProofOfKeags, line 257: This cursor height will be wrong due to only incrementing by one instead of by the backfill size.

crates/duty-tracker/src/contract_manager.rs
```diff
@@ -516,7 +517,21 @@ impl BlockFetcher for BtcFetcher {
...
+        const MAX_RETRIES: usize = 10;
+        const INITIAL_DELAY: Duration = Duration::from_secs(1);
+        const MAX_DELAY: Duration = Duration::from_secs(60);
+        const MULTIPLIER: f64 = 2.0;
+
+        let retry_strategy = Strategy::exponential_backoff(INITIAL_DELAY, MAX_DELAY, MULTIPLIER)
+            .with_max_retries(MAX_RETRIES);
+
+        let client = self.0.clone();
+        retry_with(retry_strategy, move || {
```
ProofOfKeags, line 530: that retry combinator is starting to pull its weight isn't it.

crates/btc-notify/src/client.rs
```diff
@@ -192,21 +196,28 @@ impl BtcZmqClient<Disconnected> {
...
-                                    // check if the height of this block is the one that we expect.
-                                    // and then, backfill any skipped blocks.
-                                    let mut blocks = Vec::new();
-                                    while cursor < received_height {
-                                        match fetcher.fetch_block(cursor).await {
-                                            Ok(b) => blocks.push(b),
+                                    // Backfill any skipped blocks in the range:
+                                    // [cursor, received_height).
+                                    let mut blocks =
+                                        match join_all((cursor..received_height).map(|height| {
```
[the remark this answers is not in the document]
Rajil1213: 1k blocks is one week worth of blocks. That's some serious downtime. Any reason you feel that is reasonable?

crates/btc-notify/src/client.rs
ProofOfKeags: Then you need to move the cursor advancement inside the backfill block processing loop. It's currently outside it afaict.

crates/btc-notify/src/client.rs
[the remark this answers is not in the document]
ProofOfKeags: Because that's what we set the take back window to. Maybe we can get away with not batching these right now but I'd consider this a buried body and it deserves a flag planted

crates/btc-notify/src/client.rs
ProofOfKeags: I must have misread the diff. You are correct. This is fine. 😅

crates/btc-notify/src/client.rs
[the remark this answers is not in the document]
Rajil1213: I've added a `FIXME` comment [here](https://github.com/alpenlabs/strata-bridge/commit/949c50c730653636fae01cfa3793d97b7e2020dd) but I don't think 2Gb of data is unreasonable to process in one-go. Plus, if the node has been down for that long, it is expected that the node will consume more resources to process the backlog and depending upon how we deploy the node, it may be possible to allocate more resources temporarily as the node catches up.

ProofOfKeags (approved):

PR #425, opened 2026-03-30
fix(docker): chmod +x binaries in runtime stage to fix Permission denied on macOS buildx

madan-oss:
## Summary

When building on macOS (arm64) with `docker buildx` targeting `linux/amd64`, `COPY --from=builder` does not reliably preserve the executable bit. This causes a `Permission denied` crash at container startup on EKS:

```
/usr/local/bin/entrypoint.sh: 7: /usr/local/bin/strata-bridge: Permission denied
```

Fix: explicit `chmod +x` in the runtime stage of both Dockerfiles.

## Changes

**`docker/strata-bridge/Dockerfile`**
```dockerfile
RUN chmod +x /usr/local/bin/strata-bridge /usr/local/bin/entrypoint.sh
```

**`docker/asm-runner/Dockerfile`**
```dockerfile
RUN chmod +x /usr/local/bin/strata-asm-runner /usr/local/bin/entrypoint.sh
```

## Test plan

- [ ] Build both images on macOS with `docker buildx --platform linux/amd64`
- [ ] Verify containers start without `Permission denied`
- [ ] Confirm on Linux native builds still work (chmod is a no-op there)

### DIFF
--- a/docker/strata-bridge/Dockerfile
+++ b/docker/strata-bridge/Dockerfile
@@ -10,12 +10,12 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
     --mount=type=cache,target=/app/target \
     cargo b -r -p strata-bridge
 
-RUN --mount=type=cache,target=/app/target cp /app/target/release/strata-bridge /app/strata-bridge
+RUN --mount=type=cache,target=/app/target cp /app/target/release/strata-bridge /tmp/strata-bridge
 
 FROM --platform=linux/amd64 bridge-rt:latest AS runtime
 
 # Copy the built binaries from the builder stage
-COPY --from=builder /app/strata-bridge /usr/local/bin/strata-bridge
+COPY --from=builder /tmp/strata-bridge /usr/local/bin/strata-bridge
 COPY docker/strata-bridge/entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod +x /usr/local/bin/strata-bridge /usr/local/bin/entrypoint.sh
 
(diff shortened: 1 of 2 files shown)


Rajil1213 (dismissed):
The changes look good but I haven't ever observed the issue this change is supposed to address.

Rajil1213:
@madan-oss are the changes in this PR still required?

madan-oss:
I think yes. Without the change pod will fail with permission issue.

docker/strata-bridge/Dockerfile
```diff
@@ -10,12 +10,13 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
     --mount=type=cache,target=/app/target \
     cargo b -r -p strata-bridge
 
-RUN --mount=type=cache,target=/app/target cp /app/target/release/strata-bridge /app/strata-bridge
+RUN --mount=type=cache,target=/app/target cp /app/target/release/strata-bridge /tmp/strata-bridge
```
Rajil1213, line 13: Why is this change needed? If this is a real issue, shouldn't it also be a problem for the `secret-service` binary?

Rajil1213 (commented):
I think it's fine to have the `chmod +x` fix (at worst, it's a no-op). But I still don't get why we need to move the binary to `/tmp` and then, move it to `/usr/local/bin`, instead of moving it directly to `/usr/local/bin`.

madan-oss:
I am closing this PR as the image built from the GitHub workflow worked. No permission issue.

PR #478, opened 2026-04-10
STR 2760:  Dev CLI cleanup I

MdTeach:
## Description
* Switch BTC RPC client from `simple_http` to `minreq_http` transport to preserve the original hostname in the `Host` header, fixing connectivity through reverse proxies
* Allow specifying the operator node index for withdrawal assignment in mock checkpoints instead of always defaulting to zero.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update


bin/dev-cli/README.md
```diff
@@ -31,6 +31,7 @@ dev-cli create-and-publish-mock-checkpoint \
   --genesis-l1-height 101 \
   --ol-start-slot 0 \
   --ol-end-slot 1 \
+  --assignee-node-idx 0 \
```
Rajil1213, line 34: The `.justfile` recipe for this also needs updating.

params.toml
```diff
@@ -0,0 +1,21 @@
+network = "signet"
+bridge_out_addr = "0x5400000000000000000000000000000000000001"
+deposit_amount = 1000000000
+stake_amount = 100000000
+burn_amount = 10000000
+refund_delay = 1008
+stake_chain_delta = { Blocks = 6 }
```
Rajil1213, line 7: While we're cleaning stuff up, you might want to remove these unnecessary params (from this file and from `params.rs` as well). Afaict, the following are not needed anymore:

```plaintext
burn_amount
stake_chain_delta
```

Rajil1213 (changes requested):
Looks good overall. Just some minor changes required.

params.toml
MdTeach: In #479 this `parms.toml` will be removed

bin/dev-cli/README.md
MdTeach: Amended in [7a01b2a](https://github.com/alpenlabs/strata-bridge/pull/478/commits/7a01b2ae3535eaa8103bcf8ce39eabaa5cf0ec62)

.justfile
```diff
@@ -362,6 +362,7 @@ bridge-in:
         --btc-pass password \
         --params bin/dev-cli/params.toml \
         --ee-address 70997970C51812dc3A010C7d01b50e0d17dc79C8 # from anvil #2
+        --assignee-node-idx 0
```
Rajil1213: Due to the comment in the previous line, this should be added before that line and the newline escaped with `\`.

Rajil1213 (changes requested):
The justfile change needs fixing.

.justfile
Rajil1213: Also, this is in the wrong place. The update should be made to the `checkpoint` recipe, not the `bridge-in` recipe.

Rajil1213: Note that the default value of `0` should be passed just like the rest of the values in that recipe.

Rajil1213 (changes requested):

.justfile
Rajil1213: While you're at it, please update the commit message as well.

MdTeach: My bad. Fixed in [adb4e95](https://github.com/alpenlabs/strata-bridge/commit/adb4e958154aff77adc7cb34f4ba3c35ef9cc896)

Rajil1213 (approved):
Looks good now.

PR #484, opened 2026-04-13
STR-2994 Flexi Test: Add  reassignment lifecycle test

MdTeach:
## Description
* Adds assignment duration timeout reassignment test

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

### DIFF
--- a/functional-tests/tests/asm/fn_assignment_duration_timeout.py
+++ b/functional-tests/tests/asm/fn_assignment_duration_timeout.py
@@ -77,7 +77,7 @@ class AssignmentDurationTimeoutTest(StrataTestBase):
         # By default, it assigns to operator-0
         recent_block_hash = bitcoin_rpc.proxy.getblockhash(bitcoin_rpc.proxy.getblockcount())
         ckp_l1_txn = dev_cli.send_mock_checkpoint_from_tip(
-            asm_rpc, recent_block_hash, num_ol_slots=1, num_withdrawals=1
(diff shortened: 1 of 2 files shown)


functional-tests/tests/asm/fn_assignment_duration_timeout.py
```diff
@@ -0,0 +1,113 @@
...
+            timeout=300,
+            error_msg="ASM did not produce assignment",
+        )
+
+        assignments = asm_rpc.strata_asm_getAssignments(ckp_block_hash)
+        assignment = AssignmentEntry.from_dict(assignments[0])
+        self.logger.info(
+            f"Initial assignment: deposit_idx={assignment.deposit_entry.deposit_idx}, "
+            f"assignee=operator-{assignment.current_assignee}"
+        )
```
Rajil1213, line 97: We should also assert that the assignee is indeed operator 0 so that if the default in the dev-cli changes, we can update this test as well instead of this test silently passing.

Rajil1213 (changes requested):
Looks good. Just requires a slight change for reliability.

functional-tests/tests/asm/fn_assignment_duration_timeout.py
MdTeach: Amended in [9ad0492](https://github.com/alpenlabs/strata-bridge/pull/484/commits/9ad04924b17020fb6a7b37322ab498757c55a1b3)

Rajil1213 (approved):

PR #492, opened 2026-04-17
feat(orchestrator): integrate Stake SM

Rajil1213:
## Description

This PR integrates the Stake SM into the orchestrator pipeline. This involves the following changes:

- The Stake SM is added as a first class citizen in the `SMId`, `SMEvent` and `SMRegistry`.
- Events related to the Stake SM are classified and routed to the appropriate instance of Stake SM.
- Deposits are gated by the presence of each known operator's stake.
- Bugfixes in the SSM executor and operator wallet.
- The funding outpoint for the stake transaction is also persisted in the database.
- An RPC to get the status of stakes has been added (to use in the functional tests and monitoring).
- In the functional tests, the funding amount used to seed the operator's wallets has been increased.

~~These changes depend upon the changes in #490 which should be merged before this PR.~~

Remaining changes that will be handled in a subsequent PR:

 - [ ] When a preimage is revealed in the `Unstaking Intent` transaction, a signal has to be produced and routed to the Graph SM for every unused graph for the unstaking operator. This preimage has to be added to every state in the Graph SM. This is a more involved change (as we have to be aware of FDB limits as well if there are too many graphs) and will be handled via https://alpenlabs.atlassian.net/browse/STR-3114.

Codex and Claude were used to brainstorm the required changes and Claude was used to apply them.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

~~The first three commits on this branch are from #490.~~

All commits are atomic and any commit with a large number of changes also contains a summary in the commit description.

The final two commits are added for robustness of the tests as there was some flakiness.

## Checklist

## Related Issues

Closes STR-2924

### DIFF
--- a/crates/bridge-exec/src/stake/staking.rs
+++ b/crates/bridge-exec/src/stake/staking.rs
@@ -22,7 +22,7 @@ use strata_bridge_primitives::{
     types::OperatorIdx,
 };
 use strata_bridge_tx_graph::musig_functor::StakeFunctor;
-use tracing::{info, warn};
+use tracing::{error, info, warn};
 
 use crate::{
     chain::publish_signed_transaction, config::ExecutionConfig, errors::ExecutorError,
@@ -103,6 +103,14 @@ async fn create_stake_funding_tx(
     info!(%fee_rate, "creating stake funding transaction");
     let psbt = {
         let mut wallet = output_handles.wallet.write().await;
+
+        match wallet.sync().await {
+            Ok(()) => info!("synced wallet successfully"),
+            Err(e) => error!(
+                ?e,
+                "could not sync wallet before creating stake funding tx; still attempting"
+            ),
+        }
 
         wallet
             .create_stake_funding_tx(fee_rate)
--- a/crates/bridge-exec/src/stake/staking.rs
+++ b/crates/bridge-exec/src/stake/staking.rs
@@ -335,27 +335,18 @@ pub(crate) async fn publish_stake(
 
     let prevouts = Prevouts::All(&[prevout]);
     let mut sighash_cache = SighashCache::new(tx);
-    let mut signed_tx = tx.clone();
-    for (input_index, _) in tx.input.iter().enumerate() {
-        let sighash = create_key_spend_hash(
-            &mut sighash_cache,
-            prevouts.clone(),
-            TapSighashType::Default,
-            input_index,
-        )
+    let sighash = create_key_spend_hash(&mut sighash_cache, prevouts, TapSighashType::Default, 0)
         .expect("must be able to create key spend sighash");
 
-        let signature = output_handles
-            .s2_client
-            .stakechain_wallet_signer()
-            .sign(sighash.as_ref(), None)
-            .await
-            .map_err(ExecutorError::SecretServiceErr)?;
+    let signature = output_handles
+        .s2_client
+        .stakechain_wallet_signer()
+        .sign(sighash.as_ref(), None)
+        .await
+        .map_err(ExecutorError::SecretServiceErr)?;
 
-        signed_tx.input[input_index]
-            .witness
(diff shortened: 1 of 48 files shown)


uncomputable (commented):
5eaab670cf21c2c6cff9b658eb5b5abf9946b5ac The changes to `bridge-sm/stake` look good.  I only skimmed the other changes.

functional-tests/tests/asm/fn_asm_rpc_test.py
```diff
@@ -37,10 +38,17 @@ def main(self, ctx: flexitest.RunContext):
...
 
         bridge_rpc = bridge_rpcs[0]
+        self.logger.info("Waiting for all operators to complete staking before broadcasting DRT")
+        wait_until_all_operators_staked(
+            bridge_rpc,
+            bitcoin_rpc,
+            expected_operator_count=num_operators,
+        )
+
+        drt_txid = dev_cli.send_deposit_request()
```
MdTeach, line 50: This is a readiness precondition, not test logic. Could we move this into `get_bridge_nodes_and_rpcs` like we do for `wait_until_p2p_connected`, so new tests don't need to remember adding this precondition every time?

crates/bridge-exec/src/stake/staking.rs
```diff
@@ -284,14 +303,83 @@ pub(crate) async fn publish_unstaking_partials(
...
+            .s2_client
+            .stakechain_wallet_signer()
+            .sign(sighash.as_ref(), None)
+            .await
+            .map_err(ExecutorError::SecretServiceErr)?;
+
+        signed_tx.input[input_index]
+            .witness
+            .push(signature.serialize());
+    }
```
MdTeach, line 358: Here, `Prevouts::All(&[prevout])` with a `for (input_index, _) in tx.input.iter().enumerate()` loop is ambiguous — as `create_key_spend_hash` will panic `if prevouts.len() != tx.input.len()`. 

Instead with better comments can we just have this:
```rust
let prevouts = Prevouts::All(&[prevout]);
let mut sighash_cache = SighashCache::new(tx);
let sighash = create_key_spend_hash(
    &mut sighash_cache,
    prevouts,
    TapSighashType::Default,
    0,
)
.expect("must be able to create key spend sighash");
```

MdTeach (changes requested):
Overall looks good!!
Few suggestions:

functional-tests/tests/asm/fn_asm_rpc_test.py
Rajil1213: > This is a readiness precondition, not test logic. Could we move this into `get_bridge_nodes_and_rpcs` like we do for `wait_until_p2p_connected`, so new tests don't need to remember adding this precondition every time?

Good point. Will add it to the precondition check.

crates/bridge-exec/src/stake/staking.rs
Rajil1213: Good catch. I'll see if I can get the prevouts to also be an array or vice-versa.

functional-tests/tests/asm/fn_asm_rpc_test.py
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/5be0932d7f80eeb927b804244ae8c3ccd5d4fbe7.

crates/bridge-exec/src/stake/staking.rs
Rajil1213: Done in https://github.com/alpenlabs/strata-bridge/commit/d5ac1c872702aaa0dcdf565a8304300c59d8934c.

MdTeach (approved):

PR #494, opened 2026-04-21
feat: add adaptor verification logic

Rajil1213:
## Description

This PR integrates adaptor verification to the pre-existing deposit processing flow. This involves:

- Moving the operator (evaluator) adaptor keys and watchtower (garbler) fault pubkeys out of the config and into the `GraphData`.
- Fetching the actual adaptor and fault pubkeys from `mosaic`.
- Updating the `tx-graph` API to support multiple adaptor pubkeys (one per watchtower) per graph instance.
- Initializing the deposit on both the garbler (in `verify_adaptors` executor) and the evaluator (in `generate_graph_data` executor).
- Validating that the fault pubkey matches the ones sent by the operator as part of the `verify_adaptors` executor. 
- Removing the earlier `AdaptorsVerified` event fabrication in the orchestrator.

There are also some auxiliary changes:

- The deposit functional test now sends only 1 deposit request before a restart since deposit processing is slow enough that we can observe a post-restart nag.
- The deposit tableset initialization RPC/API has been made idempotent.
- Mosaic is updated to the latest commit on `main`.
- The test for concurrent deposits has been updated so that it does not depend on the DTs confirming in the same order as their DRTs.
- The CODEOWNERS file has been updated to include mosaic-specific files.

Claude was used for all code-level changes.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

All commits are atomic and come with explanations/reasoning (for complex changes).

## Checklist

## Related Issues

Closes STR-2666
Closes STR-2669


bin/dev-cli/src/handlers/graph.rs
```diff
@@ -22,8 +22,8 @@ pub(crate) fn build_game_graph(
         ),
         // TODO: <https://alpenlabs.atlassian.net/browse/STR-2945>
         // use the COUNTERPROOF_N_BYTES constant in a future refactor
-        // proof bytes (groth16) + deposit_idx (4 bytes) + operator pubkey (32 bytes)
-        counterproof_n_bytes: NonZero::new(128 + 32 + 4).expect("non-zero"),
+        // proof bytes (groth16) + deposit_idx (4 bytes)
+        counterproof_n_bytes: NonZero::new(128 + 4).expect("non-zero"),
```
MdTeach, line 26: Why are we removing the operator pubkey (32 bytes)
From the counterproof spec this is also one of the public params ?

Is this because operator pubkey is already known value ?

MdTeach (commented):
I haven’t fully reviewed this yet, but I have one clarifying question:

bin/dev-cli/src/handlers/graph.rs
Rajil1213: > Is this because operator pubkey is already known value ?

I'm not sure about this. But we need the sighashes of the counterproof to match the adaptor verification API in mosaic. The number of sighashes is restricted via [this](https://github.com/alpenlabs/strata-bridge/blob/49e04b514408b71b6b9c071f2582b79cf1913021/crates/mosaic-client-api/src/types.rs#L44) definition.

CC: @sapinb

MdTeach: Confirmed with @sapinb,
Its actually `128` only.
We should use [this](https://github.com/alpenlabs/strata-bridge/blob/49e04b514408b71b6b9c071f2582b79cf1913021/crates/mosaic-client-api/src/types.rs#L50) instead

Rajil1213: This change is required due to the definition of `DepositSighashes` whose length is `N_DEPOSIT_INPUT_WIRES` + `N_WITHDRAWAL_INPUT_WIRES` [here](https://github.com/alpenlabs/strata-bridge/blob/49e04b514408b71b6b9c071f2582b79cf1913021/crates/mosaic-client-api/src/types.rs#L44). Could you confirm this as well?

crates/connectors/src/contest_counterproof.rs
```diff
@@ -62,33 +72,38 @@ impl Connector for ContestCounterproofOutput {
...
-        for _ in 0..self.n_data.get() - 1 {
-            builder = builder
-                .push_opcode(opcodes::all::OP_TUCK)
-                .push_opcode(opcodes::all::OP_CHECKSIGVERIFY)
-                .push_opcode(opcodes::all::OP_CODESEPARATOR);
-        }
-
-        let counterproof_script = builder.push_opcode(opcodes::all::OP_CHECKSIG).into_script();
-        vec![counterproof_script]
+        self.operator_pubkeys
```
uncomputable: 2514dfa58bdcca85d025261823ae8ee899eafc5e: This is a massive change to the tx graph spec! We need to talk about this

uncomputable (changes requested):
NACK c4fb9a9688c42bd0e55b091cd41b1055a933ff57 Flagging this because we need to discuss the changes to the tx graph spec that are implemented here.

crates/connectors/src/contest_counterproof.rs
Rajil1213: > 2514dfa58bdcca85d025261823ae8ee899eafc5e: This is a massive change to the tx graph spec! We need to talk about this

I've amended that commit. Check out: https://github.com/alpenlabs/strata-bridge/commit/f94674eadda8100e61381f2624c5e09395365030. This keeps the connectors API intact but updates the `GameConnectors` to have a `Vec` of `ContestCounterproof` connectors (one per watchtower as each has its own adaptor key). This then changes the construction of the `GameGraph` as well.

bin/dev-cli/src/handlers/graph.rs
sapinb: Sighashes is needed for `N_DEPOSIT_INPUT_WIRES + N_WITHDRAWAL_INPUT_WIRES` (4 + 128)
counterproof (g16) is only `N_WITHDRAWAL_INPUT_WIRES` (128)

`counterproof_n_bytes` looks confusing, it seems to refers to the counterproof itself, and not the sighashes.

Rajil1213: Both are related. The number of bytes of the counterproof (including the public values) is the same number as the number of sighashes of the counterproof transaction that need to be signed (uniquely).

sapinb: Flow:
During deposit, bridge provides (4+128) sighashes.

During withdrawal, on garbler side, bridge provides the 128 byte counterproof, and gets 4 + 128 completed adaptor signatures to post onchain.
Evaluator side also receives 4 + 128 signatures to perform evaluation.

Rajil1213: Does the deposit index (4 bytes, public value) not have to be provided during withdrawal (on the garbler side)?

crates/bridge-sm/src/graph/transitions/uncontested.rs
```diff
@@ -101,12 +129,37 @@ impl GraphSM {
...
-                                graph_data_event.into(),
+                                graph_data_event.clone().into(),
                                 Some(format!(
                                     "Missing counterproof for watchtower {pov_operator_idx}"
                                 )),
                             )
                         })?;
+                    let adaptor_pubkey = *deposit_params
+                        .adaptor_pubkeys
+                        .get(pov_counterproof_idx)
```
sapinb, line 140: will this index remain stable after an operator gets kicked out following a fault?

bin/dev-cli/src/handlers/graph.rs
sapinb: Withdrawal is performed on a specific deposit. Mosaic already has the deposit index as the deposit inputs.

crates/mosaic-client/src/client.rs
```diff
@@ -154,30 +154,49 @@ impl<R: MosaicRpcClient + Send + Sync + 'static, P: MosaicIdResolver> MosaicClie
...
-                rpc.init_evaluator_deposit(tableset_id, rpc_deposit_id, deposit_config)
+                rpc.get_deposit_status(tableset_id, rpc_deposit_id)
                     .await
                     .map_err(MosaicError::rpc_error)
             }
         })
         .await?;
 
+        // Must be a new deposit to attempt init
+        if preexisting_status.is_none() {
```
sapinb, line 177: There is a small window where this check can return None but the init_evaluator_deposit call fails on duplicate deposit: A deposit for same id was already in mosaic state machine's input queue, but not yet processed. 
It is a very small window on an unlikely case though.

crates/mosaic-client/src/client.rs
```diff
@@ -206,26 +239,43 @@ impl<R: MosaicRpcClient + Send + Sync + 'static, P: MosaicIdResolver> MosaicClie
...
-                rpc.init_garbler_deposit(tableset_id, rpc_deposit_id, deposit_config)
+                rpc.get_deposit_status(tableset_id, rpc_deposit_id)
                     .await
                     .map_err(MosaicError::rpc_error)
             }
         })
         .await?;
 
+        // Must be a new deposit to attempt init
+        if preexisting_status.is_none() {
```
sapinb, line 260: same case as above

bin/dev-cli/src/handlers/graph.rs
Rajil1213: But isn't that only because the tablesets are indexed by the deposit index? For example, if the public values of the proof were something different or had more values (say, deposit txid), wouldn't we have had to pass those values as well?

crates/bridge-sm/src/graph/transitions/uncontested.rs
Rajil1213: Yes. For any operator in the set, their own pov is not subject to change when an operator leaves. For example, if an operator's index is 5, and the operator at index 4 leaves, their index still remains 5.

crates/mosaic-client/src/client.rs
Rajil1213: That can only happen in subsequent calls, right? The first call we ever make will succeed? If that's the case, we should be fine (any subsequent retries will be minutes apart).

functional-tests/tests/bridge/fn_deposit_test.py
```diff
@@ -97,16 +101,13 @@ def main(self, ctx: flexitest.RunContext):
...
-            self.logger.info("Observed post-restart nag emission in operator logs")
-        except TimeoutError:
-            self.logger.info("No post-restart nag emission observed within timeout")
+        wait_until_logs_match(
+            operator_log_offsets,
+            lambda line: "nagging peer for missing data" in line,
+            timeout=60,
+            error_msg="Timed out waiting for post-restart nag emission",
+        )
+        self.logger.info("Observed post-restart nag emission in operator logs")
```
MdTeach, line 111: The log check has gone from a soft observation (previously wrapped in try/except TimeoutError) to a hard assertion. 

Let's track a follow-up ticket to replace this with a proper signal rather than depending on a log substring.

bin/strata-bridge/src/mode/operator.rs
MdTeach, line 1: From 🤖 🤖: 

> The mosaic deposit poller is never spawned. [poll_watched_deposits](https://github.com/alpenlabs/strata-bridge/blob/STR-2669-integrate-mosaic-adaptor-verification/crates/mosaic-client/src/task.rs#L20) has no caller in bin/.
> 
> If mosaic is slow and [init_garbler_deposit](https://github.com/alpenlabs/strata-bridge/blob/STR-2669-integrate-mosaic-adaptor-verification/crates/mosaic-client/src/client.rs#L307-L315) hits the Incomplete branch, the deposit is inserted into watched_deposits expecting the poller to re-check it — but with no poller running, there's no retry and the GSM is stuck in GraphGenerated forever.

MdTeach (commented):
Looks good.
Left some comments.

bin/dev-cli/src/handlers/graph.rs
uncomputable: I think the confusion partly stems from the documentation of the contest-counterproof connector:

> `n_data` is the length of the serialized counterproof (including public values).
> This is equal to the number of required operator signatures.

This is wrong, since `n_data` is just the number of bytes that are revealed via adaptor signatures. The operator pubkey is revealed during the setup stage, without using adaptor signatures.

uncomputable: I'm happy to create a follow-up PR that clarifies the documentation.

Rajil1213: Sounds good. Do create a ticket first. This is not urgent.

uncomputable (dismissed):

functional-tests/tests/bridge/fn_deposit_test.py
Rajil1213: I'm not sure what ticket would look like to be honest. We'll have to connect to the bridge nodes via P2P and inspect whether there has been a nag message. 🤔 

Will add that as one of the approaches but feels like an overkill.

Rajil1213: Done: https://alpenlabs.atlassian.net/browse/STR-3182

bin/dev-cli/src/handlers/graph.rs
uncomputable: https://alpenlabs.atlassian.net/browse/STR-3183

bin/strata-bridge/src/mode/operator.rs
Rajil1213: Amazing catch! Fixed in https://github.com/alpenlabs/strata-bridge/commit/b26bc353a4ee63de977ff9fb4c16518fdb214266.

MdTeach (approved):

PR #495, opened 2026-04-21
STR 3084: STF event validation missing for Bridge Proof and Counterproof NACK transactions

MdTeach:
## Description
* Validates bridge proof tx spends the contest outpoint
* Adds spend-validation for CounterproofNackConfirmedEvent

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

### DIFF
--- a/crates/bridge-sm/src/graph/events.rs
+++ b/crates/bridge-sm/src/graph/events.rs
@@ -95,9 +95,6 @@ pub struct ContestConfirmedEvent {
 /// Event notifying that a bridge proof transaction has been confirmed on-chain.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BridgeProofConfirmedEvent {
-    /// The txid of the bridge proof transaction.
-    pub bridge_proof_txid: Txid,
-
     /// The block height at which the bridge proof transaction was confirmed.
     pub bridge_proof_block_height: BitcoinBlockHeight,
 
--- a/crates/bridge-sm/src/graph/events.rs
+++ b/crates/bridge-sm/src/graph/events.rs
@@ -150,6 +150,9 @@ pub struct CounterProofNackConfirmedEvent {
     /// The txid of the counterproof NACK transaction.
     pub counterproof_nack_txid: Txid,
 
+    /// The counterproof NACK transaction.
+    pub tx: Transaction,
+
     /// The index of the watchtower who submitted the corresponding counterproof transaction.
     pub counterprover_idx: OperatorIdx,
 }
--- a/crates/bridge-sm/src/graph/events.rs
+++ b/crates/bridge-sm/src/graph/events.rs
@@ -6,7 +6,7 @@
 
 use std::fmt::Display;
 
-use bitcoin::{OutPoint, Txid};
+use bitcoin::{OutPoint, Transaction, Txid};
 use bitcoin_bosd::Descriptor;
 use musig2::{PartialSignature, PubNonce};
 use strata_bridge_p2p_types::NagRequestPayload;
@@ -100,6 +100,9 @@ pub struct BridgeProofConfirmedEvent {
 
     /// The block height at which the bridge proof transaction was confirmed.
     pub bridge_proof_block_height: BitcoinBlockHeight,
+
+    /// The bridge proof transaction.
+    pub tx: Transaction,
 
     /// The bridge proof.
     pub proof: ProofReceipt,
--- a/crates/bridge-sm/src/graph/events.rs
+++ b/crates/bridge-sm/src/graph/events.rs
@@ -147,9 +147,6 @@ pub struct CounterProofAckConfirmedEvent {
 /// Event notifying that a counterproof NACK transaction has been confirmed on-chain.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CounterProofNackConfirmedEvent {
-    /// The txid of the counterproof NACK transaction.
-    pub counterproof_nack_txid: Txid,
-
     /// The counterproof NACK transaction.
     pub tx: Transaction,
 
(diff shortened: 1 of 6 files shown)


uncomputable (dismissed):
ACK 539499cba5f330e3b13a642bf327447223a9e6bc Looks good to me. Not much to see here

crates/bridge-sm/src/graph/transitions/contested.rs
```diff
@@ -93,6 +96,19 @@ impl GraphSM {
         cfg: Arc<GraphSMCfg>,
         event: BridgeProofConfirmedEvent,
     ) -> GSMResult<GSMOutput> {
+        // Ensure the bridge proof tx actually spends the contest proof connector.
+        let validate_spend = |contest_txid, event: BridgeProofConfirmedEvent| {
+            if spends_contest_proof_connector(contest_txid, &event.tx) {
```
Rajil1213: This isn't sufficient. We also need to check that the txid of the transaction does not match the `BridgeProofTimeout` transaction's txid (similar to the check for NACK).

crates/bridge-sm/src/graph/events.rs
```diff
@@ -150,6 +150,9 @@ pub struct CounterProofNackConfirmedEvent {
     /// The txid of the counterproof NACK transaction.
     pub counterproof_nack_txid: Txid,
```
Rajil1213: This needs to be removed. Since the source of truth is the actual `tx`, that is what we should use to compute the txid. Having this here can lead to footguns in the downstream impl.

crates/bridge-sm/src/graph/tests/contested/process_counterproof_nackd.rs
```diff
@@ -178,13 +202,68 @@ fn event_rejected_no_counterproof_posted() {
...
+    test_graph_invalid_transition(GraphInvalidTransition {
+        from_state: counter_proof_posted_state_with_nacks(&[]),
+        event: GraphEvent::CounterProofNackConfirmed(event),
+        expected_error: |e| matches!(e, GSMError::Rejected { .. }),
+    });
+}
+
+#[test]
+fn valid_nack_ignores_spoofed_event_txid() {
+    // The event's `counterproof_nack_txid` field is set to the ACK txid, but the STF
```
Rajil1213: This is a nice test and brought my attention to the duplication. But after removing the txid from the event, this test can also be removed.

Rajil1213 (changes requested):
This needs a couple of fixes and then, it should be good to go.

crates/bridge-sm/src/graph/transitions/contested.rs
MdTeach: Resolved in [60f8962](https://github.com/alpenlabs/strata-bridge/pull/495/commits/60f8962f44d1331b0c308ba1ff7cb8080e41d1c9)

crates/bridge-sm/src/graph/events.rs
MdTeach: Resolved in [60f8962](https://github.com/alpenlabs/strata-bridge/pull/495/commits/60f8962f44d1331b0c308ba1ff7cb8080e41d1c9)

crates/bridge-sm/src/graph/tests/contested/process_counterproof_nackd.rs
MdTeach: Resolved in [60f8962](https://github.com/alpenlabs/strata-bridge/pull/495/commits/60f8962f44d1331b0c308ba1ff7cb8080e41d1c9)

crates/bridge-sm/src/graph/tests/contested/process_bridge_proof.rs
```diff
@@ -372,6 +372,43 @@ fn event_rejected_wrong_outpoint_from_counterproof_posted() {
...
 }
 
+#[test]
+fn event_rejected_bridge_proof_timeout_txid_from_contested() {
+    // Use a valid bridge proof tx (spends contest proof connector) but set the
+    // event txid to the bridge proof timeout txid, simulating a misclassified
+    // timeout transaction.
+    let tx = test_bridge_proof_tx();
+    let event = BridgeProofConfirmedEvent {
+        bridge_proof_txid: TEST_GRAPH_SUMMARY.bridge_proof_timeout,
```
Rajil1213: Is this field even used anymore now that the whole `tx` is being passed as part of the `event`? If not, then this should be removed as well.

crates/bridge-sm/src/graph/transitions/contested.rs
```diff
@@ -541,6 +544,16 @@ impl GraphSM {
     }
 }
 
+/// Validates that the bridge proof tx spends the contest proof connector
+/// and is not the bridge proof timeout transaction.
+fn validate_bridge_proof_spend(
+    summary: &GameGraphSummary,
+    event: &BridgeProofConfirmedEvent,
+) -> bool {
+    event.bridge_proof_txid != summary.bridge_proof_timeout
```
Rajil1213: Since we're not checking whether the `bridge_proof_txid` matches `tx.compute_txid()`, we should definitely remove the `bridge_proof_txid`. In fact, I'm not even sure we need to pre-extract the proof but that's a more extensive change so can be left as is.

Rajil1213 (changes requested):
Similar to the NACK event, I think the bridge proof txid field also needs to be removed from the event payload.

crates/bridge-sm/src/graph/tests/contested/process_bridge_proof.rs
MdTeach: Fixed in [8c8d1dd](https://github.com/alpenlabs/strata-bridge/pull/495/commits/8c8d1ddc8698d53faea2e91b9a8ec93edd360e4c)

crates/bridge-sm/src/graph/transitions/contested.rs
MdTeach: Fixed in [8c8d1dd](https://github.com/alpenlabs/strata-bridge/pull/495/commits/8c8d1ddc8698d53faea2e91b9a8ec93edd360e4c)

Rajil1213 (approved):
Okay. This looks good to go now. Thanks!

PR #501, opened 2026-04-24
refactor(asm): adopt upstream alpenlabs/asm artifacts, drop local asm_runner

MdTeach:
## Description
This PR switches the project over to the upstream `alpenlabs/asm` artifacts and removes their now-redundant in-tree copies. Specifically:

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [x] Dependency Update


Rajil1213 (approved):
The changes look good. Should this be merged after your branch on the `asm` repo is merged?

Also, I think we should cache the `mosaic` and `asm` binaries in CI as well (just like how the `openssl` binary is cached). This can be done in a followup PR too.

Rajil1213 (changes requested):
Wait, let's also remove the asm-only functional tests, and update CODEOWNERS.

prajwolrg (dismissed):
Mostly looks good. Can you please use [`v0.1-alpha.6`](https://github.com/alpenlabs/asm/releases/tag/v0.1-alpha.6) `tag` instead of the `rev`.

MdTeach:
> Wait, let's also remove the asm-only functional tests, and update CODEOWNERS.

Updated the CODEOWNERS here [52a7c34](https://github.com/alpenlabs/strata-bridge/pull/501/commits/52a7c34ddb047d8d057a4061015dcaa7413d3a58). 
But we can’t remove the ASM-only functional tests because the ASM repository doesn’t include the corresponding functional tests.

MdTeach:
> Mostly looks good. Can you please use [`v0.1-alpha.6`](https://github.com/alpenlabs/asm/releases/tag/v0.1-alpha.6) `tag` instead of the `rev`.

Thanks for the new release.
Updated in [a1f45a6](https://github.com/alpenlabs/strata-bridge/pull/501/commits/a1f45a6607d82337fe6942165be0dc261eccaa80) & [7f11a81](https://github.com/alpenlabs/strata-bridge/pull/501/commits/7f11a816c0b6f0bd93c7ecc1b4e47d1f971584cf)

Rajil1213:
> But we can’t remove the ASM-only functional tests because the ASM repository doesn’t include the corresponding functional tests.

I strongly believe that the bridge repo shouldn't be owning end-to-end tests for ASM (for example, checking whether blocks progress) -- this is like having a test to see if all the bitcoin RPCs we're using also work.

However, I understand that we'll need to preempt any breaking changes in the absence of any such tests in our org. So, I'm fine keeping these around for the time being.

@prajwolrg could you let us know when you have functional tests for the `asm-runner` in the `asm` repo?

Rajil1213 (approved):
Looks good now. I'll implement the binary caching once this is merged.

PR #506, opened 2026-04-28
STR-3009: GSM: PublishCounterProofNack executor

MdTeach:
## Description

* Implements the `PublishCounterProofNack`
* Scopes the `PublishStake` duty to its owning operator so the executor skips foreign-operator publishes.
* Functional test driving the Counterproof → NACK → Contested-Payout path

### Type of Change
- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

### Notes to Reviewers
Please review the pull request as a whole rather than examining each commit individually.

## Related Issues
Closes STR-3009
Closes STR-3089

### DIFF
--- a/crates/bridge-exec/src/graph/counterproof_nack.rs
+++ b/crates/bridge-exec/src/graph/counterproof_nack.rs
@@ -125,18 +125,22 @@ fn decode_completed_sigs(
     // that follow the per-byte operator signatures in the counterproof witness.
     const WANT: usize = N + 3;
 
-    let witness = &counterproof_tx.input[0].witness;
-    if witness.len() != WANT {
+    let witness_len = counterproof_tx.input[0].witness.len();
+    if witness_len != WANT {
         return Err(ExecutorError::InvalidTxStructure(format!(
-            "counterproof witness has {} elements, expected {WANT}",
-            witness.len(),
+            "counterproof witness has {witness_len} elements, expected {WANT}"
         )));
     }
 
-    let sigs: [Signature; N] = std::array::from_fn(|i| {
-        // operator_signatures were pushed `.rev()`, so undo that ordering.
-        Signature::from_slice(&witness[N - 1 - i])
-            .expect("on-chain counterproof signature must parse")
-    });
-    Ok(sigs)
+    // The witness layout is `[sig_{N-1}, .., sig_0, n-of-n sig, leaf script, control
+    // block]` — operator signatures pushed `.rev()`, then 3 trailing items. Reverse +
+    // skip(3) recovers `[sig_0, .., sig_{N-1}]`.
+    let mut items = counterproof_tx.input[0].witness.to_vec();
+    items.reverse();
+    let sigs: Vec<Signature> = items
+        .into_iter()
+        .skip(3)
+        .map(|w| Signature::from_slice(&w).expect("on-chain counterproof signature must parse"))
+        .collect();
+    Ok(sigs.try_into().expect("witness length validated above"))
 }
--- a/crates/bridge-exec/src/graph/counterproof_nack.rs
+++ b/crates/bridge-exec/src/graph/counterproof_nack.rs
@@ -96,7 +96,7 @@ async fn sign_and_broadcast_nack(
     info!(
         %deposit_idx,
         %counterprover_idx,
-        elapsed_ms = evaluate_and_sign_start.elapsed().as_millis() as u64,
+        elapsed = ?evaluate_and_sign_start.elapsed(),
         "mosaic evaluate_and_sign completed",
     );
 
--- a/crates/bridge-exec/src/graph/counterproof_nack.rs
+++ b/crates/bridge-exec/src/graph/counterproof_nack.rs
@@ -1,0 +1,249 @@
+//! Executor for the counterproof NACK transaction.
+
+use bitcoin::{
+    Amount, OutPoint, Sequence, TapSighashType, Transaction, TxIn, TxOut,
+    hashes::Hash,
+    sighash::{Prevouts, SighashCache},
+};
+use btc_tracker::event::TxStatus;
+use musig2::secp256k1::schnorr::Signature;
+use secret_service_proto::v2::traits::{SchnorrSigner, SecretService};
+use strata_bridge_primitives::{
+    scripts::{prelude::create_key_spend_hash, taproot::TaprootTweak},
+    types::{DepositIdx, OperatorIdx},
+};
+use strata_bridge_tx_graph::transactions::prelude::CounterproofNackTx;
+use strata_mosaic_client_api::types::{
+    CompletedSignatures, G16ProofRaw, N_DEPOSIT_INPUT_WIRES, N_WITHDRAWAL_INPUT_WIRES, Sighash,
+    Tweak,
+};
+use tracing::{info, warn};
+
+use crate::{
+    chain::publish_signed_transaction, errors::ExecutorError, output_handles::OutputHandles,
+};
+
+/// Signs and publishes the counterproof NACK transaction to reject an invalid counterproof.
+pub(super) async fn publish_counterproof_nack(
+    output_handles: &OutputHandles,
+    deposit_idx: DepositIdx,
+    counter_prover_idx: OperatorIdx,
+    counterproof_tx: Transaction,
+    counterproof_nack_tx: CounterproofNackTx,
+) -> Result<(), ExecutorError> {
+    info!(%deposit_idx, %counter_prover_idx, "preparing counterproof nack");
+
+    let completed_signatures = decode_completed_sigs(&counterproof_tx)?;
+
+    let AttachedFunding {
+        nack_tx,
+        funding_outpoint,
+        prevouts,
+    } = attach_operator_funding(output_handles, counterproof_nack_tx).await?;
+
+    let result = sign_and_broadcast_nack(
+        output_handles,
+        counter_prover_idx,
+        deposit_idx,
+        completed_signatures,
+        nack_tx,
+        &prevouts,
+    )
+    .await;
+
+    if result.is_err() {
+        output_handles
+            .wallet
+            .write()
+            .await
+            .release_outpoints(&[funding_outpoint]);
+    }
+    result
+}
+
+/// Output of [`attach_operator_funding`]: the fully-populated NACK plus the leased funding
+/// outpoint (so the caller can release on failure) and a snapshot of all prevouts (needed
+/// because [`CounterproofNackTx::finalize_partial`] consumes the NACK by value).
+struct AttachedFunding {
+    nack_tx: CounterproofNackTx,
+    funding_outpoint: OutPoint,
+    prevouts: Vec<TxOut>,
+}
+
+/// Picks an operator-funded UTXO from the general wallet, leases it, and pushes a funding
+/// input + change output onto the NACK template. Single wallet write-lock — no race window
+/// between selection and the change-script lookup.
+async fn attach_operator_funding(
+    output_handles: &OutputHandles,
+    mut nack_tx: CounterproofNackTx,
+) -> Result<AttachedFunding, ExecutorError> {
+    const NACK_FIXED_FEE: Amount = Amount::from_sat(1_000);
+    const NACK_MIN_FUNDING_VALUE: Amount = Amount::from_sat(10_000);
+
+    let (funding_outpoint, funding_prevout, change_script) = {
+        let mut wallet = output_handles.wallet.write().await;
+        if let Err(e) = wallet.sync().await {
+            warn!(
+                ?e,
+                "could not sync wallet before nack funding selection; continuing"
+            );
+        }
+        let selected = wallet
+            .select_and_lease_general_utxo(|u| u.txout.value >= NACK_MIN_FUNDING_VALUE)
+            .ok_or_else(|| {
+                ExecutorError::WalletErr(format!(
+                    "no general utxo >= {} sat for counterproof nack funding",
+                    NACK_MIN_FUNDING_VALUE.to_sat()
+                ))
+            })?;
+        let change_script = wallet.general_script_buf().clone();
+        let outcome = (selected.outpoint, selected.txout, change_script);
+        drop(wallet);
+        outcome
+    };
+
+    let change_value = funding_prevout
+        .value
+        .checked_sub(NACK_FIXED_FEE)
+        .ok_or_else(|| {
+            ExecutorError::WalletErr(format!(
+                "funding utxo value {} sat is below fixed nack fee {} sat",
+                funding_prevout.value.to_sat(),
+                NACK_FIXED_FEE.to_sat(),
+            ))
+        })?;
+
+    nack_tx.push_input(
+        TxIn {
+            previous_output: funding_outpoint,
+            sequence: Sequence::MAX,
+            ..Default::default()
+        },
+        funding_prevout,
+    );
+    nack_tx.push_output(TxOut {
+        value: change_value,
+        script_pubkey: change_script,
+    });
+
+    let prevouts = nack_tx.prevouts().to_vec();
+    Ok(AttachedFunding {
+        nack_tx,
+        funding_outpoint,
+        prevouts,
+    })
+}
+
+/// Has mosaic sign the connector input, signs the funding input via the operator's general
+/// wallet, and broadcasts. The funding input is always the last one (`push_input` appends).
+#[allow(clippy::large_types_passed_by_value)] // mosaic's evaluate_and_sign takes CompletedSignatures by value
+async fn sign_and_broadcast_nack(
+    output_handles: &OutputHandles,
+    counter_prover_idx: OperatorIdx,
+    deposit_idx: DepositIdx,
+    completed_signatures: CompletedSignatures,
+    nack_tx: CounterproofNackTx,
+    prevouts: &[TxOut],
+) -> Result<(), ExecutorError> {
+    // Sighash + tap tweak for the connector input (key-path spend with `wt_i_fault` as the
+    // internal key; script-path is impossible by construction).
+    let signing_info = nack_tx.signing_info_partial();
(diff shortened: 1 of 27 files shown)


crates/bridge-sm/src/graph/events.rs
```diff
@@ -128,12 +128,12 @@ pub struct BridgeProofTimeoutConfirmedEvent {
...
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CounterProofConfirmedEvent {
-    /// The txid of the counterproof transaction.
-    pub counterproof_txid: Txid,
-
     /// The block height at which the counterproof transaction was confirmed.
     pub counterproof_block_height: BitcoinBlockHeight,
 
+    /// The counterproof transaction.
+    pub tx: Transaction,
```
Rajil1213, line 135: This should be fine but you only really need to pass the signatures and not the entire transaction, right?

crates/bridge-exec/src/stake/staking.rs
```diff
@@ -305,8 +305,15 @@ pub(crate) async fn publish_unstaking_partials(
...
     cfg: &ExecutionConfig,
     output_handles: &OutputHandles,
+    operator_idx: OperatorIdx,
     tx: &Transaction,
 ) -> Result<(), ExecutorError> {
+    let pov_idx = output_handles.operator_table.pov_idx();
+    if operator_idx != pov_idx {
+        debug!(%operator_idx, %pov_idx, "skipping PublishStake duty for foreign operator");
+        return Ok(());
+    }
```
Rajil1213: This is one way to do it but a better fix is to have the retry handler not produce this duty at all if the stake does not belong to the PoV operator. Since this check does not involve any side-effects/external calls, there is no need to have an execution gate on the executor rather than on the duty creator. It will also allow us to add a unit test to make sure we properly check this and catch any regressions.

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/// Picks an operator-funded UTXO from the general wallet, leases it, and pushes a funding
+/// input + change output onto the NACK template. Single wallet write-lock — no race window
+/// between selection and the change-script lookup.
+async fn attach_operator_funding(
+    output_handles: &OutputHandles,
+    mut nack_tx: CounterproofNackTx,
+) -> Result<AttachedFunding, ExecutorError> {
+    const NACK_FIXED_FEE: Amount = Amount::from_sat(1_000);
```
Rajil1213: You should check the fee rate from `bitcoind` and then fallback to the `max_fee_rate` in the config if it's too high. 

Implementing this is a bit more involved though since you'll need to tune the output to match the fee rate. This can be done in a future PR. Do create a ticket.

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -0,0 +1,249 @@
...
+    Ok(AttachedFunding {
+        nack_tx,
+        funding_outpoint,
+        prevouts,
+    })
+}
+
+/// Has mosaic sign the connector input, signs the funding input via the operator's general
+/// wallet, and broadcasts. The funding input is always the last one (`push_input` appends).
+#[allow(clippy::large_types_passed_by_value)] // mosaic's evaluate_and_sign takes CompletedSignatures by value
```
Rajil1213: Use `expect(..)` instead of `allow`.

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -0,0 +1,249 @@
...
+        .await
+        .map_err(|e| {
+            warn!(%deposit_idx, %counter_prover_idx, ?e, "evaluate_and_sign failed");
+            ExecutorError::MosaicErr(format!("evaluate_and_sign: {e:?}"))
+        })?
+        .ok_or_else(|| {
+            ExecutorError::MosaicErr(
+                "evaluator failed to extract fault secret from counterproof".into(),
+            )
+        })?;
```
Rajil1213, line 95: Can you also add a log after this completes? I want to check both here and in our infra how long this call actually takes.

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -0,0 +1,249 @@
...
+            witness.len(),
+        )));
+    }
+
+    let sigs: [Signature; N] = std::array::from_fn(|i| {
+        // operator_signatures were pushed `.rev()`, so undo that ordering.
+        Signature::from_slice(&witness[N - 1 - i])
+            .expect("on-chain counterproof signature must parse")
+    });
+    Ok(sigs)
```
Rajil1213: Suggestion: This whole logic might be easier to read and there'll be less of a chance of an off-by-one error, if you reverse the witness `Vec` first, `.skip(3)` and then, `try_into::<[Signature; N>().ok_or_else(|| ExecutorErr::InvalidTxStructure(..))?`.

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -0,0 +1,249 @@
...
+    )
+    .await;
+
+    if result.is_err() {
+        output_handles
+            .wallet
+            .write()
+            .await
+            .release_outpoints(&[funding_outpoint]);
+    }
```
Rajil1213: You should release the outpoint even if the call succeeds. There is no need to track this outpoint in memory outside the wallet anymore once it has been spent.

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
```diff
@@ -0,0 +1,169 @@
...
+    2. Submit a contest against the active claim
+    3. Wait for the bridge proof to be posted
+    4. Wait for the counterproof to be posted by watchtowers
+    5. Verify the deposit UTXO is spent after the NACK + contested-payout path completes
+    """
+
+    def __init__(self, ctx: flexitest.InitContext):
+        self.bridge_protocol_params = BridgeProtocolParams(
+            contest_timelock=5,
+            ack_timelock=10,
```
Rajil1213: I recommend increasing the `ack_timelock` further (like `100`) so that we can be sure that if the counterproof output is spent, it is more likely to be via the NACK path (and not ACK).

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # With NeverAccept predicate, watchtowers reject the proof and publish counterproofs.
+        wait_until_counter_proof_posted(pov_rpc, active_claim.deposit_idx)
+        self.logger.info("Counterproof posted — watchtowers rejected the invalid bridge proof")
+
+        # The POV operator runs evaluate_and_sign via mosaic and publishes a NACK for each
+        # watchtower's counterproof. Wait for the deposit UTXO to be spent — this fires once the
+        # full contested-payout path (including the NACKs) completes and the operator sweeps the
+        # deposit.
+        wait_until_deposit_utxo_spent(bitcoin_rpc, deposit_txid, timeout=450)
```
Rajil1213, line 190: We should first wait for the ACK-NACK vout of the counterproof to be spent.

functional-tests/utils/withdrawal.py
```diff
@@ -132,3 +132,27 @@ def check():
             f"Claim phase for deposit {deposit_idx} did not advance to bridge_proof_timedout"
         ),
     )
+
+
+def wait_until_all_nackd(
```
Rajil1213: Looks like this function is not used. YAGNI.

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -29,124 +24,46 @@ pub(super) async fn publish_counterproof_nack(
...
+    let payout_script = output_handles
+        .wallet
+        .read()
+        .await
+        .general_script_buf()
+        .clone();
+    counterproof_nack_tx.push_output(TxOut {
+        value: connector_value,
+        script_pubkey: payout_script,
+    });
```
Rajil1213, line 41: This technically violates the spec but I think it's fine. The NACK transaction is not part of the covenant (once the private key is known via mosaic, there should be no restriction to how the output is to be spent).

crates/bridge-sm/src/graph/tx_classifier.rs
```diff
@@ -221,6 +222,16 @@ impl TxClassifier for GraphSM {
...
+                } else if let Some(counterprover_idx) =
+                    counterproof_operator_idx(graph_summary, &txid, self.context().operator_idx())
+                {
+                    Some(GraphEvent::CounterProofConfirmed(
+                        CounterProofConfirmedEvent {
+                            counterproof_block_height: height,
+                            tx: tx.clone(),
+                            counterprover_idx,
+                        },
+                    ))
```
Rajil1213, line 234: 🫣

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
```diff
@@ -159,10 +165,28 @@ def main(self, ctx: flexitest.RunContext):
...
+                timeout=300,
+            )
+            self.logger.info(
+                f"Counterproof {counterproof_txid} ack/nack output spent (NACK published)"
+            )
+
+        # Wait for the deposit UTXO to be spent — this fires once the full contested-payout path
+        # (including the NACKs) completes and the operator sweeps the deposit.
         wait_until_deposit_utxo_spent(bitcoin_rpc, deposit_txid, timeout=450)
         self.logger.info("Deposit UTXO confirmed spent after counterproof NACK + contested payout")
```
Rajil1213: Note that this check is necessary to ensure that it was actually the NACK transaction (and not ACK) that spent the ACK-NACK vout (asserted before this).

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
Rajil1213: This can be resolved now.

crates/bridge-sm/src/graph/transitions/contested.rs
```diff
@@ -356,7 +356,7 @@ impl GraphSM {
                 // ACK/NACK output and is not a known counterproof ACK.
                 if !validate_counterproof_nack(
                     &graph_summary,
-                    self.context().operator_table().pov_idx(),
+                    self.context().operator_idx(),
```
Rajil1213, line 359: There should be an accompanying unit test since this is a bug.

Rajil1213 (changes requested):
Overall looks good. Only needs some semantic fixes in code, some more unit tests and minor changes in the functional tests. Some of the reviews left on the earlier commits (for older changes) can be resolved/ignored.

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
Rajil1213: This is an extra safety guarantee and not a must.

crates/bridge-sm/src/graph/tx_classifier.rs
Rajil1213: Might be a good idea to add a unit test to specifically cover this case.

crates/bridge-sm/src/graph/events.rs
MdTeach: This would require moving decode_completed_sigs into the STF, which would add extra work for STF tests.

MdTeach: Leaving this as of now
We can revisit as cleanup task

Rajil1213: Sounds good. Do create a ticket.

crates/bridge-exec/src/graph/counterproof_nack.rs
MdTeach: Added in [3f60c24](https://github.com/alpenlabs/strata-bridge/pull/506/commits/3f60c24225e76e37de46b72923dddef2f12098da)

crates/bridge-exec/src/stake/staking.rs
MdTeach: Made the refactor: [38dc974](https://github.com/alpenlabs/strata-bridge/pull/506/commits/38dc974466633b4eb7d46015f1cdffcf14c246fd)

crates/bridge-exec/src/graph/counterproof_nack.rs
MdTeach: Refactored in. [861aba2](https://github.com/alpenlabs/strata-bridge/pull/506/commits/861aba26f06f831fa2c7388d4c5f5b0300a325aa)

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
MdTeach: Refactored in [ae90a79](https://github.com/alpenlabs/strata-bridge/pull/506/commits/ae90a79c091f630f1298ac22d08084f2f39be0c0)

functional-tests/utils/withdrawal.py
MdTeach: Refactored in [ae90a79](https://github.com/alpenlabs/strata-bridge/pull/506/commits/ae90a79c091f630f1298ac22d08084f2f39be0c0)

functional-tests/tests/uncontested_payout/fn_publish_counterproof_nack.py
MdTeach: Updated the comment [ae90a79](https://github.com/alpenlabs/strata-bridge/pull/506/commits/ae90a79c091f630f1298ac22d08084f2f39be0c0)

crates/bridge-sm/src/graph/transitions/contested.rs
MdTeach: Added in [ae90a79](https://github.com/alpenlabs/strata-bridge/pull/506/commits/ae90a79c091f630f1298ac22d08084f2f39be0c0)

crates/bridge-sm/src/graph/tx_classifier.rs
MdTeach: Added in [ae90a79](https://github.com/alpenlabs/strata-bridge/pull/506/commits/ae90a79c091f630f1298ac22d08084f2f39be0c0)

crates/bridge-sm/src/graph/events.rs
MdTeach: Forgot we have 🤖 
Added in [8892843](https://github.com/alpenlabs/strata-bridge/pull/506/commits/8892843492eace0b0192fdae14b9247b76ec33da)

crates/bridge-exec/src/graph/counterproof_nack.rs
```diff
@@ -96,6 +97,12 @@ async fn sign_and_broadcast_nack(
                 "evaluator failed to extract fault secret from counterproof".into(),
             )
         })?;
+    info!(
+        %deposit_idx,
+        %counterprover_idx,
+        elapsed_ms = evaluate_and_sign_start.elapsed().as_millis() as u64,
```
Rajil1213: This should be seconds instead. It will probably take several minutes in practice. Probably better to just display the actual `Duration` struct instead of converting it to any units (I think the `Display` impl on `Duration` takes care of the format).

crates/bridge-sm/src/graph/transitions/counterproof.rs
```diff
@@ -190,4 +196,38 @@ impl GraphSM {
 
         Ok(duties)
     }
+
+    /// Decodes the per-byte operator signatures from input[0] of an on-chain Counterproof tx.
+    ///
+    /// The witness layout is `[sig_{N-1}, .., sig_0, n-of-n sig, leaf script, control block]`
+    /// — operator signatures pushed `.rev()`, then 3 trailing items. Reverse + skip(3)
+    /// recovers `[sig_0, .., sig_{N-1}]`.
```
Rajil1213: Nit: this is not really a documentation for this API. This should just say that it recovers signatures in byte-order of the counterproof. The implementation details should go into the function body (before the `reverse()` call).

crates/bridge-sm/src/graph/transitions/counterproof.rs
```diff
@@ -190,4 +196,38 @@ impl GraphSM {
...
+    fn decode_completed_sigs(
+        &self,
+        counterproof_tx: &Transaction,
+        event: &CounterProofConfirmedEvent,
+    ) -> GSMResult<CompletedSignatures> {
```
Rajil1213, line 205: Suggestion: This function should have its own unit test (perhaps even a proptest).

Rajil1213 (dismissed):
Some nits but no major blockers now. Nice work!

crates/bridge-exec/src/graph/counterproof_nack.rs
MdTeach: Added in [c30e308](https://github.com/alpenlabs/strata-bridge/pull/506/commits/c30e3082ae89b89f38f22d329382239da6837c26)

crates/bridge-sm/src/graph/transitions/counterproof.rs
MdTeach: Added in [c30e308](https://github.com/alpenlabs/strata-bridge/pull/506/commits/c30e3082ae89b89f38f22d329382239da6837c26)

crates/bridge-sm/src/graph/transitions/counterproof.rs
MdTeach: Will do as follow task: [STR-3316](https://alpenlabs.atlassian.net/browse/STR-3316)

Rajil1213 (approved):
Good to go now!

PR #508, opened 2026-04-29
refactor: remove P2A from tx-graph

uncomputable:
## Description

This PR removes the P2A CPFP connector from the transaction graph. The spec was already updated and this is the implementation.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

AI wrote some of the code, although I wrote most.

We decided to use Taproot outputs (public keys) for separate CPFP outputs, and to use descriptors for outputs (payout and slash) that hold funds and are reused as CPFP outputs.

## Checklist

## Related Issues

Closes https://alpenlabs.atlassian.net/browse/STR-3115


crates/connectors/src/keyed_anchor.rs
```diff
@@ -1,124 +1,108 @@
...
-/// CPFP connector that uses a custom locking script.
+use crate::{Connector, TaprootWitness};
+
+/// Keyed anchor.
 ///
-/// Because of the custom locking script,
-/// the input where this connector is spent requires a custom witness.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct CpfpConnector {
+/// This is a CPFP connector that requires a given key.
```
Rajil1213, line 11: You should also mention that this key will be used as an internal key in the taproot address. It's non-obvious how the key in the `KeyedAnchor` is used.

crates/tx-graph/src/game_graph.rs
```diff
@@ -104,7 +103,7 @@ pub struct KeyData {
     pub unstaking_image: sha256::Hash,
     /// For each watchtower, a fault key from Mosaic.
     pub wt_fault_pubkeys: Vec<XOnlyPublicKey>,
-    /// Operator descriptor that is used for CPFP and for receiving payouts.
+    /// Operator descriptor that is used for receiving payouts.
```
Rajil1213, line 106: It _is_ used for CPFP too in some cases (such as in the case of the payout transactions), right?

crates/bridge-sm/src/stake/context.rs
```diff
@@ -64,6 +64,12 @@ impl StakeSMCtx {
     ) -> SetupParams {
         SetupParams {
             operator_index: self.operator_idx(),
+            operator_pubkey: self
+                .operator_table()
+                .idx_to_btc_key(&self.operator_idx())
+                .expect("operator index must be valid")
+                .x_only_public_key()
+                .0,
```
Rajil1213: This is fine for now but the key used for the `KeyedAnchor` should actually be generated at runtime and included as part of the `DepositParams`. The reason is that the musig2 key is only used in the covenant and is unlikely to have funds and so cannot be directly used for CPFP. We will of course need to extra inputs to the child tx either way but the CPFP API will be neater if we don't use the musig2 key here.

crates/connectors/src/multi_anchor.rs
```diff
@@ -0,0 +1,158 @@
...
+pub struct MultiAnchorSpendPath {
+    /// Index of the spending watchtower.
+    pub watchtower_index: u32,
+}
+
+/// Witness data to spend a [`MultiAnchor`].
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct MultiAnchorWitness {
+    /// Index of the spending watchtower.
+    pub watchtower_index: u32,
```
Rajil1213: Use the `OperatorIdx` type both here and in the `*SpendPath`:

```suggestion
    pub watchtower_index: OperatorIdx,
```

crates/tx-graph/src/transactions/slash.rs
```diff
@@ -64,22 +62,36 @@ impl SlashTx {
...
-        debug_assert!(contest_slash_connector.network() == stake_connector.network());
-        let cpfp_connector = P2AConnector::new(
-            contest_slash_connector.network(),
-            contest_slash_connector.value(),
+        let excess_input = contest_slash_connector.value();
+        let first_extra = excess_input / n_watchtowers + excess_input % n_watchtowers;
+        let other_extra = excess_input / n_watchtowers;
+        debug_assert_eq!(
+            excess_input,
+            first_extra + other_extra * n_watchtowers.saturating_sub(1)
```
Rajil1213: Could you add a message here stating what you're trying to assert?

```rust
debug_assert_eq!(excess_input, first_extra + other_extra * n_watchtowers.saturating_sub(1), "<thing being asserted>");
```

crates/connectors/src/multi_anchor.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn value(&self) -> Amount {
+        self.value
+    }
+
+    fn to_leaf_index(&self, spend_path: Self::SpendPath) -> Option<usize> {
+        Some(spend_path.watchtower_index as usize) // cast safety: 32-bit arch or higher
+    }
+
+    fn get_taproot_witness(&self, witness: &Self::Witness) -> TaprootWitness {
+        let leaf_index = witness.watchtower_index as usize; // cast safety: 32-bit arch or higher
```
Rajil1213, line 67: I don't think this is accurate. You're assuming that the watchtowers are indexed separately local to any given graph but in reality, they are indexed globally. For example, if a graph belongs to operator at index `2`, then watchtowers will be indexed `0, 1, 3, ...`. There will be `N - 1` leaves in the taptree but only the first two leaves will correspond exactly to the operator indices while the rest will be off by one.

Rajil1213: There is a `watchtower` module in the `graph-sm` crate (under `graphs` module) which you might want to move to the `connectors` crate.

Rajil1213 (changes requested):
Given the extent of these changes, these are extremely well organized commits. Thanks!

There is only one real blocker for this PR that has to do with how watchtower indices (or operator indices) map to taproot indices.

crates/tx-graph/src/game_graph.rs
uncomputable: This question is about semantics. The operator descriptor is exclusively used for receiving payouts. This usage includes CPFP because the UTXO has two functions.

I removed "CPFP" because this referred to the CPFP output of the Claim transaction, which was changed to use the operator public key instead.

crates/tx-graph/src/transactions/slash.rs
uncomputable: Sure. This debug assert is always true. It mainly helps the developer see what is going on.

crates/connectors/src/keyed_anchor.rs
uncomputable: Isn't that an implementation detail that we shouldn't tell the user? Users should use the `finalize_input` method.

uncomputable:
I addressed all comments. I updated the commit that removes P2A to actually delete `p2a.rs`.

uncomputable:
The CI failures are unrelated with this PR.

Rajil1213:
> The CI failures are unrelated with this PR.

The security issues should be fixed via #519.

crates/connectors/src/keyed_anchor.rs
Rajil1213: Doesn't `finalize_input` only accept the `witness`? The caller needs to know how to create the witness/signature _before_ calling `finalize_input`, and how the taproot address is created determines how it will be signed.

crates/primitives/src/types.rs
```diff
@@ -15,6 +15,12 @@ use serde::{Deserialize, Serialize};
 /// The index of an operator.
 pub type OperatorIdx = u32;
 
+/// The index of a watchtower.
+///
+/// Watchtower indices correspond to the watchtower slots in each game graph.
+/// The index is relative to the operator index of the graph owner.
```
Rajil1213, line 21: An example might add more clarity here.

Rajil1213 (approved):
Approving this PR as the only changes needed now are some doc comments which are not strictly necessary for the purposes of this PR (feel free to address those at your own discretion).

crates/connectors/src/keyed_anchor.rs
uncomputable: Yes, the witness for the connector is required. However, the witness is documented in a self-contained way such that details about the tap tree don't need to be known. Also, `SigningInfo` abstracts away key tweaking, so the signer doesn't need to know if he's signing the output key or some untweaked key in a tap leaf.

Rajil1213:
@uncomputable can you rebase this PR to resolve the security issue? I'll re-approve once done.

uncomputable:
why do I need to rebase? I thought the merge queue applies my PR to `main` with all the latest updates

Rajil1213:
> why do I need to rebase? I thought the merge queue applies my PR to main with all the latest updates

I thought so too but apparently, it doesn't work 🤷

Rajil1213 (approved):

PR #509, opened 2026-04-30
fix(devcli): support new deposit descriptor

Rajil1213:
## Description

This PR updates the `bridge-in` handler in the `dev-cli` to use the `DepositDescriptor` as expected in the OL with multi-EE support. This is required only for completeness (and end-to-end testing of the strata stack) as both the ASM and the bridge continue to accept deposit requests with malformed deposit descriptors.

AI was not used for these changes.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,9 @@ strata-bridge-tx-graph = { path = "crates/tx-graph" }
 strata-mosaic-client = { path = "crates/mosaic-client" }
 strata-mosaic-client-api = { path = "crates/mosaic-client-api" }
 
+# Dependencies from alpenlabs/alpen pinned to commit 565fafcc (main @ 2026-04-30)
+strata-ol-bridge-types = { git = "https://github.com/alpenlabs/alpen.git", rev = "565fafccf2fe181d2f93003f9135ceed126a01ad" }
+
 # Dependencies from alpenlabs/asm pinned to commit a38db601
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -76,6 +76,9 @@ strata-bridge-tx-graph = { path = "crates/tx-graph" }
 strata-mosaic-client = { path = "crates/mosaic-client" }
 strata-mosaic-client-api = { path = "crates/mosaic-client-api" }
 
+# Dependencies from alpenlabs/alpen pinned to commit 565fafcc (main @ 2026-04-30)
+strata-ol-bridge-types = { git = "https://github.com/alpenlabs/alpen.git", rev = "565fafccf2fe181d2f93003f9135ceed126a01ad" }
+
```
MdTeach, line 81: Should we move this inside dev-cli ?
Since thats only where `alpenlabs/alpen` is used.

MdTeach (dismissed):

Cargo.toml
Rajil1213: I'd like to keep this here so that it's easier to check the versions of all the `alpenlabs/*` deps in one place, and if we ever need to add more, they'll all stay in sync.

MdTeach (approved):

PR #524, opened 2026-05-06
STR-1980: Bridge proof executor integration - Native Mode Proofs

MdTeach:
## Description
Implement the bridge proof input preparation and proof generation in GenerateAndPublishBridgeProof executor.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update


AlpenCodex:
@codex Please review this PR

(This is a test being run by @sampkaALP (security))

crates/bridge-sm/src/graph/duties.rs
```diff
@@ -202,6 +202,9 @@ pub enum GraphDuty {
         /// The index of the graph this duty is associated with.
         graph_idx: GraphIdx,
 
+        /// The index of the graph owner operator.
+        operator_index: OperatorIdx,
```
Rajil1213, line 206: If this is the graph owner's index, it should already be present in the `graph_idx`.

crates/bridge-sm/src/graph/duties.rs
```diff
@@ -205,6 +205,9 @@ pub enum GraphDuty {
         /// The index of the graph owner operator.
         operator_index: OperatorIdx,
 
+        /// The last Bitcoin block height seen by the graph state.
+        last_block_height: BitcoinBlockHeight,
```
Rajil1213, line 209: Probably worth mentioning why this is necessary since the executor can just query for the latest tip height.

crates/bridge-exec/src/graph/bridge_proof.rs
```diff
@@ -76,3 +93,128 @@ pub(super) async fn generate_and_publish_bridge_proof(
...
+    output_handles: &OutputHandles,
+    deposit_idx: DepositIdx,
+    operator_index: OperatorIdx,
+    last_block_height: BitcoinBlockHeight,
+) -> Result<Vec<u8>, ExecutorError> {
+    let proof_input = fetch_bridge_proof_input(
+        output_handles,
+        deposit_idx,
+        operator_index,
+        last_block_height,
```
Rajil1213, line 110: We might want to have a config that lets the operators decide how much PoW they want accumulated in their proof (in terms of number of blocks after the fulfillment confirmation). Then we can just wait till that many blocks are present before producing the proof (this can be done either in the executor or ideally as part of the state machine). Right now, we just use the height of whatever block the contest tx landed in (that means its 3xbury_depth (fulfillment + claim + contest, assuming operator does not wait for cooperative payout and even worse if a malicious watchtower does not even wait till the claim is buried before contesting).

crates/bridge-sm/src/graph/duties.rs
Rajil1213: Looks like this has been fixed.

Rajil1213 (changes requested):
Just some suggestions. I was also going to suggest updating the `dev-cli` to post a bogus proof instead of having to wait for the actual proof to be generated but looks like the functional test runtimes aren't too different. So it should be fine.

crates/bridge-sm/src/graph/duties.rs
MdTeach: For now naively going with contest block height.
Create [this](https://alpenlabs.atlassian.net/browse/STR-3382) ticket to make this configurable.

crates/bridge-exec/src/graph/bridge_proof.rs
MdTeach: For now naively going with contest block height.
Create [this](https://alpenlabs.atlassian.net/browse/STR-3382) ticket to make this configurable.

Rajil1213 (approved):
This looks good now.

PR #526, opened 2026-05-08
fix(gsm): only transition to Aborted if both stake and payout connector are spent

Rajil1213:
## Description

Adopts the two-fact terminal model for the Graph SM: abort only when both the stake outpoint and payout connector are gone (or, in single-path states, when that one is gone). Previously the graph aborted on connector-spend alone, even though slash was still possible. This allowed an operator skip punishment by spending the payout connector (using the unstaking burn preimage) themselves.

This involved the following changes:

- Typed `AbortReason` (`StakeSpent` / `PayoutConnectorSpent` / `Both`) replaces the stringly-typed reason.
- Per-state spend tracking via new `stake_spent` and `payout_connector_spent` fields.
- State-sensitive STFs for `process_stake_spent` and `process_payout_connector_spent`: record-and-stay in two-fact states, abort with `Both` on the second hit, direct-abort or reject in single-path states per spec.
- Proactive abort on transitions into `BridgeProofTimedout`/`Acked` (when stake spent) and `AllNackd` (when connector spent).
- Withhold `GraphAvailable` when the final partial arrives on a graph whose stake is already gone. So that we short-circuit deposit processing.
- Executor-side guard: `PublishGraphPartials` carries `stake_outpoint`; the executor refuses to publish if the outpoint is already spent.

Codex and Claude were used to plan these changes. Claude was used to write the code, and codex to review.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Please review the [spec](https://github.com/alpenlabs/bridge-sm-design-docs/pull/26) change before this one.

## Checklist

## Related Issues

Closes STR-3359


crates/bridge-sm/src/graph/transitions/payout.rs
```diff
@@ -118,30 +122,140 @@ impl GraphSM {
...
+                    payout_connector_spending_txid: spending_txid,
+                },
+            };
+            return Ok(GSMOutput::new());
+        }
+
+        // Two-fact post-`Claimed` state with neither field set yet: record
+        // the connector spend and stay. The GSM will react to a subsequent
+        // stake spend.
+        if self.state.set_payout_connector_spent(spending_txid) {
```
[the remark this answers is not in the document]
Rajil1213: While this is a good point, the nature of the presigned transactions that can legitimately spend the stake outpoint (`UnstakingTx` and `SlashTx`) precludes this possibility.

Rajil1213:
Converting this to draft as I re-evaluate how best to mitigate the attack.

Rajil1213:
Marking this as ready as the design-doc PR has been merged and alternative approaches for mitigating this attack aren't substantially easier to reason about.

crates/bridge-sm/src/graph/state.rs
```diff
@@ -80,6 +131,10 @@ pub enum GraphState {
 
         /// Partial signature from each operator.
         partial_signatures: BTreeMap<OperatorIdx, Vec<PartialSignature>>,
+
+        /// Set when the stake outpoint is consumed by something other than
+        /// this graph's slash.
+        stake_spent: Option<Txid>,
```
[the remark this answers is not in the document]
Rajil1213: Backward compatibility is not a requirement at the moment.

crates/bridge-sm/src/graph/transitions/contested.rs
```diff
@@ -496,70 +556,101 @@ impl GraphSM {
...
 
-                Ok(GSMOutput::new())
-            }
-            state @ GraphState::Slashed { .. } => Err(GSMError::duplicate(state, event.into())),
-            state => Err(GSMError::invalid_event(state, event.into(), None)),
+        // States that carry `stake_spent` but have not yet seen the connector
+        // spent: record the spend and stay. The GSM will react to subsequent
+        // events that might still complete the game.
+        if self.state.set_stake_spent(spend_txid) {
+            return Ok(GSMOutput::new());
```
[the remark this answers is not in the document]
Rajil1213: This PR already considers this possibility. The executor for deposit partials checks whether the stake outpoint has already been spent before proceeding. It is still possible that the stake is spent _after_ this check has been made at which point nothing else can be done but to follow the normal graph flow `Claimed` -> (`PayoutConnectorSpent`) -> `Aborted`.

crates/bridge-sm/src/graph/transitions/payout.rs
```diff
@@ -108,8 +108,8 @@ impl GraphSM {
         }
     }
 
-    /// Processes the event where the payout connector has been spent by an
-    /// unexpected transaction.
+    /// Processes the event where the payout connector has been spent by a
+    /// transaction other than the legitimate payout.
```
mukeshdroid: should it be "legitimate payouts"?

crates/bridge-sm/src/graph/transitions/contested.rs
```diff
@@ -426,6 +426,17 @@ impl GraphSM {
                 // (all watchtower slots), otherwise stay in CounterProofPosted.
                 let expected_nacks = graph_summary.counterproofs.len();
                 if counterproof_nacks.len() == expected_nacks {
+                    // The only path forward from `AllNackd` is the contested
```
mukeshdroid, line 433: This is correct as per specs but not per the current implementation where non-pov operators publish slash duty from the `AllNacked` state.

Crated a ticket to update the implementation. https://alpenlabs.atlassian.net/browse/STR-3565

crates/bridge-sm/src/graph/tx_classifier.rs
```diff
@@ -246,14 +247,10 @@ impl TxClassifier for GraphSM {
                     ))
                 } else if is_payout_connector_spent(&graph_summary.claim, tx) {
```
mukeshdroid, line 252: This state should also expect to get the contested payout tx (which is emitted from this state). as of now, it will misclassify it as a `PayoutConnectorSpent`.

mukeshdroid (changes requested):
Although, there was lots of changes it was a delight to review to the atomic structure of commits. Reviewing the specs earlier also helped a lot.

Only comment requiring change is to expect the contested payout in `CounterProofPosted` state.

crates/bridge-sm/src/graph/transitions/payout.rs
Rajil1213: > should it be "legitimate payouts"?

Amended in https://github.com/alpenlabs/strata-bridge/commit/e33b7c5b834dfa379ad447c0e640219dd69a5bb1.

crates/bridge-sm/src/graph/tx_classifier.rs
Rajil1213: > This state should also expect to get the contested payout tx (which is emitted from this state). as of now, it will misclassify it as a `PayoutConnectorSpent`.

This is an amazing find! Fixed in https://github.com/alpenlabs/strata-bridge/commit/da4cc58e3c2a517db65bc7082be6bc30292f4567.

mukeshdroid (approved):
The changes requested have landed.
Looks good to merge once CI is resolved.

PR #527, opened 2026-05-08
hack(tx-graph): hardcode fees into bridge txs

Rajil1213:
## Description

This PR hardcodes fees into each transaction broadcasted by the bridge node at a fixed fee rate of 2 sat/vb (subject to bikeshedding). This change is required until we implement CPFP. So for now, the approach used here optimizes for speed of implementation rather than elegance of the overall API.

As the shape of the transaction graph is dictated only by the number of watchtowers and the counterproof data, all the vsize's of the transactions have been precomputed (and asserted against the actual size via tests). Then the input amount of each transaction has been inflated to cover the fees. The same check has also been applied to non-presigned transactions namely:
- the deposit transaction so that any DRT with insufficient fees are dropped, and
- the cooperative payout transaction. 

Transactions funded by wallets already use the current fee rate but an additional change has been made so that the fee rate does not drop below the hardcoded fee rate value (of 2 sat/vb). This applies to the transactions that create UTXOs to fund claim and stake transactions.

All the tests have been updated so that that spawned bitcoin node does not accept 0-fee transactions anymore.

Claude was used for planning and making the changes (especially the size calculations). Codex was used to review the changes.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

To make it easier to revert these _code changes_, this PR should be squash-merged.
Note that reverting _this feature_ only involves setting the hardcoded fee rate to 0.
Setting the priority to P0 as this blocks our imminent deployment on signet.

## Checklist

## Related Issues

Closes STR-3379

### DIFF
--- a/bin/strata-bridge/src/mode/services/operator_wallet.rs
+++ b/bin/strata-bridge/src/mode/services/operator_wallet.rs
@@ -50,7 +50,8 @@ pub(in crate::mode) async fn init_operator_wallet(
     info!(%general_key, "operator wallet general key");
     let stakechain_key = s2_client.stakechain_wallet_signer().pubkey().await?;
     info!(%stakechain_key, "operator wallet stakechain key");
-    let operator_funds = compute_funding_amount(params);
+    let own_musig2_key = s2_client.musig2_signer().pubkey().await?;
+    let operator_funds = compute_funding_amount(params, own_musig2_key);
     let operator_wallet_config =
         OperatorWalletConfig::new(operator_funds, SEGWIT_MIN_AMOUNT, params.network);
     debug!(?operator_wallet_config, "operator wallet config");
@@ -74,7 +75,7 @@ pub(in crate::mode) async fn init_operator_wallet(
 ///
 /// This amount is not a constant since it depends upon the number of watchtowers that are allowed
 /// to contest a claim.
-fn compute_funding_amount(params: &Params) -> Amount {
+fn compute_funding_amount(params: &Params, own_musig2_key: XOnlyPublicKey) -> Amount {
     // Must match the value used in `orchestrator.rs::COUNTERPROOF_N_BYTES`. Hardcoded here too
     // because `Params` does not currently expose it.
     const COUNTERPROOF_N_BYTES: NonZero<usize> =
@@ -90,8 +91,15 @@ fn compute_funding_amount(params: &Params) -> Amount {
         sha256::Hash::from_slice(&[0u8; 32]).expect("must be a valid sha256 hash");
 
     // NOTE: (@Rajil1213)  musig2 keys are the watchtower keys for the time being until we separate
-    // the sets
-    let watchtower_keys: Vec<_> = params.keys.covenant.iter().map(|c| c.musig2).collect();
+    // the sets. Exclude the owner — graph construction in `bridge-sm` excludes the owner from
+    // watchtowers (see `GraphContext::watchtower_pubkeys`), so the funding amount must too.
+    let watchtower_keys: Vec<_> = params
+        .keys
+        .covenant
+        .iter()
+        .map(|c| c.musig2)
+        .filter(|k| *k != own_musig2_key)
+        .collect();
     // cast safety: covenant.len() is bounded by the number of operators, much smaller than u32::MAX
     let n_watchtowers = watchtower_keys.len() as u32;
     let contest_timelock = relative::Height::from_height(params.protocol.contest_timelock);
--- a/bin/strata-bridge/src/mode/services/operator_wallet.rs
+++ b/bin/strata-bridge/src/mode/services/operator_wallet.rs
@@ -76,10 +76,10 @@ pub(in crate::mode) async fn init_operator_wallet(
 /// This amount is not a constant since it depends upon the number of watchtowers that are allowed
 /// to contest a claim.
 fn compute_funding_amount(params: &Params, own_musig2_key: XOnlyPublicKey) -> Amount {
-    // Must match the value used in `orchestrator.rs::COUNTERPROOF_N_BYTES`. Hardcoded here too
+    // Must match the value used in `orchestrator.rs::COUNTERPROOF_N_DATA`. Hardcoded here too
     // because `Params` does not currently expose it.
-    const COUNTERPROOF_N_BYTES: NonZero<usize> =
-        NonZero::new(128 + 4).expect("counterproof_n_bytes must be non-zero");
+    const COUNTERPROOF_N_DATA: NonZero<usize> =
+        NonZero::new(128 + 4).expect("counterproof_n_data must be non-zero");
 
     let network = params.network;
 
@@ -109,7 +109,7 @@ fn compute_funding_amount(params: &Params, own_musig2_key: XOnlyPublicKey) -> Am
         n_of_n_key,
         watchtower_keys,
         contest_timelock,
-        fee::claim_contest_surcharge(n_watchtowers, COUNTERPROOF_N_BYTES),
+        fee::claim_contest_surcharge(n_watchtowers, COUNTERPROOF_N_DATA),
     );
 
     let admin_key = params.keys.admin;
--- a/bin/strata-bridge/src/mode/services/operator_wallet.rs
+++ b/bin/strata-bridge/src/mode/services/operator_wallet.rs
@@ -1,6 +1,6 @@
 //! Provides operator wallet initialization.
 
-use std::sync::Arc;
+use std::{num::NonZero, sync::Arc};
 
 use anyhow::anyhow;
 use bdk_bitcoind_rpc::bitcoincore_rpc;
@@ -13,12 +13,10 @@ use operator_wallet::{OperatorWallet, OperatorWalletConfig, sync::Backend};
 use secret_service_client::SecretServiceClient;
 use secret_service_proto::v2::traits::{SchnorrSigner, SecretService};
 use strata_bridge_common::params::Params;
-use strata_bridge_connectors::{
-    Connector,
-    prelude::{ClaimContestConnector, ClaimPayoutConnector},
-};
+use strata_bridge_connectors::prelude::{ClaimContestConnector, ClaimPayoutConnector};
 use strata_bridge_db::{fdb::client::FdbClient, traits::BridgeDb};
 use strata_bridge_primitives::constants::SEGWIT_MIN_AMOUNT;
+use strata_bridge_tx_graph::{fee, transactions::prelude::ClaimTx};
 use tracing::{debug, info};
 
 use crate::config::Config;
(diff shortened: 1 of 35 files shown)


uncomputable (changes requested):
NACK 6971e9d9a79f4902156cebe5351be8fc8d6131ab This seems like an invasive change. We need to discuss this more.

bin/strata-bridge/src/mode/services/operator_wallet.rs
```diff
@@ -77,6 +75,11 @@ pub(in crate::mode) async fn init_operator_wallet(
 /// This amount is not a constant since it depends upon the number of watchtowers that are allowed
 /// to contest a claim.
 fn compute_funding_amount(params: &Params) -> Amount {
+    // Must match the value used in `orchestrator.rs::COUNTERPROOF_N_BYTES`. Hardcoded here too
```
uncomputable: Nit: This was renamed to `COUNTERPROOF_N_DATA` recently

crates/tx-graph/src/fee.rs
```diff
@@ -0,0 +1,246 @@
+//! Hardcoded fees for presigned transactions.
```
uncomputable, line 1: I'm sure you are aware of this, but I will write it down here for the record: All of this file has to go in the final version. Hardcoding dust amounts and transaction sizes is brittle and can break with every small change.

uncomputable: I haven't checked these values. I will just assume they are correct.

uncomputable (changes requested):
6971e9d9a79f4902156cebe5351be8fc8d6131ab Please split the game graph unit tests in 2: One case where CPFP is used and extra fees are set to 0. And one case where CPFP is not used (no CPFP children should be created) and extra fees are set. This will give us a firmer understanding that the new fee structure works.

Currently the fees are set by in the end, CPFP is used. The tests effectively ignore the extra fees. We also don't know if CPFP still works if we set the extra fees to zero.

bin/strata-bridge/src/mode/services/operator_wallet.rs
Rajil1213: > Nit: This was renamed to `COUNTERPROOF_N_DATA` recently

Nice catch. I fixed the references in code while rebasing but didn't catch this comment. Will amend the commit.

crates/tx-graph/src/fee.rs
Rajil1213: > I'm sure you are aware of this, but I will write it down here for the record: All of this file has to go in the final version. Hardcoding dust amounts and transaction sizes is brittle and can break with every small change.

Yes. This is just a hack to add fees to the presigned transactions in a way that the fee rate remains consistent across all transactions. These values are pinned by unit tests. So if there is any deviation in the transaction sizes, those tests will fail. Of course, this is not ideal either (tests can change), but this should suffice for the lifetime of this feature.

Rajil1213:
> Currently the fees are set by in the end, CPFP is used. The tests effectively ignore the extra fees. We also don't know if CPFP still works if we set the extra fees to zero.

Good point. How do you feel about changing the `node.submit_package(&[parent, child])` API to submit just the parent if the parent already has fees? This feels like the smallest diff to me and would be easier to update once we have proper CPFP and we can get rid of intrinsic fees.

PS: We already know this works because the functional tests (without CPFP) are working.

crates/tx-graph/src/transactions/slash.rs
```diff
@@ -115,13 +115,19 @@ impl SlashTx {
             },
         ];
 
+        // cast safety: n_watchtowers fits in u32 (asserted via slice length being non-empty above
+        // and by upstream callers). The slash tx fee is taken from the first watchtower's stake
+        // share — that share is bounded by the per-watchtower stake amount, which is real funds
+        // and orders of magnitude larger than the fee.
+        let fee = crate::fee::slash_fee(watchtower_descriptors.len() as u32);
+        let first_output_value = watchtower_stake + first_extra - fee;
```
uncomputable: This forces the first watchtower to pay fees. Other watchtowers cannot pay fees unless they use CPFP. I will assume that this is the intended behavior for the hack.

uncomputable:
> How do you feel about changing the node.submit_package(&[parent, child]) API to submit just the parent if the parent already has fees? This feels like the smallest diff to me and would be easier to update once we have proper CPFP and we can get rid of intrinsic fees.

We can do that. I that case we don't assert that CPFP still works, but that's probably fine.

crates/tx-graph/src/transactions/slash.rs
Rajil1213: I can update it to divide the required fees among the watchtowers which I think is better even though there will be some slippage.

Rajil1213: > I can update it to divide the required fees among the watchtowers which I think is better even though there will be some slippage.

Done in https://github.com/alpenlabs/strata-bridge/commit/e280b7699e5390d0614998ff1bff5f9ffa91a2fd.

bin/strata-bridge/src/mode/services/operator_wallet.rs
Rajil1213: > Nit: This was renamed to `COUNTERPROOF_N_DATA` recently

Updated all references (this commit will probably not be reverted): https://github.com/alpenlabs/strata-bridge/commit/d49aa3001f3e1fda30a928ee68cce41bf4ba935e.

Rajil1213:
> We can do that. I that case we don't assert that CPFP still works, but that's probably fine.

Yes. We don't have CPFP anyway. Done in https://github.com/alpenlabs/strata-bridge/commit/5ecdeb8ab9002a2579824786aaeebb6b819c6fb9.

crates/connectors/src/test_utils.rs
```diff
@@ -307,13 +307,20 @@ impl BitcoinNode {
...
+    /// and only the parent is broadcast — this exercises the no-CPFP path. Otherwise the
+    /// pair is submitted as a CPFP package via `submitpackage`.
     ///
     /// # Panics
     ///
-    /// This method panics if the package is not accepted by the mempool.
+    /// This method panics if neither the parent on its own nor the package is accepted.
     pub fn submit_package(&self, transactions: &[Transaction; 2]) {
+        let [parent, _child] = transactions;
+        if self.client().send_raw_transaction(parent).is_ok() {
```
uncomputable: 5ecdeb8ab9002a2579824786aaeebb6b819c6fb9 I think this is over-engineering, but I won't block on this one. I would keep the method fully non-CPFP or fully CPFP, so we have a firmer understanding what the method is doing during testing.

crates/tx-graph/src/transactions/slash.rs
```diff
@@ -116,29 +116,31 @@ impl SlashTx {
...
         // cast safety: n_watchtowers fits in u32 (asserted via slice length being non-empty above
-        // and by upstream callers). The slash tx fee is taken from the first watchtower's stake
-        // share — that share is bounded by the per-watchtower stake amount, which is real funds
-        // and orders of magnitude larger than the fee.
+        // and by upstream callers). The slash tx fee is distributed across every watchtower
+        // output by rounding the per-watchtower share up — the tx may overpay by at most
+        // `n_watchtowers - 1` sats, which is acceptable slippage.
         let fee = crate::fee::slash_fee(watchtower_descriptors.len() as u32);
-        let first_output_value = watchtower_stake + first_extra - fee;
+        let fee_per_watchtower = Amount::from_sat(fee.to_sat().div_ceil(n_watchtowers));
```
uncomputable, line 123: e280b7699e5390d0614998ff1bff5f9ffa91a2fd The fee that is hardcoded for the slash transaction (`fees.rs`) needs to be multiplied by the number of watchtowers. Otherwise it's not large enough. Yes, this means that the txin must be larger than strictly necessary.

uncomputable: Also related: The spec for this is wrong, since the fee for the slash transaction should be `m_slash * w` where `m_slash` covers the slash transaction and `m_slash * w` is the extra input that is spread across all watchtowers.

uncomputable (changes requested):
d49aa3001f3e1fda30a928ee68cce41bf4ba935e Please scale the fee for the slash transaction by the number of watchtowers. If the tests are currently passing, then this is probably because of a low-fee environment.

crates/tx-graph/src/transactions/slash.rs
Rajil1213: The `vsize` of the slash transaction is calculated as a function of the number of watchtowers: https://github.com/alpenlabs/strata-bridge/blob/d49aa3001f3e1fda30a928ee68cce41bf4ba935e/crates/tx-graph/src/fee.rs#L332-L334

Let me know if that clears things up or if I misunderstood something.

Rajil1213: Interestingly enough, the spec already deducts fees evenly from each of the watchtower payouts: https://github.com/alpenlabs/txgraph/blob/b663360edf4eda3ff6b91a93db6ceebe090f0656/v1_tmp/slash.json#L21-L31


Here, `m_slash` isn't fee per watchtower but rather the cumulative fee for a given number of watchtowers.

crates/connectors/src/test_utils.rs
Rajil1213: > 5ecdeb8ab9002a2579824786aaeebb6b819c6fb9 I think this is over-engineering, but I won't block on this one. I would keep the method fully non-CPFP or fully CPFP, so we have a firmer understanding what the method is doing during testing.

Good point. I've amended that commit, see: https://github.com/alpenlabs/strata-bridge/commit/340e4aab8c363a3e943cc0ad0d2dc645b5e8efa1

crates/tx-graph/src/transactions/slash.rs
Rajil1213: Just to be sure. I also added the following lines to the `proof timeout + slash` scenario (not committed or pushed):

```rust
            let total_output_amount: Amount = slash.output.iter().map(|o| o.value).sum();
            let total_input_amount: Amount = slash
                .input
                .iter()
                .map(|i| i.previous_output)
                .map(|outpoint| {
                    let tx = node
                        .client()
                        .get_raw_transaction(outpoint.txid)
                        .unwrap()
                        .transaction()
                        .unwrap();
                    tx.output[outpoint.vout as usize].value
                })
                .sum();
            let slash_fee = total_input_amount - total_output_amount;
            let slash_fee_rate = slash_fee.to_sat() as f64 / slash.vsize() as f64;

            println!("Slash fee: {slash_fee}, fee rate: {slash_fee_rate:.2} sat/vB");
            assert_eq!(
                slash_fee,
                crate::fee::slash_fee(N_WATCHTOWERS as u32),
                "slash fee mismatched"
            );
```

The result when run is:
```text
running 1 test
Slash fee: 0.00001190 BTC, fee rate: 2.51 sat/vB
test game_graph::tests::proof_timeout_slash ... ok
```

uncomputable (approved):
ACK ff5231510a97bb46f9b5a6fd7b6c11d989d3ed54

PR #529, opened 2026-05-11
feat: handle availability of unstaking preimage

Rajil1213:
## Description

This PR handles the availability of the unstaking preimage when an operator posts the `UnstakingIntent` transaction. Once this transaction is posted, the preimage can be used to short-circuit any ongoing claims by burning the `claim-payout` connector. And so, this PR also implements the executor to post the `UnstakingBurn` transaction. The way the confirmation of `UnstakingIntent` transaction is handled is that:

* The preimage becomes available in the Stake SM.
* A new `post_stf_hook` has been added as part of the `StateMachine` trait which accepts a cross-state-machine context.
* This `cross-state-machine context` optionally holds the revealed preimage.
* In the implementation of this trait on the `GSM`, a `PublishUnstakingBurn` duty is emitted if the preimage is available in the `cross-state-machine context`.
* This duty gets appended to the duties queue from the normal STF.

A functional test group has been added (`unstaking`) that demonstrates the `UnstakingBurn` transaction being posted. This required adding an RPC to retrieve `StakeData` and a new command to the `dev-cli` to post the `UnstakingIntent` transaction.

Note that this PR depends on #526 (as `PublishUnstakingBurn` is unconditionally produced and so the abort path must be correctly handled).

Claude and Codex were used to plan this PR, and Codex to implement it.

### Type of Change

- [ ] Documentation update
- [x] Refactor (`preimage` derivatio is now part of the `key-deriv` crate and not the `secret-service`)
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This PR builds on #526 and should be reviewed and merged after that one.

## Checklist

## Related Issues

Closes STR-3114

### DIFF
--- a/crates/bridge-sm/src/graph/transitions/contested.rs
+++ b/crates/bridge-sm/src/graph/transitions/contested.rs
@@ -635,10 +635,7 @@ impl GraphSM {
         // abort.
         if let Some(payout_connector_spending_txid) = self.state.payout_connector_spent_txid() {
             self.state = GraphState::Aborted {
-                claim_txid: self
-                    .state
-                    .claim_txid()
-                    .expect("claim txid must be in state if aborting due to stake spend"),
+                claim_txid: self.state.claim_txid().unwrap_or(Txid::all_zeros()),
                 reason: AbortReason::Both {
                     stake_spending_txid: spend_txid,
                     payout_connector_spending_txid,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11316,7 +11316,10 @@ version = "0.1.0"
 dependencies = [
  "bitcoin",
  "ed25519-dalek",
+ "hkdf",
+ "make_buf",
  "secp256k1 0.29.1",
+ "sha2 0.10.9",
  "strata-bridge-primitives",
  "thiserror 2.0.18",
  "zeroize",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3791,6 +3791,7 @@ dependencies = [
  "strata-asm-proto-checkpoint-txs",
  "strata-asm-proto-checkpoint-types",
  "strata-bridge-common",
+ "strata-bridge-connectors",
  "strata-bridge-key-deriv",
  "strata-bridge-primitives",
  "strata-bridge-rpc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11270,6 +11270,7 @@ name = "strata-bridge-exec"
 version = "0.1.0"
 dependencies = [
  "algebra",
+ "bdk_bitcoind_rpc",
  "bdk_wallet",
  "bitcoin",
  "bitcoin-bosd 0.10.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,6 +183,7 @@ foundationdb = { git = "https://github.com/foundationdb-rs/foundationdb-rs", rev
 ] }
 futures = "0.3.32"
 hex = { version = "0.4.3", features = ["serde"] }
+hkdf = "0.12.4"
 indexmap = "2.13.0"
 jsonrpsee = "0.26.0"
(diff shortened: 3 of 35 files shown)


crates/bridge-sm/src/graph/tx_classifier.rs
```diff
@@ -106,17 +127,17 @@ impl TxClassifier for GraphSM {
                     }))
                 } else if is_payout_connector_spent(&graph_summary.claim, tx) {
                     Some(GraphEvent::PayoutConnectorSpent(
-                        PayoutConnectorSpentEvent {
-                            spending_txid: txid,
-                        },
+                        PayoutConnectorSpentEvent { tx: tx.clone() },
                     ))
+                } else if spends_stake_outpoint(&self.context.stake_outpoint(), tx) {
```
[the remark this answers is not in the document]
Rajil1213: This is not possible as the transactions are presigned and enforced by the covenant.

crates/bridge-sm/src/graph/transitions/contested.rs
```diff
@@ -635,10 +635,7 @@ impl GraphSM {
         // abort.
         if let Some(payout_connector_spending_txid) = self.state.payout_connector_spent_txid() {
             self.state = GraphState::Aborted {
-                claim_txid: self
-                    .state
-                    .claim_txid()
-                    .expect("claim txid must be in state if aborting due to stake spend"),
+                claim_txid: self.state.claim_txid().unwrap_or(Txid::all_zeros()),
```
mukeshdroid, line 638: Why is this change needed? Do we expect to ever reach this from a state where claim might not be available?

mukeshdroid (approved):
Looks good.

I tried to reason why we would not make the reveal of `Unstaking Presage` just be a signal. I guess in that route, we would need to add this value of all GSM states and update all STFs to handle this separately in retry. 

The only concern I had is earlier was that If the `Unstaking Burn` fails to go through, we are still advancing through the GSM. I think I have convinced myself that it is actually intended because the malicious operator would advance (by not following the canonical implementation) while others will be stuck. Is this correct?

Rajil1213:
> I tried to reason why we would not make the reveal of Unstaking Presage just be a signal. I guess in that route, we would need to add this value of all GSM states and update all STFs to handle this separately in retry.

The reasoning for this is included in the Orchestrator design doc but this is not the only reason. If we were to make this a signal, whenever an unstaking intent transaction is posted, it will update all GSMs associated with that operator. This means a whole bunch of state machines will end up getting updated. By our current persistence semantics, all these will need to be flushed to disk over a single transaction. There are hard limits with regards to the size of an FDB transaction (10 MB) and how long (5s) it can take to complete. Overall, not having to update a bunch of state machines in sync is better at the cost of introducing a new primitive in the orchestrator/SMs.

crates/bridge-sm/src/graph/transitions/contested.rs
Rajil1213: This is just so that we don't crash because of a bug in the `claim_txid()` implementation. I've explained this in the other PR which also makes this change in other places (somehow I forgot to update this one). This is admittedly a weird thing to do but if there is any drift in how the `claim_txid` is fetched from the state machine and how it's used here, we might crash even though the only reason the `claim_txid` even exists in the `Aborted` state is for monitoring needs and not functionally important.

Rajil1213:
> The only concern I had is earlier was that If the Unstaking Burn fails to go through, we are still advancing through the GSM. I think I have convinced myself that it is actually intended because the malicious operator would advance (by not following the canonical implementation) while others will be stuck. Is this correct?

Yes. The other PR already accounts for cases where the Unstaking Burn makes it on chain while the "normal" GSM operations are in flight. There are situations where we want to advance through the GSM (for example, if the stake has not been spent yet) and there are situations where we want to just Abort the GSM.

PR #532, opened 2026-05-12
fix: harden DRT parsing and allow absurd fee rate on DT

Rajil1213:
## Description

This PR hardens the DRT parsing logic in the bridge. Previously, we relied solely on the parsing logic (`parse_drt`) from the `asm` repo. This function only checks for structural soundness of a DRT and not its crptographic soundness. Now, we check both the shape of the transaction (number of outputs, magic bytes, etc.) _and_ whether the DRT output is locked to the right script pubkey. These checks still use the helpers from the `asm` repo. (STR-3429)

Another issue is that if a user is to send a valid DRT with a large amount (> `deposit_amount` + 0.034 BTC), deposit transaction cannot be created causing the bridge node to crash. This happens because the call to `Psbt::finalize` errors if the fee rate of the finalized transaction exceeds 25k sat/vb. Now, we just log a warning. (STR-3430)

Claude was used to implement most of the code changes in this PR.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

@uncomputable please review the last two commits (that touch the `tx-graph` crate).

## Checklist

## Related Issues

Closes STR-3429
Closes STR-3430

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11363,6 +11363,7 @@ dependencies = [
  "strata-bridge-test-utils",
  "strata-l1-txfmt",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11153,6 +11153,7 @@ dependencies = [
  "strata-bridge-sm",
  "strata-bridge-test-utils",
  "strata-bridge-tx-graph",
+ "strata-l1-txfmt",
  "strata-mosaic-client-api",
  "strata-p2p",
  "strata-predicate",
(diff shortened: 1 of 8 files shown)


MdTeach:
@codex review

prajwolrg (dismissed):
LGTM!

Rajil1213:
Made some minor refactors (to use `const`'s and improve logs) since the last review: https://github.com/alpenlabs/strata-bridge/compare/b3550980b8bf2eb24570a8fc9c6c936cd5da2071..86fce63a376838d09dca950b78bc33149d3b9079

Specifically, `DRT_OUTPUT_INDEX` (from `asm`) is used to get the actual DRT output, and now we only warn if there is a transaction that has the same SPS-50 envelope as the DRT but is not a valid DRT for the current operator set.

uncomputable (approved):
ACK 86fce63a376838d09dca950b78bc33149d3b9079 Only reviewed changes to `tx-graph`

PR #537, opened 2026-05-15
ci: add Trivy hard gate before ECR push for all bridge images

madan-oss:
## Description

Implements [STR-2637](https://alpenlabs.atlassian.net/browse/STR-2637): add Trivy container image scanning as a hard gate before any ECR push for strata-bridge images.

Trivy now scans every image locally against the runner's Docker daemon **before** any push to ECR. A build fails immediately on unfixed CRITICAL or HIGH CVEs, preventing vulnerable images from reaching either private or public ECR.

### What changed

**`.github/actions/trivy-scan/action.yml`**
- Restructured from fully advisory to gate-first: hard gate scan (JSON, CRITICAL/HIGH, `exit-code: 1`) runs first without `continue-on-error`. Failure propagates to the caller, causing all push steps to be skipped.
- Advisory SARIF scan (CRITICAL/HIGH/MEDIUM/LOW, all severities for Security tab visibility) runs after with `if: always()`.
- CycloneDX SBOM scan added as advisory artifact with `if: always()`.
- New `trivyignores` and `skip_dirs` inputs wired through all three scan steps.
- All post-gate steps (`if: always()`) so scan results and SARIF uploads happen even when the gate fails.

**`.github/workflows/docker-publish-ecr.yml`**
- **Dependency images** (`bridge-base`, `bridge-rt`): changed from `--push` (no local image) to `--load` into runner daemon → Trivy gate scan → `docker push`. Images previously reached ECR before any scan.
- **App images** (`strata-bridge`, `strata-asm-runner`, `secret-service`): already built locally before push — gate enforced by removing advisory `continue-on-error` from gate step.
- **Dual checkout** applied to all 4 jobs (`plan`, `build_dependencies`, `build_app_images`, `publish_report`): root checkout always from `main` for runner-executed scripts; build ref checked out into a subdirectory (`plan-src/` or `build-src/`) used only as Docker build context or for SHA resolution.
- `.trivyignore` resolved from root (main) checkout in all jobs — untrusted `inputs.ref` cannot suppress CVEs.
- `trivyignores: .trivyignore` wired into every action call.
- `push: branches: [main]` trigger added for automatic CD on merge to main (in addition to `workflow_dispatch`).

**`.trivyignore`** (new)
- 6 HIGH CVEs suppressed with review-by dates and cargo upgrade tracking notes. All are Cargo.lock dependencies (libp2p-gossipsub, rustls-webpki, yamux) pending upstream releases.

### Security model

`.trivyignore` and all runner-executed scripts are always checked out from `main` — only reviewed, merged code can affect the gate or suppress CVEs. The build ref (`inputs.ref` or `resolved_sha`) is used exclusively as a Docker build context; no code from it runs on the runner.

### Type of Change

## Notes to Reviewers

- Gate is CRITICAL/HIGH only. MEDIUM/LOW appear in SARIF (Security tab) but don't block.
- `.trivyignore` suppressions require a PR — format documented in the file header.
- `bridge-base` scan uses `skip_dirs: /root/.cargo,/root/.rustup` to exclude Cargo build cache false positives. App images built from `bridge-rt` do not need this.
- Auto-CD on push to main gates on the `AWS` environment's required reviewers (`Zk2u`, `sapinb`) before any deployment step runs.

## Checklist

## Related Issues

Closes STR-2637

[STR-2637]: https://alpenlabs.atlassian.net/browse/STR-2637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Rajil1213:
Right now, the trivy scan flags many positives in the base image but it does not flag anything for the actual runner images. Does this PR also account for this? Otherwise, most deployments will fail.

madan-oss:
> Right now, the trivy scan flags many positives in the base image but it does not flag anything for the actual runner images. Does this PR also account for this? Otherwise, most deployments will fail.

App images show clean because Trivy is blind to Rust CVEs without a Cargo.lock.

The final app images (strata-bridge, strata-asm-runner, secret-service) only contain the compiled binary — no Cargo.lock or source is copied in. Trivy detects Rust CVEs by scanning lock files, not compiled binaries, so it finds nothing and reports clean. That's a blind spot, not confirmation the images are safe.

The real signal is in bridge-base's app/Cargo.lock scan, which found 6 genuine HIGHs in dependencies that are almost certainly compiled into the shipped binaries:

libp2p-gossipsub 0.49.2 → fix: 0.49.4 (remote crash via gossip overflow)
yamux 0.12.1 / 0.13.5 → fix: 0.13.10 (stream multiplexer vuln)
rustls-webpki 0.101.7 → fix: 0.103.13 (DoS via malformed CRL)

Rajil1213:
> The final app images (strata-bridge, strata-asm-runner, secret-service) only contain the compiled binary — no Cargo.lock or source is copied in. Trivy detects Rust CVEs by scanning lock files, not compiled binaries, so it finds nothing and reports clean. That's a blind spot, not confirmation the images are safe.

I see. Currently, we use the output of `cargo-audit` to discover and address vulnerabilities. Looks like, in this case, trivy is more thorough. Given that you want to make this a hard gate, could you add a separate trivy workflow that runs on every PR so that we can identify issues earlier rather than being blocked at deployment time?

madan-oss:
Good call. A PR-time filesystem scan on `Cargo.lock` (no image build needed) would catch these early without the cost of building all 5 images on every PR. Tracking that as a follow-up — out of scope for STR-2637 which is specifically the deployment gate, but worth a separate story.

Rajil1213:
This needs a rebase. I also want to fix the issues flagged here so that we don't suppress legitimate fixable alerts.

Rajil1213:
I've created a new PR (#556) that fixes the HIGH vulnerabilities reported by trivy (and ignores those that are not fixable at the moment via a `.trivyignore.yaml` file). PTAL.

.github/actions/trivy-scan/action.yml
```diff
@@ -104,22 +114,45 @@ runs:
         timeout: ${{ inputs.trivy_timeout }}
         scanners: ${{ inputs.trivy_scanners }}
         docker-host: unix:///var/run/docker.sock
+        severity: CRITICAL,HIGH,MEDIUM,LOW
```
Rajil1213: This should also just be CRITICAL, HIGH.

.github/scripts/resolve_build_plan.sh
```diff
@@ -28,9 +28,9 @@ if [[ -z "${checkout_ref}" ]]; then
   checkout_ref="${GITHUB_SHA}"
 fi
 
-resolved_sha="$(git rev-parse HEAD)"
+resolved_sha="$(git -C plan-src rev-parse HEAD)"
```
Rajil1213, line 34: What is this doing?

.trivyignore.yaml
Rajil1213, line 1: This can be removed after #556 is merged.

.github/actions/trivy-scan/action.yml
```diff
@@ -38,13 +38,21 @@ inputs:
   artifact_name:
     description: "Name of the GitHub Actions artifact to upload scan results under."
     required: true
+  trivyignores:
+    description: "Path to .trivyignore.yaml file for CVE suppressions. Pass empty string to use no ignore file."
+    required: false
+    default: ""
```
Rajil1213: Default should be .trivyignore.yaml.

.github/scripts/resolve_build_plan.sh
Rajil1213: I see that it's switching to the plan-src directory where the current branch is checked out. Could you explain that in the comment? Since plan-src is created and populated in a different file (the actual workflow), this is brittle but also fine for now.

.github/CODEOWNERS
```diff
@@ -79,6 +79,12 @@
 # CI/CD
 /.github/ @Rajil1213 @storopoli
 
+# Security and publish workflows
+/.github/workflows/security.yml @madan-oss @Rajil1213
```
Rajil1213, line 83: Also  .github/scripts

.github/workflows/docker-publish-ecr.yml
```diff
@@ -1,6 +1,9 @@
 name: Docker Publish to ECR
 
 on:
+  push:
+    branches:
+      - main
```
Rajil1213: This feels excessive (right now, we have about 2 merges to main everyday). We should gate this on tags being pushed to main instead.

Rajil1213 (changes requested):
Did a second pass and it needs a few more changes that I didn't notice the last time.

.github/actions/trivy-scan/action.yml
madan-oss: Fixed — updated to `CRITICAL,HIGH` for both the SARIF scan and SBOM steps.

.github/scripts/resolve_build_plan.sh
madan-oss: Added a comment explaining the scope. `plan-src/` is the build-ref checkout created by the workflow's second checkout step; `-C plan-src` keeps SHA derivation scoped to the ref being built rather than the workflow scripts checkout (always pinned to main) that sits at the workspace root.

.trivyignore.yaml
madan-oss: #556 is merged. The file now carries the YAML-format suppressions from that PR (picked up via rebase) — no stale content remains.

.github/actions/trivy-scan/action.yml
madan-oss: Updated — default is now `.trivyignore.yaml`.

.github/CODEOWNERS
madan-oss: Added `/.github/scripts/ @madan-oss @Rajil1213` to the CODEOWNERS section.

.github/workflows/docker-publish-ecr.yml
madan-oss: Changed to `push: tags: ['v*']`. Automatic CD will now only trigger on version tags, not every merge to main.

.github/workflows/docker-publish-ecr.yml
```diff
@@ -168,18 +202,25 @@ jobs:
...
           TRIVY_STATUS: ${{ steps.trivy.outputs.status }}
         run: |
-          image_ref="${ECR_REGISTRY}/${ECR_REPOSITORY_PREFIX}/${{ matrix.repository }}:${IMAGE_TAG}"
           {
-            echo "- \`${image_ref}\`"
-            echo "  scan: advisory status \`${TRIVY_STATUS:-unknown}\`"
-            echo "  artifact: \`trivy-${{ matrix.component }}-${IMAGE_TAG}\`"
+            echo "- \`${IMAGE_REF}\`"
+            echo "  scan: \`${TRIVY_STATUS:-unknown}\`"
+            echo "  artifact: \`trivy-${{ matrix.component }}-${{ needs.plan.outputs.image_tag }}\`"
```
Rajil1213: This is being flagged by zizmor. Please check.

Rajil1213 (changes requested):
My previous comments have all been addressed but zizmor check is failing.

Rajil1213 (approved):

PR #552, opened 2026-05-18
fix(orchestrator): skip no-op db writes on nag and retry ticks

Rajil1213:
## Description

In our current design, the orchestrator feeds events to the state machines, collects the IDs of all state machines where their STF returned an `Ok(())` and then persists these state machines in the database. Each state machine maintains a property where they return an `Ok(())` only if the state mutates. However, this is not true for non-transitions/handlers (namely, the nag and retry handlers) which _always_ produce an `Ok(())` but do not mutate the state. This leads to unnecessary writes to the database and as nag and retry intervals are regular, this can degrade the node/db performance over prolonged runs. The right way to fix this is to add another top-level handler that deals with nags and retries, separate from the `process_event` STF. However, this introduces a substantial code churn. So for now, a separate [ticket](https://alpenlabs.atlassian.net/browse/STR-3493) has been created to handle this.

Instead, this PR introduces a new field `state_mutation` to `SMOutput` that indicates whether the state has changed. By default, this is true so at worst, we end up making more db calls than strictly necessary but never miss out on any changes. Outputs from retry and nag handlers are explicitly marked as `Unchanged` via `.mark_unchanged()`.

Claude was used to implement the changes. Codex was used to review.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

Closes STR-3488

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -202,8 +202,8 @@ libp2p = { version = "0.56.0", features = [
   "identify",
 ] }
 libp2p-identity = { version = "0.2.12", features = ["ed25519"] }
-miniscript = "12.3.0"
 metrics = "0.24"
+miniscript = "12.3.0"
 # NOTE: (@Rajil1213) Using an alpenlabs fork that bumps serdect 0.2 → 0.3 to fix a
 # serialize/deserialize mismatch in non-human-readable binary formats (e.g.
 # postcard). Upstream musig2 ≥0.3.0 ships serdect 0.3 natively, but requires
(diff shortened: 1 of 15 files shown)


mukeshdroid (approved):
The changes look good.
Ack [e8f9e42](https://github.com/alpenlabs/strata-bridge/pull/552/commits/e8f9e429e1cca68e1b7663edc9c1cd93ccd1e9c1).

PR #554, opened 2026-05-19
Counterproof: Authenticate bridge proof txn 

MdTeach:
## Description
* In the counterproof statement add bridge proof transaction authentication
* Add unit tests for the counterproof statement

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update


MdTeach:
@codex review

crates/proofs/bridge-counterproof/src/statements.rs
```diff
@@ -1,20 +1,27 @@
...
-    genesis::{BridgeCounterproofGenesis, load_genesis},
+    genesis::BridgeCounterproofGenesis,
     types::{CounterproofInput, CounterproofOutput},
 };
 
 /// Native entry point: loads genesis and runs the counterproof.
 #[cfg(not(target_os = "zkvm"))]
 pub fn process_counterproof(zkvm: &impl ZkVmEnv) {
-    let genesis = load_genesis();
+    let genesis = crate::genesis::load_genesis();
```
Rajil1213: Why the absolute path?

crates/proofs/bridge-counterproof/src/statements.rs
```diff
@@ -65,17 +81,70 @@ fn process_counterproof_inner(zkvm: &impl ZkVmEnv, genesis: &BridgeCounterproofG
...
+    let (internal_key, _parity) = operator_pubkey
+        .to_xonly_public_key()
+        .add_tweak(SECP256K1, &game_tweak)
+        .expect("tweak addition only fails for hostile operator keys");
+
+    // Single-leaf tap tree, so the merkle root is just the leaf hash.
+    let leaf_script = contest_proof_timelocked_leaf(n_of_n_pubkey, proof_timelock);
+    let merkle_root: TapNodeHash =
+        TapLeafHash::from_script(&leaf_script, LeafVersion::TapScript).into();
+    let (output_key, _parity) = internal_key.tap_tweak(SECP256K1, Some(merkle_root));
```
Rajil1213: Assuming this logic will be eventually refactored so that we only do this via the connectors.

crates/proofs/bridge-counterproof/src/statements.rs
```diff
@@ -65,17 +81,70 @@ fn process_counterproof_inner(zkvm: &impl ZkVmEnv, genesis: &BridgeCounterproofG
...
+
+    // Without this, an attacker could swap in a signature the operator made for
+    // some unrelated taproot output.
+    assert_eq!(
+        prevouts[0].script_pubkey,
+        ScriptBuf::new_p2tr_tweaked(output_key),
+        "prevouts[0] is not the ContestProofConnector output for this operator + game_idx",
+    );
```
Rajil1213: Can you explain this a bit more? The `msg` that we verify against already encodes the script pubkey in it. So there should not be any way for an operator to "fake" a signature corresponding to the tweaked operator pubkey.

crates/proofs/bridge-counterproof/src/statements.rs
```diff
@@ -65,17 +81,70 @@ fn process_counterproof_inner(zkvm: &impl ZkVmEnv, genesis: &BridgeCounterproofG
...
+fn contest_proof_timelocked_leaf(
+    n_of_n_pubkey: &BitcoinXOnlyPublicKey,
+    proof_timelock: bitcoin::relative::Height,
+) -> ScriptBuf {
+    script::Builder::new()
+        .push_slice(n_of_n_pubkey.to_xonly_public_key().serialize())
+        .push_opcode(opcodes::all::OP_CHECKSIGVERIFY)
+        .push_sequence(Sequence::from_height(proof_timelock.value()))
+        .push_opcode(opcodes::all::OP_CSV)
+        .into_script()
```
Rajil1213: This too should only be constructed via the `connectors` crate.

crates/connectors/src/timelocked.rs
```diff
@@ -200,6 +200,11 @@ impl ContestProofConnector {
         Scalar::from_be_bytes(tweak_bytes)
             .expect("the game index must be less than the secp curve order")
     }
+
+    /// Returns the P2TR output key.
+    pub fn output_key(&self) -> bitcoin::key::TweakedPublicKey {
```
Rajil1213, line 205: Nit: Import this at the top and replace with:

```suggestion
    pub fn output_key(&self) -> TweakedPublicKey {
```

Rajil1213: Looks like this is already done.

crates/proofs/bridge-counterproof/src/statements.rs
```diff
@@ -65,17 +84,48 @@ fn process_counterproof_inner(zkvm: &impl ZkVmEnv, genesis: &BridgeCounterproofG
...
+/// the `ContestProofConnector` output key and verifying against it.
+fn verify_operator_signature(
+    tx: &Transaction,
+    prevouts: &[TxOut],
+    operator_pubkey: &BitcoinXOnlyPublicKey,
+    game_idx: NonZero<u32>,
+    n_of_n_pubkey: &BitcoinXOnlyPublicKey,
+    proof_timelock: relative::Height,
 ) {
+    let wit_elem = tx.input[0]
```
Rajil1213: Flagged by local codex run:

There is no guarantee that the proof transaction will have the operator signature on the first input -- it is just a transaction that spends a fixed output of the claim transaction. So we need to either iterate through all the inputs to find the right prevout (via the script pubkey) or supply the input index and check whether it matches the expected script pubkey.

Rajil1213 (changes requested):
There is just one blocker. The rest are nits/questions.

crates/connectors/src/timelocked.rs
```diff
@@ -201,12 +196,35 @@ impl ContestProofConnector {
...
+        let leaf = timelocked_leaf_script(n_of_n_pubkey, proof_timelock);
+        let merkle_root: TapNodeHash =
+            TapLeafHash::from_script(&leaf, LeafVersion::TapScript).into();
+        internal_key.tap_tweak(SECP256K1, Some(merkle_root)).0
     }
 }
 
+/// Builds the timelocked tap-leaf script shared by every timelocked connector
+/// variant.
+fn timelocked_leaf_script(timelocked_key: XOnlyPublicKey, timelock: relative::Height) -> ScriptBuf {
```
uncomputable: This breaks encapsulation of the connectors. We should never expose the `ScriptBuf` unless we absolutely have to. I will continue reviewing to see how this is used.

uncomputable: Also, "the timelocked connectors" is a category that doesn't really exist in the code. It's an implementation detail because these connectors have been derived from the same macro, and because they live in the same file, but there is no semantic connection via a trait or something like that.

crates/proofs/bridge-counterproof/src/statements.rs
uncomputable: True but also false: The bridge proof transaction spec mandates that the first txin must spend the contest-proof connector. Otherwise the bridge proof transaction is malformed, which leads to a valid counterproof.

crates/connectors/src/timelocked.rs
```diff
@@ -201,12 +196,35 @@ impl ContestProofConnector {
             .expect("the game index must be less than the secp curve order")
     }
 
-    /// Returns the P2TR output key.
-    pub fn output_key(&self) -> bitcoin::key::TweakedPublicKey {
-        self.spend_info().output_key()
+    /// Computes the P2TR output key without constructing a connector.
+    pub fn output_key(
```
uncomputable: Why is this change necessary? Just create the connector and call `.output_key()` or `.spend_info().output_key()`. Yes, you have to pass a dummy surcharge, but that's fine. We will remove the surcharge in a future commit.

crates/connectors/src/timelocked.rs
uncomputable: Update: I can see that this function is private, so that makes it better. However, this function is not necessary if we remove the public function `.output_key()`. See my other comment on that.

uncomputable (changes requested):
6fabccd0a183936baea5c52ec0c9913753648345 I think the `output_key` function is not necessary.

crates/proofs/bridge-counterproof/src/statements.rs
Rajil1213: Is there a protocol-level way to enforce this?

The counterprover has to provide the bridge proof transaction as an input to the counterproof. If that transaction is not a valid bridge proof, then the counterproof fails. So if an operator spends the contest-proof connector in a transaction using it as a non-first input, we'll need another mode in the counterproof that proves that the operator violated the "soft" covenant.

Rajil1213: Adding a new mode feels completely avoidable to me. As far as the transaction graph is concerned, there is no covenant enforcing the nature of the Bridge Proof transaction. It's just a UTXO that has to be spent by an operator lest a watchtower post a Bridge Proof Timeout transaction.

uncomputable: Update: I can see there is a soundness issue with the current counterproof spec.

As Rajil said, the operator can spend the contest-proof connector in txin[1]. This is perfectly fine by Bitcoin consensus rules. If that happens, then watchtowers are not able to create a counterproof, because they don't have a "valid" bridge proof transaction as input.

As Rajil said, the counterproof must iterate over all txin's and look for the contest-proof connector + operator signature. This validates the bridge proof transaction.

uncomputable: @Rajil1213 are we in agreement?

Rajil1213: Didn't see your earlier message (so deleted my other reply).

Yes, we're in agreement.

uncomputable: I updated the counterproof statement spec to fix the error.

crates/proofs/bridge-counterproof/src/statements.rs
MdTeach: Refactored in [d407320](https://github.com/alpenlabs/strata-bridge/pull/554/commits/d407320e0582997ed10558f9c0690c5308765323)

crates/proofs/bridge-counterproof/src/statements.rs
MdTeach: Amended in [0d6e81a](https://github.com/alpenlabs/strata-bridge/pull/554/commits/0d6e81a00162e09aed141079b83fe93ae751204f)
Made the txin index as private input

crates/connectors/src/timelocked.rs
MdTeach: Continuing the discussion on the other comment

crates/connectors/src/timelocked.rs
MdTeach: Amended in [3650d12](https://github.com/alpenlabs/strata-bridge/pull/554/commits/3650d122fb28376133a6b9ecf5a045008a6c924a).
I have reverted by changes on connector crate and used connector with dummy surcharge and network.

MdTeach: Using dummy surcharge and network works for now, but `ContestProofConnector` will need a refactor — threading them through `new()` just to call `.output_key()` adds proving cycles that the output-key derivation doesn't need.

crates/proofs/bridge-counterproof/src/statements.rs
MdTeach: My bad, the schnorr verify against `output_key` already binds the signature to whatever `prevouts[txin_idx].script_pubkey` is in msg, so the assertion is redundant; removed it in [6637baa](https://github.com/alpenlabs/strata-bridge/pull/554/commits/6637baa6d4e6ec47181aac7a4306308d3d6cc166)

crates/connectors/src/timelocked.rs
Rajil1213: Agree with Abishek here. There is value in keeping the proof as lean as possible. Since, this has been reverted, could you create a ticket for this instead? We can take it up in a cleanup sprint (before mainnet).

crates/proofs/bridge-counterproof/src/statements.rs
Rajil1213: This will also require updating the counterproof spec because it still iterates over inputs in the bridge proof tx.

crates/connectors/src/timelocked.rs
MdTeach: Created the [ticket](https://alpenlabs.atlassian.net/browse/STR-3599)

crates/proofs/bridge-counterproof/src/statements.rs
MdTeach: It was updated there first;

uncomputable: Yeah, the counterproof spec is already updated.

uncomputable (dismissed):
ACK 6637baa6d4e6ec47181aac7a4306308d3d6cc166 Everything looks good. Let's optimize the proof generation in a separate PR.

Rajil1213 (dismissed):

Rajil1213:
@MdTeach for some reason ci is constantly unable to fetch flexitest. Can you try updating it to the latest main?

Rajil1213 (approved):

PR #555, opened 2026-05-19
ci: add Trivy Cargo.lock hard gate to security workflow (STR-3526)

madan-oss:
## Description

Implements [STR-3526](https://alpenlabs.atlassian.net/browse/STR-3526): adds a PR-time Trivy filesystem scan on \`Cargo.lock\` so CVEs are caught during code review rather than only at deployment.

### Context

\`cargo audit\` (already in \`security.yml\`) checks against the RustSec advisory DB (RUSTSEC-\* identifiers). Trivy \`fs\` is complementary — it covers CVE and GHSA identifiers from the OSV/NVD databases that don't always have a RUSTSEC counterpart. The 6 HIGH CVEs suppressed in STR-2637 (libp2p-gossipsub, yamux, rustls-webpki) are CVE/GHSA-format advisories and would not be caught by \`cargo audit\` alone.

### What changed

**\`.github/workflows/security.yml\`**
- New \`trivy-cargo\` job added alongside the existing \`supply-chain\` (cargo audit) job.
- Runs \`trivy fs .\` on every PR, merge group, push to main, and \`workflow_dispatch\`.
- **Hard gate**: CRITICAL/HIGH unfixed CVEs fail the check (\`exit-code: 1\`). Suppressions via \`.trivyignore\` require a PR.
- **Advisory SARIF**: all severities uploaded to Security tab with \`if: always()\` so findings are visible even when the gate fails.
- Guard creates an empty \`.trivyignore\` for refs that predate the file.

**\`.trivyignore\`** (new)
- Suppresses the 6 known Rust HIGHs (libp2p-gossipsub, rustls-webpki, yamux) with review-by dates — same set as STR-2637.

### Validation

Test run found the gate catches a real finding beyond cargo audit: \`urllib3 2.5.0\` in \`functional-tests/uv.lock\` has 3 HIGH CVEs (CVE-2025-66418, CVE-2025-66471, CVE-2026-21441, fixed in 2.6.3). This is tracked as dev work to update the uv.lock separately — the gate correctly blocks until fixed.

### Type of Change

## Notes to Reviewers

- \`cargo audit\` and Trivy cover different advisory DBs — both jobs should stay.
- The gate will fail on \`urllib3\` until \`functional-tests/uv.lock\` is updated to urllib3 ≥ 2.6.3. That is intentional — the gate is doing its job.
- This PR and STR-2637 (PR #537) both introduce \`.trivyignore\` at the repo root with the same Rust CVE suppressions. Whichever merges second will need a trivial conflict resolution.

## Checklist

## Related Issues

Closes STR-3526

[STR-3526]: https://alpenlabs.atlassian.net/browse/STR-3526


.github/workflows/security.yml
```diff
@@ -26,3 +26,72 @@ jobs:
...
+      # Cargo.lock in the repo. Complements cargo audit (RustSec DB) with CVE/GHSA coverage.
+      # Add justified exceptions to .trivyignore via PR.
+      - name: Scan Cargo.lock with Trivy (hard gate)
+        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
+        with:
+          scan-type: fs
+          scan-ref: .
+          scanners: vuln
+          severity: CRITICAL,HIGH
+          ignore-unfixed: "true"
```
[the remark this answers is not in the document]
Rajil1213: This is fine. If there is no upstream fix, it will be ignored anyways.

Rajil1213 (changes requested):
zizmor is flagging an unchanged part of the security workflow which needs fixing before this can be merged. Looks like trivy is also scanning `uv.lock` -- this is unnecessary as those only apply to functional tests where security is not a concern.

.github/workflows/security.yml
```diff
@@ -26,3 +27,74 @@ jobs:
...
+      # in the Security tab for triage.
+      - name: Scan Cargo.lock with Trivy (SARIF — Security tab)
+        if: always()
+        continue-on-error: true
+        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
+        with:
+          scan-type: fs
+          scan-ref: .
+          scanners: vuln
+          severity: CRITICAL,HIGH,MEDIUM,LOW
```
Rajil1213: This should also just be CRITICAL, HIGH. Why upload the full report when the lower severity ones will be ignored anyway?

.github/workflows/security.yml
```diff
@@ -26,3 +27,74 @@ jobs:
...
+          exit-code: "0"
+          format: sarif
+          output: trivy-cargo.sarif
+          trivyignores: .trivyignore.yaml
+          skip-dirs: functional-tests
+          hide-progress: "true"
+          skip-setup-trivy: "true"
+
+      - name: Upload SARIF to GitHub Security tab
+        if: always()
```
Rajil1213: Shouldn't this step depend upon the success of the previous step?

.trivyignore.yaml
Rajil1213, line 1: This file can be removed after #556 is merged.

Rajil1213 (changes requested):
This requires some minor changes.

.github/workflows/security.yml
madan-oss: Fixed — SARIF scan severity changed to `CRITICAL,HIGH` to match the hard gate.

.github/workflows/security.yml
madan-oss: Fixed — added `id: trivy_sarif_cargo` to the SARIF scan step and changed the upload condition to `if: steps.trivy_sarif_cargo.outcome == 'success'`.

.trivyignore.yaml
madan-oss: #556 is merged. The file now carries the YAML-format suppressions from that PR (picked up via rebase) — no stale content remains.

Rajil1213 (approved):
Looks good to go now.

PR #557, opened 2026-05-19
feat: reconnect logic in secret-service-client

Zk2u:
## Summary
Adds transparent reconnect to the secret-service QUIC client. Previously a single connection drop forced a bridge-node restart to recover.

- New `ConnHandle` wraps `RwLock<Connection>` + reusable `Endpoint`/`ClientConfig`/dial params.
- `ConnHandle::make_v2_req` is the single hot path: snapshot conn → request → on dead-connection errors, redial under write lock + retry once. Race-safe via `Connection::stable_id()`.
- Per-signer clients (`GeneralWalletClient`, `StakechainWalletClient`, `Musig2Client`, `P2PClient`, `StakeChainPreimgClient`) all hold a `ConnHandle` clone — reconnects are transparent to callers.
- `tracing::info!`/`warn!` on reconnect path.
- Application errors (`Timeout`, `TryAgain` exhaustion, serialization) propagate without redialing.

## What's not in this PR
- Backoff/cooldown on consecutive reconnect failures — under high RPS against a crash-looping server, callers will each pay one redial. Worth a follow-up; not urgent for current bridge load.
- Integration test that kills a running server mid-test. Existing tests cover the happy path end-to-end (\`bin/secret-service/src/tests.rs\`); a kill-and-recover test would need server-stop plumbing. Unit test for \`should_reconnect\` classifier included.

## Test plan
- [x] \`just lint\` passes
- [x] \`cargo check --workspace --tests --all-features\` passes
- [x] Unit test \`should_reconnect_classifies_errors\`
- [ ] Functional test against a real bridge with server restart (manual)

Closes STR-3305.

### DIFF
--- a/crates/secret-service-client/src/lib.rs
+++ b/crates/secret-service-client/src/lib.rs
@@ -21,7 +21,8 @@ use preimage::PreimgClient;
 pub use quinn::rustls;
 use quinn::{
     crypto::rustls::{NoInitialCipherSuite, QuicClientConfig},
-    ClientConfig, ConnectError, Connection, ConnectionError, Endpoint, TransportConfig,
+    ClientConfig, ConnectError, Connection, ConnectionError, Endpoint, ReadExactError,
+    TransportConfig, WriteError,
 };
 use rkyv::{deserialize, rancor, util::AlignedVec};
 use secret_service_proto::{
@@ -35,10 +36,15 @@ use secret_service_proto::{
     },
 };
 use terrors::OneOf;
-use tokio::time::timeout;
+use tokio::{sync::RwLock, time::timeout};
+use tracing::{info, warn};
 use wallet::{GeneralWalletClient, ReservedWalletClient};
 
 const KEEP_ALIVE_INTERVAL: Duration = Duration::from_secs(25);
+
+/// Maximum number of times a request is retried inside [`v2_req`] when the server responds with
+/// [`ServerMessage::TryAgain`].
+const MAX_RETRIES: usize = 10;
 
 /// Configuration for the Secret Service client.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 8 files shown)


Zk2u:
Addressed the timeout bound: `connecting.await` is now wrapped in `tokio::time::timeout(self.inner.timeout, ...)` so a hung handshake returns `ClientError::Timeout` instead of blocking every other caller behind the write lock. `warn!` on the timeout path.

crates/secret-service-client/src/lib.rs
```diff
@@ -63,14 +65,131 @@ pub struct Config {
...
+            // construction time, so a failure here means the system is in a degenerate state.
+            // Surface it as a generic reset.
+            .map_err(|e| {
+                warn!(?e, "secret-service reconnect dial failed");
+                ClientError::ConnectionError(ConnectionError::Reset)
+            })?;
+        let new_conn = connecting.await.map_err(|e| {
+            warn!(?e, "secret-service reconnect handshake failed");
+            ClientError::ConnectionError(e)
+        })?;
```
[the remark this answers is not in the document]
Rajil1213: The part about missing timeout is valid, I think. The part about other connections being blocked is acceptable since the connection is already stale.

crates/secret-service-client/src/lib.rs
```diff
@@ -63,14 +65,131 @@ pub struct Config {
...
+                | ConnectionError::Reset
+                | ConnectionError::TimedOut
+                | ConnectionError::LocallyClosed
+        ),
+        ClientError::WriteError(we) => matches!(we, WriteError::ConnectionLost(_)),
+        ClientError::ReadError(re) => matches!(
+            re,
+            ReadExactError::ReadError(quinn::ReadError::ConnectionLost(_))
+        ),
+        _ => false,
```
Rajil1213: I see the documentation but it'd be better to list out the errors we are deeming fatal explicitly.

crates/secret-service-client/src/lib.rs
```diff
@@ -63,14 +65,131 @@ pub struct Config {
...
+    ) -> Result<ServerMessage, ClientError> {
+        let conn = self.current().await;
+        // Try once on the current connection. On dead-connection errors only, reconnect and retry
+        // once more with the fresh connection. Note: we clone `msg` only on the retry path so the
+        // happy path pays just one clone (inside `v2_req` for serialization).
+        match v2_req(&conn, msg.clone(), self.inner.timeout, 10).await {
+            Ok(m) => Ok(m),
+            Err(e) if should_reconnect(&e) => {
+                let new_conn = self.reconnect_if_stale(&conn).await?;
+                v2_req(&new_conn, msg, self.inner.timeout, 10).await
```
Rajil1213: Why is the timeout configurable but the max_retries parameter not? At worst, this is just weird and so I'm fine with keeping this as a constant but use an actual constant here.

Rajil1213 (changes requested):
The changes look good. My comments are mostly about hygiene. The real blocker for this PR is the lack of a functional test. Since there is a real-world issue that this PR is trying to solve, the first commit should ideally be a regression test that demonstrates the issue. The second should be the fix that makes sure that the issue no longer exists. To that end, a functional test needs to be added to the `deposit` group where the secret service is stopped mid deposit, and then restarted. The deposit should still go through (there's already a test in that group that checks whether the bridge node is resilient to restarts).

Zk2u:
Addressed the review:

**Hygiene**
- Extracted the magic `10` retry count to `MAX_TRY_AGAIN_RETRIES: usize = 10` constant near the top of `lib.rs`. Used at both `v2_req` call sites.
- Replaced the `_ => false` catch-all in `should_reconnect` with an exhaustive list of non-reconnect `ClientError` variants. Any new variant added later will fail to compile here, forcing a deliberate classification decision.

**Functional test**
- Added `functional-tests/tests/deposit/fn_deposit_s2_restart_test.py` in the deposit group. Submits a DRT, waits for it to enter the signing phase, stops every s2, restarts each on the same port, verifies the deposit completes.
- Per your suggestion, this is the **first commit** in the PR; against pre-fix code it would fail (the bridge node has no reconnect logic and every signing request post-s2-restart errors). The fix is the **second commit** and makes the test pass.

Force-pushed (`69ca4146 → a11507ab`). Two commits now visible: `test(deposit): demonstrate s2-restart resilience regression` followed by `feat: reconnect logic in secret-service-client`.

Rajil1213:
This needs a rebase. Meanwhile, I'll try to run the functional tests locally to see if the test fails pre-fix.

crates/secret-service-client/src/lib.rs
```diff
@@ -35,11 +36,16 @@ use secret_service_proto::{
...
-use tokio::time::timeout;
+use tokio::{sync::RwLock, time::timeout};
+use tracing::{info, warn};
 use wallet::{GeneralWalletClient, StakechainWalletClient};
 
 const KEEP_ALIVE_INTERVAL: Duration = Duration::from_secs(25);
 
+/// Maximum number of times to retry a request after the server responds with
+/// [`ServerMessage::TryAgain`].
+const MAX_TRY_AGAIN_RETRIES: usize = 10;
```
Rajil1213: Nit: Why the weird name, instead of just `MAX_RETRIES`?

crates/secret-service-client/src/lib.rs
```diff
@@ -63,14 +69,155 @@ pub struct Config {
...
+            re,
+            ReadExactError::ReadError(quinn::ReadError::ConnectionLost(_))
+        ),
+        // Per-request errors that do not indicate a dead connection. Listed exhaustively rather
+        // than caught by a wildcard so any new `ClientError` variant forces a deliberate
+        // classification decision here at compile time.
+        ClientError::SerializationError(_)
+        | ClientError::DeserializationError(_)
+        | ClientError::BadData
+        | ClientError::Timeout
```
Rajil1213: Isn't a timeout subject to a retry?

Rajil1213 (changes requested):

Zk2u:
Addressed:

- `MAX_TRY_AGAIN_RETRIES` → `MAX_RETRIES` (with doc clarifying it's the `TryAgain` retry cap inside `v2_req`).
- `ClientError::Timeout` now triggers reconnect. You're right — a slow server costs at most one wasted reconnect, but a server that's died without yet tripping QUIC's idle-timeout / keep-alive would otherwise keep timing out forever. The single-retry cap inside `make_v2_req` bounds the wasted-work case.
- Updated `should_reconnect_classifies_errors` test to assert the new `Timeout → reconnect` classification.
- Doc comments on both `make_v2_req` and `should_reconnect` updated.

Amended into the fix commit. Latest is `d14ef1a1`.

Rajil1213 (approved):
Looks good to go now (also verified that the regression test fails before the fix and passes afterwards).

PR #553, opened 2026-05-18
refactor: rename stakechain to reserved

Zk2u:
## Summary
- Renames the operator's reserved-funds wallet identifier from `stakechain` to `reserved`
- `SignerTarget::Stakechain` → `Reserved` and `StakeChainGetPreimage` → `GetPreimage` are pure variant renames; rkyv encodes positionally so on-wire bytes are unchanged
- BIP32 paths/indices unchanged — `RESERVED_WALLET_KEY_PATH` derives to `m/20'/103'` exactly as before
- Per review, also drops "stake chain" from the preimages API entirely: `StakeChainPreimages` trait → `Preimages`, `stake_chain_preimages()` method → `preimages()`, `StakeChainPreimgClient` → `PreimgClient`, `StakeChain` seeded-impl struct → `Preimg`, files `stake_chain.rs` → `preimage.rs`, `pub mod stake_chain` → `pub mod preimage`
- `STAKE_CHAIN_WALLET_*` env vars deleted from `.sample_env` (no longer referenced)
- Functional-tests JSON fixture, Python dataclass, factory prop, dev-cli JSON output key, and idempotency-test local vars all renamed in lockstep

## Test plan
- [x] `just lint` passes (`cargo fmt --check`, `taplo fmt --check`, `cargo clippy -- -D warnings`, codespell, orphan-check)
- [x] `cargo check --workspace --tests --all-features` passes
- [x] Repo-wide grep for `stakechain` / `stake_chain` / `StakeChain` returns zero hits
- [ ] functional-tests pass (recommend running before merge given the env-var, JSON, and Python prop renames)

Closes STR-3425.

PR #559, opened 2026-05-19
refactor: remove bdk::Wallet leak from OperatorWallet public API

Zk2u:
**Stacked on #553** (\`azz/str-3425-rename-stakechain-to-reserved\`). Will rebase onto main when #553 merges.

## Summary

Drops the two \`pub\` accessors on \`OperatorWallet\` that returned \`&bdk::Wallet\`:

- \`reserved_wallet()\` had zero callers.
- \`general_wallet()\` had exactly one caller chain (\`graph/common.rs:170\` → \`graph/utils.rs:35-37\`) that used it to fetch prevout \`TxOut\`s for sighash construction.

Refactors \`finalize_claim_funding_tx\` to read prevouts from \`psbt.inputs[i].witness_utxo\` directly. BDK's \`TxBuilder::finish()\` populates \`witness_utxo\` for Taproot inputs (the general wallet uses a \`tr(...)\` descriptor), so this is sound.

## Why

This is the only \`&bdk::Wallet\` leak in \`OperatorWallet\`'s public API. Removing it unblocks the upcoming \`OperatorWalletBackend\` trait extraction (STR-3435), where a non-BDK backend (Fireblocks) cannot satisfy a \`&bdk::Wallet\` return type.

## Out of scope (deferred to STR-3435)

\`OperatorWallet\` still re-exposes BDK's \`LocalOutput\` via \`claim_funding_outputs() -> impl Iterator<Item = LocalOutput>\`. The trait extraction will need to abstract over this — out of scope for this PR's stated "drop \`&bdk::Wallet\` accessors" task.

## Test plan

- [x] \`just lint\` passes
- [x] \`cargo check --workspace --tests --all-features\` passes (verified locally before the diff was finalized)
- [ ] Functional tests (recommend running before merge given the executor signature change)

Closes STR-3426.

### DIFF
--- a/crates/bridge-exec/src/graph/common.rs
+++ b/crates/bridge-exec/src/graph/common.rs
@@ -167,7 +167,6 @@ async fn ensure_claim_funding_outpoint(
                 finalize_claim_funding_tx(
                     &output_handles.s2_client,
                     &output_handles.tx_driver,
-                    wallet.general_wallet(),
                     psbt,
                 )
                 .await?;
--- a/crates/bridge-exec/src/graph/utils.rs
+++ b/crates/bridge-exec/src/graph/utils.rs
@@ -1,7 +1,6 @@
 //! Shared helpers for graph executors.
 
 use algebra::predicate;
-use bdk_wallet::Wallet;
 use bitcoin::{
     Psbt, TapSighashType,
     hashes::Hash,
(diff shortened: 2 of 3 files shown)


Rajil1213 (dismissed):
Looks good!

Rajil1213:
This needs a rebase now. Will approve afterwards.

Rajil1213 (approved):

PR #560, opened 2026-05-19
feat: extract GeneralWallet trait + native impl + migrate bridge-exec

Zk2u:
**Stacked on #559** (`azz/str-3426-drop-bdk-wallet-leak`). Will rebase onto main when #559 merges. **Closes STR-3435 and supersedes STR-3436** (combined into this single PR — see the ticket update for context).

## Summary

Reshapes the operator wallet around a **composition model**:

- A small `GeneralWallet` trait covering only what genuinely differs between backends (UTXO discovery, funding, signing).
- A concrete `OperatorWallet<G: GeneralWallet>` composer holding the always-native reserved wallet, lease bookkeeping, anchor filtering, and cross-wallet transaction construction.

The original `OperatorWalletBackend` design covered everything (general + reserved + leasing + signing) on one trait. That breaks down for Fireblocks because Fireblocks only hosts the general wallet — the reserved wallet stays native, and lease management is identical between backends. The composition design puts only the differing surface on a trait; everything shared lives on the composer.

## What's in this PR

- `crates/operator-wallet/src/general.rs` — `GeneralWallet` trait + `FundedPsbt` + `UtxoInfo`. Methods use AFIT desugared to `impl Future + Send` for codebase consistency.
- `crates/operator-wallet/src/general/native.rs` — `NativeGeneralWallet`, BDK-backed, descriptor-only. All returned PSBTs have `witness_utxo` populated but no signatures — caller signs via secret-service.
- `crates/operator-wallet/src/lib.rs` — `OperatorWallet<G>` composer with all the high-level wallet operations.
- `crates/operator-wallet/src/sync.rs` — `Backend::sync_wallet` no longer takes a lease set; the composer prunes leases after sync by intersecting with the live UTXO set.
- `crates/bridge-exec/src/{deposit,stake/staking,graph/common,graph/counterproof_nack}.rs` — call sites updated.
- `crates/bridge-exec/src/output_handles.rs` — `NativeWallet` type alias = `OperatorWallet<NativeGeneralWallet>`. `Arc<RwLock<_>>` preserved so executors can hold multi-step critical sections.
- `bin/strata-bridge/src/mode/services/{operator_wallet,orchestrator}.rs` — construct `OperatorWallet<NativeGeneralWallet>` at startup.

## What's deferred

- `NativeGeneralWallet::build_cpfp_child` returns `CpfpChildNotImplemented`. **STR-3439** (CPFP wiring in tx-driver) fills the body once there's an integration path to exercise it.
- Fireblocks `GeneralWallet` impl — **STR-3437**.

## CPFP-child contract (for forward reference)

- Native impl: PSBT returned with both inputs unsigned, `witness_utxo` + `tap_internal_key` populated. Caller signs both via secret-service.
- Fireblocks impl (forthcoming): PSBT returned with anchor unsigned, funding input ECDSA-signed by Fireblocks. Caller signs anchor via secret-service.

## Test plan

- [x] `just lint` passes
- [x] `cargo check --workspace --tests --all-features` passes
- [ ] Functional tests (recommend running before merge given the executor API changes)
- [ ] `NativeGeneralWallet::build_cpfp_child` will be exercised by STR-3439's regression test

Closes STR-3435.
Supersedes STR-3436.

### DIFF
--- a/bin/strata-bridge/src/mode/services/orchestrator.rs
+++ b/bin/strata-bridge/src/mode/services/orchestrator.rs
@@ -53,6 +53,7 @@ pub(crate) async fn init_orchestrator<M>(
     req_resp_handle: ReqRespHandle,
     p2p_keypair: Keypair,
     wallet: Arc<RwLock<NativeWallet>>,
+    claim_funding_utxo_value: bitcoin::Amount,
     btc_rpc_client: BitcoinClient,
     asm_rpc_client: HttpClient,
     fdb_client: Arc<FdbClient>,
@@ -114,7 +115,7 @@ where
         retry_tick,
     };
 
-    let exec_cfg = build_exec_config(params, config, &sm_config);
+    let exec_cfg = build_exec_config(params, config, &sm_config, claim_funding_utxo_value);
     let tx_driver = TxDriver::new(zmq_client, btc_rpc_client.clone()).await;
     let ProofBackend { bridge_proof_host } = ProofBackend::new(&config.bridge_proof).await?;
     let output_handles = OutputHandles {
@@ -242,7 +243,12 @@ pub(in crate::mode) fn build_sm_config(config: &Config, params: &Params) -> SMCo
     }
 }
 
-fn build_exec_config(params: &Params, config: &Config, sm_config: &SMConfig) -> ExecutionConfig {
+fn build_exec_config(
+    params: &Params,
+    config: &Config,
+    sm_config: &SMConfig,
+    claim_funding_utxo_value: bitcoin::Amount,
+) -> ExecutionConfig {
     ExecutionConfig {
         network: params.network,
         min_withdrawal_fulfillment_window: config.min_withdrawal_fulfillment_window,
@@ -250,6 +256,7 @@ fn build_exec_config(params: &Params, config: &Config, sm_config: &SMConfig) -> 
         maximum_fee_rate: FeeRate::from_sat_per_vb(config.max_fee_rate).unwrap(),
         operator_fee: params.protocol.operator_fee,
         stake_amount: params.protocol.stake_amount,
+        claim_funding_utxo_value,
         funding_uxto_pool_size: config.operator_wallet.claim_funding_pool_size,
         graph_sm_cfg: sm_config.graph.clone(),
     }
--- a/bin/strata-bridge/src/mode/services/orchestrator.rs
+++ b/bin/strata-bridge/src/mode/services/orchestrator.rs
@@ -8,13 +8,15 @@ use bitcoind_async_client::Client as BitcoinClient;
 use btc_tracker::tx_driver::TxDriver;
 use jsonrpsee::http_client::HttpClient;
 use libp2p_identity::ed25519::Keypair;
-use operator_wallet::OperatorWallet;
 use secret_service_client::SecretServiceClient;
 use strata_bridge_asm_events::client::AsmEventFeed;
 use strata_bridge_common::params::Params;
 use strata_bridge_counterproof::build_bridge_counterproof_host;
 use strata_bridge_db::fdb::client::FdbClient;
-use strata_bridge_exec::{config::ExecutionConfig, output_handles::OutputHandles};
+use strata_bridge_exec::{
+    config::ExecutionConfig,
+    output_handles::{NativeWallet, OutputHandles},
+};
 use strata_bridge_orchestrator::{
     duty_dispatcher::DutyDispatcher, events_mux::EventsMux, persister::Persister,
     pipeline::Pipeline, sm_registry::SMConfig,
@@ -50,7 +52,7 @@ pub(crate) async fn init_orchestrator<M>(
     gossip_handle: GossipHandle,
     req_resp_handle: ReqRespHandle,
     p2p_keypair: Keypair,
-    wallet: Arc<RwLock<OperatorWallet>>,
+    wallet: Arc<RwLock<NativeWallet>>,
     btc_rpc_client: BitcoinClient,
     asm_rpc_client: HttpClient,
     fdb_client: Arc<FdbClient>,
--- a/crates/bridge-exec/src/graph/common.rs
+++ b/crates/bridge-exec/src/graph/common.rs
@@ -4,7 +4,7 @@ use std::num::NonZero;
 
 use algebra::predicate;
 use bitcoin::{
-    OutPoint, TapSighashType, Txid, XOnlyPublicKey,
+    OutPoint, TapSighashType, TxOut, Txid, XOnlyPublicKey,
     hashes::sha256,
     sighash::{Prevouts, SighashCache},
 };
@@ -159,12 +159,23 @@ async fn ensure_claim_funding_outpoint(
             ),
         }
 
-        match wallet.claim_funding_utxo(predicate::never::<UtxoInfo>).0 {
+        match wallet
+            .reserve_utxo_with_value(cfg.claim_funding_utxo_value, predicate::never::<UtxoInfo>)
+            .0
+        {
             Some(outpoint) => outpoint,
             None => {
                 warn!("could not acquire claim funding utxo. attempting refill...");
+                // How many we need to top the pool back up to the configured target. The
+                // existing pool count is read by `reserve_utxo_with_value`'s peer
+                // `reserved_utxos_with_value` — we use the current size to compute the
+                // batch ourselves so the wallet stays denomination-agnostic.
+                let current_pool_size = wallet
+                    .reserved_utxos_with_value(cfg.claim_funding_utxo_value)
+                    .len();
+                let batch_size = cfg.funding_uxto_pool_size.saturating_sub(current_pool_size);
                 let funded = wallet
-                    .refill_claim_funding_utxos(fee::FEE_RATE, cfg.funding_uxto_pool_size)
+                    .create_reserved_utxos(fee::FEE_RATE, cfg.claim_funding_utxo_value, batch_size)
                     .await
                     .map_err(|e| ExecutorError::WalletErr(format!("refill failed: {e}")))?;
                 finalize_claim_funding_tx(
@@ -178,7 +189,10 @@ async fn ensure_claim_funding_outpoint(
                     ExecutorError::WalletErr(format!("wallet sync failed after refill: {e:?}"))
                 })?;
                 wallet
-                    .claim_funding_utxo(predicate::never::<UtxoInfo>)
+                    .reserve_utxo_with_value(
+                        cfg.claim_funding_utxo_value,
+                        predicate::never::<UtxoInfo>,
+                    )
                     .0
                     .expect("funding utxos must be available after refill")
             }
@@ -528,6 +542,7 @@ pub(super) async fn publish_graph_partials(
 
 /// Publishes the claim transaction to Bitcoin.
 pub(super) async fn publish_claim(
+    cfg: &ExecutionConfig,
     output_handles: &OutputHandles,
     claim_tx: &ClaimTx,
 ) -> Result<(), ExecutorError> {
@@ -538,14 +553,14 @@ pub(super) async fn publish_claim(
         "signing claim transaction"
     );
 
-    let claim_prevout = {
+    let claim_prevout: TxOut = {
         let wallet = output_handles.wallet.read().await;
         wallet
-            .claim_funding_outputs()
+            .reserved_utxos_with_value(cfg.claim_funding_utxo_value)
             .into_iter()
             .find(|utxo| utxo.outpoint == claim_tx.as_ref().input[0].previous_output)
             .expect("claim funding outpoint not found in wallet")
-            .as_txout()
+            .into()
     };
 
     let prevouts = Prevouts::All(&[claim_prevout]);
--- a/crates/bridge-exec/src/graph/common.rs
+++ b/crates/bridge-exec/src/graph/common.rs
@@ -166,13 +166,21 @@ async fn ensure_claim_funding_outpoint(
             Some(outpoint) => outpoint,
             None => {
                 warn!("could not acquire claim funding utxo. attempting refill...");
-                // How many we need to top the pool back up to the configured target. The
-                // existing pool count is read by `reserve_utxo_with_value`'s peer
-                // `reserved_utxos_with_value` — we use the current size to compute the
-                // batch ourselves so the wallet stays denomination-agnostic.
-                let current_pool_size = wallet
-                    .reserved_utxos_with_value(cfg.claim_funding_utxo_value)
-                    .len();
+                // How many we need to top the pool back up to the configured target. We
+                // compute the batch ourselves (the wallet stays denomination-agnostic),
+                // counting only *unleased* pool members: `reserved_utxos_with_value`
+                // returns every matching UTXO including leased ones, but a leased UTXO is
+                // already committed to another graph and can't satisfy this reservation.
+                // Counting them would understate the deficit and could yield a zero-size
+                // batch, after which the post-refill `reserve_utxo_with_value` below would
+                // panic with nothing to hand out.
+                let current_pool_size = {
+                    let pool = wallet.reserved_utxos_with_value(cfg.claim_funding_utxo_value);
+                    let leased = wallet.leased_outpoints();
+                    pool.iter()
+                        .filter(|u| !leased.contains(&u.outpoint))
+                        .count()
+                };
                 let batch_size = cfg.funding_uxto_pool_size.saturating_sub(current_pool_size);
                 let funded = wallet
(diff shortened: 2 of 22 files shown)


Rajil1213:
> just lint passes

Are you sure?

Rajil1213:
You should also just set the target branch to `main`. We review PRs commit by commit. So we'll just ignore the commits from the earlier PRs. That way, github won't discard older reviews when you change the base branch on a PR.

Zk2u:
@Rajil1213 github automatically changes the PR base when the parent PR is merged in, see above :P i didn't know this was a feature until recently

crates/operator-wallet/src/lib.rs
```diff
@@ -1,34 +1,51 @@
...
+//! - anchor identification and exclusion,
+//! - cross-wallet transaction construction (refill claim-funding pool, create stake funding tx).
+//!
+//! The `general` backend handles only what genuinely varies between native and Fireblocks:
+//! its own UTXO discovery, its own signing, and its share of transaction building (funding,
+//! CPFP child).
+//!
+//! Methods on [`OperatorWallet`] are `&mut self`. Callers serialize via an outer lock (today
+//! `Arc<RwLock<OperatorWallet<_>>>` in `bridge-exec`) which also lets the executor span
+//! multi-step critical sections (e.g. DB-lookup-then-fund-then-persist).
```
Rajil1213: This is not a durable API doc. In general, an API should not be concerned about specific downstream implementations.

crates/operator-wallet/src/lib.rs
```diff
@@ -1,34 +1,51 @@
...
+pub use crate::general::{native::NativeGeneralWallet, FundedPsbt, GeneralWallet, UtxoInfo};
+use crate::sync::{Backend, SyncError};
+
+/// How many times we should reattempt after an error during a wallet sync.
 const SYNC_RETRIES: u32 = 5;
-/// The wallet will delay a retry by SYNC_BASE_DELAY*SYNC_BACKOFF^current_retry,
-/// exponential backoff
+/// The wallet will delay a retry by SYNC_BASE_DELAY * SYNC_BACKOFF.pow(current_retry).
 const SYNC_BACKOFF: u32 = 3;
 const SYNC_BASE_DELAY: Duration = Duration::from_millis(100);
```
Rajil1213: Suggestion: move these to the `OperatorWalletConfig` (these values stay as default).

crates/operator-wallet/src/lib.rs
```diff
@@ -47,394 +64,340 @@ impl OperatorWalletConfig {
...
+    pub fn lease(&mut self, outpoints: &[OutPoint]) {
+        for outpoint in outpoints {
+            self.leased_outpoints.insert(*outpoint);
+        }
     }
 
-    fn is_leased_pred<'a>(&'a self) -> impl Fn(&OutPoint) -> bool + 'a {
-        |outpoint| self.leased_outpoints.contains(outpoint)
+    /// Removes each of `outpoints` from the lease set. Logs (at `warn`) but does not error
+    /// when a given outpoint wasn't leased — releases are idempotent.
```
Rajil1213: Logging is an implementation detail. This doc comment should only talk about the safety invariant (i.e., it is safe to call this repeatedly).

crates/operator-wallet/src/lib.rs
```diff
@@ -47,394 +64,340 @@ impl OperatorWalletConfig {
...
+            }
+        }
     }
 
-    fn release(&mut self, outpoint: &OutPoint) -> bool {
-        self.leased_outpoints.remove(outpoint)
+    // ── Reserved-wallet operations ─────────────────────────────────────────
+
+    /// Returns the reserved-wallet UTXOs that match the claim-funding value (i.e. the dust
+    /// pool used to fund claim transactions).
```
Rajil1213: Why is this being called "dust pool"?

crates/operator-wallet/src/lib.rs
```diff
@@ -47,394 +64,340 @@ impl OperatorWalletConfig {
...
-            .iter()
-            .copied()
-            .chain(anchor_outpoints)
-            .chain(leased_outpoints)
+        let current_size = claim_funding_outpoints.len();
+        let batch_size = target_size.saturating_sub(current_size);
+
+        let outputs = (0..batch_size)
+            .map(|_| TxOut {
+                value: self.config.stake_funding_utxo_value,
```
Rajil1213: The amount of claim funding utxo is not the same as that of the stake funding utxo. We should separate these two even if they had the same amount. I realize now that this bug also existed in the previous code. I don't want to be too prescriptive but see if you can come up with a general enough API for this (where we create UTXOs in the reserved wallet address of a specific size based on the use case). The simplest approach is to just let the caller handle this (it's already controlling the `target_size` and `fee_rate`) but then the function name will have to change.

crates/operator-wallet/src/lib.rs
Rajil1213, line 1: It might be time to split this file into sub-modules (`config`, `wallet`). You should also add unit tests (use `corepc-node` to spawn a bitcoind instance on regtest). There aren't any tests but that's only because when this crate was originally created velocity was prioritized above quality. Now, we can afford to invest in proper testing.

crates/operator-wallet/src/general.rs
```diff
@@ -0,0 +1,124 @@
+//! The `GeneralWallet` trait — abstraction over the part of an operator's funds that pays for
+//! bridge operations (fronting withdrawals, CPFP, top-ups). The trait isolates the surface that
+//! genuinely differs between backends: today the native BDK-backed wallet, tomorrow a Fireblocks
+//! adapter.
```
Rajil1213: No need to mention Fireblocks as _the_ adaptor of tomorrow. It's just one of many that we might end up supporting that this API generalizes over.

crates/operator-wallet/src/general.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Returns the receive script for this wallet. Stable across calls for native backends;
+    /// may rotate for backends that mint fresh deposit addresses per call.
+    fn script_pubkey(&self) -> ScriptBuf;
+
+    /// Returns every UTXO this wallet currently controls (confirmed and unconfirmed). The
+    /// caller is responsible for filtering anchors, leases, and other domain-specific
+    /// exclusions before requesting funding.
+    fn list_utxos(&self) -> Vec<UtxoInfo>;
+
+    /// Builds and (where possible) signs a v3 TRUC funding transaction.
```
Rajil1213: This is a pretty confusing language. Just say that it signs the inputs that it has key material for.

crates/operator-wallet/src/general.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// Builds a v3 TRUC CPFP child for `parent`, spending the BIP-431 ephemeral anchor at
+    /// `parent.output[anchor_vout]` plus one fee-paying input drawn from this wallet.
+    ///
+    /// * `target_pkg_fee_rate` — sat-per-vbyte target for the (parent, child) package as a whole.
+    /// * `exclude` — fee-paying-input selection skips these outpoints. Used to avoid re-selecting
+    ///   the funding input of a prior child being replaced via RBF.
+    ///
+    /// **The anchor input is never signed** — backends populate its `witness_utxo` and
+    /// `tap_internal_key` and leave the caller to sign it via the secret service. The
```
Rajil1213: No need to mention the secret service here -- it is just one way to sign it externally. API docs must be generic.

crates/operator-wallet/src/general.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// A funded PSBT returned by [`GeneralWallet`] funding operations.
+#[derive(Debug, Clone)]
+pub struct FundedPsbt {
+    /// The funded PSBT. Inputs the backend could sign are signed; the rest carry
+    /// `witness_utxo` (and `tap_internal_key` for Taproot) for downstream signing.
+    pub psbt: Psbt,
+    /// Outpoints consumed as inputs to this PSBT. The caller leases these so they aren't
+    /// re-selected by concurrent duties.
+    pub spent: Vec<OutPoint>,
```
Rajil1213: Why is this re-exposed? Doesn't the `Psbt` already have this?

crates/operator-wallet/src/general.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// A snapshot of a single UTXO controlled by a [`GeneralWallet`] (or, by convention, the
+/// reserved wallet that the [`crate::OperatorWallet`] composer manages internally).
+#[derive(Debug, Clone)]
+pub struct UtxoInfo {
+    /// On-chain outpoint of the UTXO.
```
Rajil1213: Might not be _on_ chain if confirmations is 0.

crates/operator-wallet/src/general.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl UtxoInfo {
+    /// Reconstructs the [`TxOut`] this UTXO refers to.
+    pub fn as_txout(&self) -> TxOut {
+        TxOut {
+            value: self.amount,
+            script_pubkey: self.script_pubkey.clone(),
+        }
+    }
+}
```
Rajil1213: This looks like an `Into` impl to me.

crates/operator-wallet/src/general/native.rs
```diff
@@ -0,0 +1,177 @@
+//! Native BDK-backed implementation of [`GeneralWallet`].
+//!
+//! The native wallet holds the operator's general-funds descriptor (`tr(general_pubkey)`) but
+//! never holds private keys: signing is delegated to the secret-service by the caller. As a
```
Rajil1213: No need to mention the secret-service here (it's an implementation detail of the consumer not this library).

crates/operator-wallet/src/lib.rs
```diff
@@ -47,394 +64,340 @@ impl OperatorWalletConfig {
...
+            .map(|lo| UtxoInfo {
+                outpoint: lo.outpoint,
+                amount: lo.txout.value,
+                confirmations: match &lo.chain_position {
+                    ChainPosition::Confirmed { anchor, .. } => {
+                        tip.saturating_sub(anchor.block_id.height).saturating_add(1)
+                    }
+                    ChainPosition::Unconfirmed { .. } => 0,
+                },
+                script_pubkey: lo.txout.script_pubkey.clone(),
```
Rajil1213: Why is this not using the `local_output_to_utxo_info` function created in the `native` module? (also is `native` module the right home for this function?)

crates/operator-wallet/src/general.rs
```diff
@@ -69,9 +69,13 @@ pub trait GeneralWallet: Send + Sync {
         exclude: &[OutPoint],
     ) -> impl std::future::Future<Output = Result<FundedPsbt, Self::Error>> + Send;
 
-    /// Builds a v3 TRUC CPFP child for `parent`, spending the BIP-431 ephemeral anchor at
+    /// Builds a v3 TRUC CPFP child for `parent`, spending the keyed-Taproot anchor at
     /// `parent.output[anchor_vout]` plus one fee-paying input drawn from this wallet.
     ///
+    /// The project uses a keyed-Taproot anchor at the segwit-dust minimum (`SEGWIT_MIN_AMOUNT`,
+    /// 330 sat) rather than a BIP-431 ephemeral P2A (value-0) output. Implementers must spend
+    /// the 330-sat anchor and account for its value when constructing the child.
```
Rajil1213: This commentary does not belong here. A consumer can do what they wish -- it's their responsibility to implement this trait in a way that meets their use cases.

Rajil1213 (changes requested):
I like the general shape of the `GeneralWallet` trait. There are only a couple of blockers:

- How `refill_*` is/was implemented
- The lack of any crate-level tests.

bin/strata-bridge/src/mode/services/orchestrator.rs
```diff
@@ -241,14 +242,20 @@ pub(super) fn build_sm_config(config: &Config, params: &Params) -> SMConfig {
...
+    claim_funding_utxo_value: bitcoin::Amount,
+) -> ExecutionConfig {
     ExecutionConfig {
         network: params.network,
         min_withdrawal_fulfillment_window: config.min_withdrawal_fulfillment_window,
         magic_bytes: params.protocol.magic_bytes,
         maximum_fee_rate: FeeRate::from_sat_per_vb(config.max_fee_rate).unwrap(),
         operator_fee: params.protocol.operator_fee,
         stake_amount: params.protocol.stake_amount,
+        claim_funding_utxo_value,
```
Rajil1213, line 259: If this value is computed once and then added to the initialized operator wallet as a field, why does it also need to be in the config?

The most durable approach is to change the API so that the caller can pass the UTXO value (which is already the case).

This is because this amount depends on the number of watchtowers which itself might change at runtime during operator entry/exit scenarios.

However, this might be too big of an unrelated change for this PR, so I'd be fine with just creating a ticket for this.

crates/bridge-exec/src/graph/common.rs
```diff
@@ -159,12 +159,23 @@ async fn ensure_claim_funding_outpoint(
...
             Some(outpoint) => outpoint,
             None => {
                 warn!("could not acquire claim funding utxo. attempting refill...");
+                // How many we need to top the pool back up to the configured target. The
+                // existing pool count is read by `reserve_utxo_with_value`'s peer
+                // `reserved_utxos_with_value` — we use the current size to compute the
+                // batch ourselves so the wallet stays denomination-agnostic.
+                let current_pool_size = wallet
+                    .reserved_utxos_with_value(cfg.claim_funding_utxo_value)
+                    .len();
```
Rajil1213: Doesn't the fact that we're in this arm mean that this value is zero?

crates/operator-wallet/tests/operator_wallet_e2e.rs
```diff
@@ -0,0 +1,483 @@
+#![allow(unused_crate_dependencies)]
```
Rajil1213: Use `expect` instead of `allow` and add a `reason = ` to explain why this lint is expected to fail.

Rajil1213 (changes requested):
This is almost good to go now. I only have a few nits and questions. I have yet to have codex take a look at this though. Will report any actionable findings from that run.

crates/bridge-exec/src/graph/common.rs
```diff
@@ -158,24 +159,40 @@ async fn ensure_claim_funding_outpoint(
...
             Some(outpoint) => outpoint,
             None => {
                 warn!("could not acquire claim funding utxo. attempting refill...");
-                let psbt =
-                    wallet.refill_claim_funding_utxos(fee::FEE_RATE, cfg.funding_uxto_pool_size)?;
+                // How many we need to top the pool back up to the configured target. The
+                // existing pool count is read by `reserve_utxo_with_value`'s peer
+                // `reserved_utxos_with_value` — we use the current size to compute the
+                // batch ourselves so the wallet stays denomination-agnostic.
+                let current_pool_size = wallet
```
Rajil1213: This computation _must_ exclude any leased UTXOs. Otherwise, we might exhaust the usable UTXO set and then create a batch of zero size.

Rajil1213 (changes requested):
Codex found an issue that needs to be addressed.

crates/bridge-exec/src/graph/common.rs
Zk2u: Ok, fixed - `reserved_utxos_with_value` returns every matching UTXO *including leased ones*, so counting them understated the deficit and could yield a zero-size batch — after which the post-refill `reserve_utxo_with_value` would panic with no UTXOs

The refill now counts only unleased members (filters against `leased_outpoints()`)

crates/bridge-exec/src/graph/common.rs
Zk2u: Right — with the current call (predicate `never`), reaching this arm implies zero *usable* unleased UTXOs, so with the leased-exclusion fix `current_pool_size` is effectively 0 and `batch_size` == the full pool target

The explicit count was kept rather than hardcoding 0 so it stays correct if the `ignore` predicate ever becomes non-trivial (unleased-but-ignored UTXOs shouldn't count toward the target). It's cheap to be correct and self-documenting

bin/strata-bridge/src/mode/services/orchestrator.rs
Zk2u: Filed STR-3641 to move it to caller-supplied-at-point-of-use and kept it out of scope here: https://alpenlabs.atlassian.net/browse/STR-3641

Rajil1213:
There is a compilation issue with the latest changes that needs fixing. Will review afterwards.

Rajil1213 (approved):
This looks good to go now.

PR #563, opened 2026-05-21
feat(bridge-exec): introduce FeeSource trait with BitcoinCore + MempoolExplorer policies (STR-3438)

Zk2u:
Closes [STR-3438](https://alpenlabs.atlassian.net/browse/STR-3438).

## Summary

Replaces the two direct \`estimate_smart_fee(1)\` calls in \`deposit.rs\` and \`stake/staking.rs\` with a configurable fee-source abstraction. Mirrors strata's btcio shape (per Slack discussion with @Rajil1213 and @bewakes), kept bridge-local per Rajil's call against a shared crate.

## Design

\`FeeSource\` trait + 3 impls in \`crates/bridge-exec/src/fees.rs\`:

| Policy | Behaviour |
|---|---|
| \`BitcoinCore { conf_target }\` | Wraps \`estimatesmartfee\` |
| \`MempoolExplorer { base_url, policy, fallback_conf_target }\` | GETs \`/api/v1/fees/recommended\` on a mempool.space-compatible explorer, with BitcoinCore as the fallback on any HTTP/decode failure (mempool error absorbed, only logged — a downed explorer must never block tx publishing) |
| \`Fixed { fee_rate }\` | Tests / manual overrides |

Operator-side config selects via \`FeeSourceConfig\` (snake_case tagged enum), built into \`Arc<dyn FeeSource>\` at orchestrator boot. \`#[serde(default)]\` → \`BitcoinCore { conf_target: 1 }\`, preserving pre-PR behaviour for existing operator config.toml files.

## Deliberate divergences from strata's impl

- **No \`×2\` multiplier**. Strata doubles the rate as a temporary hedge until proper fee bumping lands (and wants to drop it). The bridge has STR-3439 (CPFP + RBF in tx-driver) for that — we skip the hack.
- **No shared crate**. Per Rajil: \"pretty straightforward and probably overkill to have a separate crate for.\"
- **Clamp adopted**: \`max(rate, 1)\` per Jose's STR-2018 fix (alpenlabs/alpen#1811) against \`bitcoind-async-client\`'s sub-1 sat/vB truncation.

## Hardening from local audit

- 10s timeout on \`SHARED_HTTP_CLIENT\` — reqwest default is no timeout; a hanging mempool explorer would otherwise block the duty future indefinitely (the fallback only fires on errors, not on never-resolving futures).
- \`clamp_to_min\` uses \`unwrap_or_else\` rather than \`expect\` to guard against absurd inputs (e.g. \`Fixed { fee_rate: u64::MAX }\`).

## Scope notes

The ticket cited three call sites but the third (\`graph/common.rs:166\`) is a hardcoded \`fee::FEE_RATE\` constant for the claim-funding refill self-spend, not a live \`estimate_smart_fee\` call. Left as-is — that's a different conversation about whether self-spend refills should use live estimates.

## Responsibility split (for context)

The \`FeeSource\` trait is stateless. Callers fetch once per tx-build. Live fee tracking — periodically re-checking to drive RBF bumps — is the tx-driver's job once CPFP wiring lands in STR-3439. The trait deliberately does not poll, cache, or push.

## Test plan

- [x] \`cargo fmt --all && cargo clippy --workspace --all-targets --all-features\` clean
- [x] 15 fee-source unit tests (happy paths, fallback on 5xx / malformed JSON / both fail, URL trailing-slash handling, overflow safety, default check, toml roundtrip for all variants incl. multi-word \`half_hour\`)
- [ ] CI green

[STR-3438]: https://alpenlabs.atlassian.net/browse/STR-3438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/strata-bridge/src/config.rs
+++ b/bin/strata-bridge/src/config.rs
@@ -20,6 +20,7 @@ use serde::{Deserialize, Serialize};
 use strata_bridge_asm_events::config::AsmRpcConfig;
 pub(crate) use strata_bridge_counterproof::ProofBackendConfig as CounterproofBackendConfig;
 use strata_bridge_db::fdb::cfg::Config as FdbConfig;
+use strata_bridge_exec::fees::FeeSourceConfig;
 use strata_bridge_p2p_service::GossipsubScoringPreset;
 pub(crate) use strata_bridge_proof::ProofBackendConfig;
 
@@ -74,6 +75,12 @@ pub(crate) struct Config {
     /// reaction to fee-market moves between blocks.
     #[serde(default = "default_cpfp_bump_check_interval")]
     pub cpfp_bump_check_interval: Duration,
+
+    /// Source of fee-rate estimates for bridge transactions. Defaults to Bitcoin Core with
+    /// `conf_target = 1`; operators on networks where `estimatesmartfee` is unreliable should
+    /// configure the mempool-explorer variant.
+    #[serde(default)]
+    pub fee_source: FeeSourceConfig,
 
     /// Configuration required to connector to a _local_ instance of the secret service server.
     pub secret_service_client: SecretServiceConfig,
--- a/bin/strata-bridge/src/config.rs
+++ b/bin/strata-bridge/src/config.rs
@@ -15,6 +15,16 @@ const fn default_cpfp_bump_check_interval() -> Duration {
     Duration::from_secs(30)
 }
 
+/// Default CPFP fee premium (percent) — see [`Config::cpfp_fee_premium_percent`].
+const fn default_cpfp_fee_premium_percent() -> u32 {
+    5
+}
+
+/// Default CPFP package fee floor (sats/vbyte) — see [`Config::cpfp_min_package_fee_rate`].
+const fn default_cpfp_min_package_fee_rate() -> u64 {
+    10
+}
+
 use libp2p::Multiaddr;
 use serde::{Deserialize, Serialize};
 use strata_bridge_asm_events::config::AsmRpcConfig;
@@ -61,6 +71,20 @@ pub(crate) struct Config {
 
     /// The maximum fee rate for any transaction (in sats/vb).
     pub max_fee_rate: u64,
+
+    /// Premium applied on top of the next-block fee estimate when pricing a CPFP bump target,
+    /// as a percentage of the estimate (e.g. `5` = +5%). Rounded up, so the premium raises
+    /// the target and never lowers it. `0` prices the child at the raw estimate. Default 5,
+    /// per the product fee decision.
+    #[serde(default = "default_cpfp_fee_premium_percent")]
+    pub cpfp_fee_premium_percent: u32,
+
+    /// Floor (in sats/vb) on the package fee rate the CPFP ladder targets. The child pays
+    /// the package shortfall, so the package always lands at or above this value.
+    /// Default 10, per the product fee decision. Must not exceed `max_fee_rate`.
+    /// The startup checks reject a floor above the cap.
+    #[serde(default = "default_cpfp_min_package_fee_rate")]
+    pub cpfp_min_package_fee_rate: u64,
 
     /// Background-refresh cadence for the cached fee rate. The bridge spawns a tokio task at
     /// startup that polls the configured fee source at this interval and stores the result in
--- a/crates/bridge-exec/src/deposit.rs
+++ b/crates/bridge-exec/src/deposit.rs
@@ -25,11 +25,8 @@ use strata_bridge_primitives::{
     types::{BitcoinBlockHeight, DepositIdx, GraphIdx, OperatorIdx},
 };
 use strata_bridge_sm::deposit::duties::{DepositDuty, NagDuty};
-use strata_bridge_tx_graph::{
-    fee,
-    transactions::prelude::{
-        CooperativePayoutTx, SweepTx, WithdrawalFulfillmentData, WithdrawalFulfillmentTx,
-    },
+use strata_bridge_tx_graph::transactions::prelude::{
+    CooperativePayoutTx, SweepTx, WithdrawalFulfillmentData, WithdrawalFulfillmentTx,
 };
 use tracing::{error, info, warn};
 
@@ -37,6 +34,7 @@ use crate::{
     chain::{self, CpfpKind, is_txid_onchain, publish_signed_transaction},
     config::ExecutionConfig,
     errors::ExecutorError,
+    fees::wallet_tx_fee_rate,
     output_handles::OutputHandles,
 };
 
@@ -505,41 +503,19 @@ async fn fulfill_withdrawal(
         .checked_sub(cfg.operator_fee)
         .expect("deposit amount must be greater than operator fee");
 
-    // Get fee rate estimate from bitcoind, bounded from below by `fee::FEE_RATE` so the
-    // withdrawal-fulfillment v3 transaction always meets the bridge's hardcoded minimum even
-    // on networks where `estimatesmartfee` returns a value below `minrelaytxfee`.
-    let smart_fee = output_handles
-        .bitcoind_rpc_client
-        .estimate_smart_fee(1)
-        .await
-        .map_err(|e| ExecutorError::WalletErr(format!("failed to estimate fee: {e}")))?;
-    info!(?smart_fee, "estimated fee rate for withdrawal fulfillment");
-
-    let fee_rate = smart_fee
-        .fee_rate
-        .unwrap_or(fee::FEE_RATE)
-        .max(fee::FEE_RATE);
-
-    // The following approach trades off maximal liveness for maximal safety:
-    // It is not safe to broadcast at a lower fee rate when the network fee rate is high as there is
-    // a chance that the fulfillment transaction will be settled after a reassignment, causing an
-    // operator to lose funds. The safest approach is to abort.
-    if fee_rate > cfg.maximum_fee_rate {
-        return Err(ExecutorError::FeeRateTooHigh {
-            fee_rate,
-            max: cfg.maximum_fee_rate,
-        });
-    }
-
-    // Check if fee rate exceeds maximum configured rate
-    if fee_rate > cfg.maximum_fee_rate {
-        warn!(
-            %fee_rate,
-            maximum_fee_rate = %cfg.maximum_fee_rate,
-            "fee rate exceeds maximum, skipping fulfillment"
-        );
-        return Ok(());
-    }
+    // Price from the shared cached fee source, floored at `MIN_WALLET_TX_FEE_RATE` so the
+    // withdrawal-fulfillment v3 transaction stays relayable. The network-backed sources clamp
+    // their reports to the ≥1 sat/vB truncation guard (`FixedFeeSource` returns its rate
+    // verbatim); the helper's floor is the higher bridge-policy minimum either way.
+    //
+    // A rate above `maximum_fee_rate` rejects with `FeeRateTooHigh`, trading liveness for
+    // safety: broadcasting at a capped-down rate risks the fulfillment settling after a
+    // reassignment, costing the operator funds. The error keeps the duty failure visible in
+    // observability, matching how `stake::staking::estimate_funding_fee_rate` surfaces the
+    // same condition; the duty dispatcher retries, and if the fee market stays above the cap
+    // the operator notices via the error log instead of a silent skip.
+    let fee_rate = wallet_tx_fee_rate(&cfg.fee_source, cfg.maximum_fee_rate)?;
+    info!(%fee_rate, "fee rate for withdrawal fulfillment");
 
     info!(%deposit_idx, "pre-conditions satisfied, attempting to fulfill withdrawal request");
 
--- a/crates/bridge-exec/src/fees.rs
+++ b/crates/bridge-exec/src/fees.rs
@@ -80,6 +80,14 @@ pub enum FeeSourceError {
     /// The configured mempool explorer URL is malformed.
     #[error("invalid mempool explorer url: {0}")]
     InvalidConfig(String),
+    /// A configured confirmation target is outside Bitcoin Core's accepted `1..=1008` range.
+    #[error("{field} = {target} is outside Bitcoin Core's estimatesmartfee range (1..=1008)")]
+    InvalidConfTarget {
+        /// Which config field carried the bad value.
+        field: &'static str,
+        /// The rejected value.
+        target: u16,
+    },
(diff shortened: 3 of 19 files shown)


bin/strata-bridge/src/config.rs
```diff
@@ -48,6 +49,12 @@ pub(crate) struct Config {
     /// The maximum fee rate for any transaction (in sats/vb).
     pub max_fee_rate: u64,
 
+    /// Source of fee-rate estimates for bridge transactions. Defaults to Bitcoin Core with
+    /// `conf_target = 1` (matching the bridge's historical behaviour); operators on networks
```
Rajil1213:
```suggestion
    /// `conf_target = 1`; operators on networks
```

Historical context does not belong here.

crates/bridge-exec/src/stake/staking.rs
```diff
@@ -155,23 +154,19 @@ async fn read_or_create_stake_funding(
...
+        .fee_source
+        .estimate()
+        .await
+        .map_err(|e| ExecutorError::WalletErr(format!("failed to estimate fee: {e}")))?;
+    info!(%estimated, "fetched fee rate from fee source");
+
+    // Floor at `fee::FEE_RATE` so this v3 (TRUC) funding transaction always meets the bridge's
+    // hardcoded minimum. The fee source already clamps to >=1 sat/vB internally; this floor is
+    // the bridge-protocol minimum, distinct from the truncation guard.
+    let fee_rate = estimated.max(fee::FEE_RATE);
```
Rajil1213: `fee::FEE_RATE` isn't the minimum. It's a static fee rate for presigned transactions (till we have proper CPFP). You should create a different constant for this.

crates/bridge-exec/src/deposit.rs
```diff
@@ -405,19 +405,18 @@ async fn fulfill_withdrawal(
...
+        .estimate()
         .await
         .map_err(|e| ExecutorError::WalletErr(format!("failed to estimate fee: {e}")))?;
-    info!(%fee_rate, "estimated fee rate for withdrawal fulfillment");
+    info!(%estimated, "estimated fee rate for withdrawal fulfillment");
 
-    let fee_rate = FeeRate::from_sat_per_vb(fee_rate)
-        .unwrap_or(fee::FEE_RATE)
-        .max(fee::FEE_RATE);
+    let fee_rate = estimated.max(fee::FEE_RATE);
```
Rajil1213: Update both the comment and the code to floor at a different constant for the minimum (or remove it if it's already clamped).

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+//! * [`BitcoindFeeSource`] — queries `estimatesmartfee` on the local Bitcoin Core RPC.
+//! * [`MempoolExplorerFeeSource`] — queries a mempool.space-compatible explorer's
+//!   `/api/v1/fees/recommended` endpoint; falls back to a wrapped [`BitcoindFeeSource`] on any HTTP
+//!   or decode failure so a downed mempool explorer never blocks tx publishing.
+//! * [`FixedFeeSource`] — returns a constant rate (tests, manual overrides).
+//!
+//! All sources clamp the returned rate to at least 1 sat/vB. `bitcoind-async-client` represents
+//! fee rates as `u64` sat/vB, so any sub-1 sat/vB estimate from the upstream source would
+//! otherwise truncate to zero and lose the fee entirely. This mirrors strata's fix in
+//! [alpenlabs/alpen#1811](https://github.com/alpenlabs/alpen/pull/1811).
```
Rajil1213: This leaks implementation details. Just preserve the first sentence here.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+//!
+//! Callers fetch via [`FeeSource::estimate`] once per tx-build. Periodic re-fetching for live
+//! fee tracking (e.g., to drive RBF bumps) is the responsibility of the caller, not the source;
+//! that loop will live in the tx-driver layer once CPFP wiring lands
+//! ([STR-3439](https://alpenlabs.atlassian.net/browse/STR-3439)).
```
Rajil1213: This is a TODO/FIXME comment, not an API doc.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+///
+/// Implementations are stateless with respect to fee tracking — each call hits the underlying
+/// source. Callers responsible for cadence; see module docs.
+///
+/// Uses `async_trait` rather than AFIT so the trait is object-safe and can be stored as
+/// `Arc<dyn FeeSource>` on [`crate::config::ExecutionConfig`]. Allocation cost per call is one
+/// boxed future, negligible against the network round-trip the call performs.
+#[async_trait]
+pub trait FeeSource: Send + Sync + std::fmt::Debug {
+    /// Returns the current fee-rate estimate, clamped to at least 1 sat/vB.
```
Rajil1213: This is just a trait doc -- it cannot guarantee that the implementation will clamp the fee rate to at least 1 sat/vb.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+}
+
+/// A source of fee-rate estimates for bridge transactions.
+///
+/// Implementations are stateless with respect to fee tracking — each call hits the underlying
+/// source. Callers responsible for cadence; see module docs.
+///
+/// Uses `async_trait` rather than AFIT so the trait is object-safe and can be stored as
+/// `Arc<dyn FeeSource>` on [`crate::config::ExecutionConfig`]. Allocation cost per call is one
+/// boxed future, negligible against the network round-trip the call performs.
```
Rajil1213: These are implementation notes that don't belong on the trait API doc. Make this a normal comment instead.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+    client: Arc<R>,
+    conf_target: u16,
+}
+
+impl<R: FeeRateRpc> BitcoindFeeSource<R> {
+    /// Creates a new source that queries `estimatesmartfee(conf_target)` on `client`.
+    pub const fn new(client: Arc<R>, conf_target: u16) -> Self {
+        Self {
+            client,
+            conf_target,
```
Rajil1213, line 122: To make this API easier to use, `conf_target` should be an argument to `estimate`, not part of the struct. For example, we should be able to call `tx_driver::drive(...)` with a deadline parameter and it should estimate fees based on that, the closer the deadline is the more aggressive should the fee rate be (without having to reconstruct the source).

Even semantically, any `FeeSource` must only include the source itself (the `client`), not how aggressive the estimate should be.

Rajil1213: For mempool explorer, you can map the fee policy to the confirmation target (`Fastest` if deadline is in the next couple of blocks, `HalfHour` if we have several blocks to go until the deadline, etc.). Feel free to use reasonable boundaries here.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+/// a Bitcoin Core fallback.
+///
+/// Falls back to the wrapped [`BitcoindFeeSource`] on any HTTP error, non-2xx status, or decode
+/// failure. The fallback error replaces the mempool error in the result; the mempool error is
+/// only logged. This matches strata's behaviour: a downed mempool explorer must never block tx
+/// publishing.
+#[derive(Debug, Clone)]
+pub struct MempoolExplorerFeeSource<R> {
+    recommended_fees_url: Url,
+    policy: MempoolFeePolicy,
```
Rajil1213, line 250: As in the case of `BitcoindFeeSource`, `policy` does not belong here.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+///
+/// Built into a concrete `Arc<dyn FeeSource>` at startup via [`FeeSourceConfig::build`], taking
+/// the operator's Bitcoin Core RPC client as the fallback source for the mempool variant.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub enum FeeSourceConfig {
+    /// Query Bitcoin Core's `estimatesmartfee(conf_target)` directly.
+    BitcoinCore {
+        /// Block confirmation target passed to `estimatesmartfee`. `1` mirrors the bridge's
+        /// historical behaviour.
```
Rajil1213: Remove the note about historical behavior. Given no context (except this enum itself), it's also just confusing.

crates/bridge-exec/src/fees.rs
```diff
@@ -0,0 +1,651 @@
...
+/// sub-1 sat/vB representation lands upstream.
+///
+/// On the upper end, `FeeRate::from_sat_per_vb` overflows its internal sat/kwu representation
+/// at sat/vB > u64::MAX/250 ≈ 7.4×10^16, far beyond anything `estimatesmartfee` could ever
+/// return. Falls back to 1 sat/vB on that overflow rather than panicking — defensive against
+/// stubbed callers, not a real production path.
+fn clamp_to_min(raw_sat_per_vb: u64) -> FeeRate {
+    let clamped = raw_sat_per_vb.max(1);
+    FeeRate::from_sat_per_vb(clamped)
+        .unwrap_or_else(|| FeeRate::from_sat_per_vb(1).expect("1 sat/vB is always a valid FeeRate"))
```
Rajil1213: Refactor the min fee rate to a const and use it wherever we need a min fee rate value.

Rajil1213 (changes requested):
The only major comment here is in regards to how this API will be exercised by the `tx-driver` during RBF in case a transaction does not confirm within the supplied deadline. The rest are mostly about improving doc comments.

Also, _love_ how the unit tests are written!

Rajil1213:
> The ticket cited three call sites but the third (`graph/common.rs:166`) is a hardcoded `fee::FEE_RATE` constant for the claim-funding refill self-spend, not a live `estimate_smart_fee` call. Left as-is — that's a different conversation about whether self-spend refills should use live estimates.

In this case, it should. But it's also fine to leave it as is. Once we have CPFP, we'll remove the `fee` module entirely. At that point, we'll need to estimate the fee rate. So it'll be good to update it right now. Otherwise, this will have to be updated later once that module is removed.

Zk2u:
reworked to layer ontop of the cpfp work and FB

Rajil1213:
I see that the base for this PR has changed. What other PR does this depend on?

crates/bridge-exec/src/fees.rs
```diff
@@ -359,32 +355,47 @@ impl FeeSourceConfig {
...
 // ────────────────────────────────────────────────────────────────────────────
-// helpers
+// minimum fee rates
 // ────────────────────────────────────────────────────────────────────────────
 
-/// Clamps a raw sat/vB rate to a minimum of 1 sat/vB.
+/// Floor every source's reported rate at this many sat/vB. `bitcoind-async-client` represents
+/// fee rates as `u64` sat/vB, so a sub-1 sat/vB estimate would otherwise truncate to 0 and lose
+/// the fee entirely (notably on public signet).
+const MIN_FEE_RATE_SAT_PER_VB: u64 = 1;
```
Rajil1213: Any reason this const is not of type `FeeRate` like the `MIN_WALLET_TX_FEE_RATE`?

crates/bridge-exec/src/fees.rs
```diff
@@ -8,7 +8,7 @@
 //!   or decode failure so a downed mempool explorer never blocks tx publishing.
 //! * [`FixedFeeSource`] — returns a constant rate (tests, manual overrides).
 //!
-//! All sources clamp the returned rate to at least 1 sat/vB (see [`clamp_to_min`]).
+//! All sources clamp the returned rate to at least 1 sat/vB.
```
Rajil1213: This should link to the const instead of saying `1 sat/vb`.

crates/bridge-exec/src/fees.rs
```diff
@@ -371,9 +371,9 @@ impl Default for FeeSourceConfig {
 const MIN_FEE_RATE_SAT_PER_VB: u64 = 1;
 
 /// Minimum rate at which the bridge broadcasts a *wallet-funded* transaction (withdrawal
-/// fulfillment, stake funding). The configured source already clamps to
-/// [`MIN_FEE_RATE_SAT_PER_VB`]; this is the higher bridge-policy floor that keeps these v3 (TRUC)
-/// transactions relayable even when the source reports a lower rate.
+/// fulfillment, stake funding). The configured source already clamps to a 1 sat/vB truncation
+/// guard; this is the higher bridge-policy floor that keeps these v3 (TRUC) transactions
```
Rajil1213: Same here, use the declared const instead of saying `1 sat/vb`.

crates/bridge-exec/src/deposit.rs
Rajil1213: This looks like a bad rebase. This should just return an error (as we might be interested in this error log showing up in our observability stack).

Rajil1213 (dismissed):
Only some nits remain in the new changes. However, there are still some unresolved comments from the previous round. And the following have been flagged by Codex:

- `fee::FEE_RATE` is still being used to refill the claim-funding UTXOs. This should also use the `cfg.fee_source.current()` estimate (clamped as necessary).
- `unstaking_burn` executor also still uses bitcoin core directly instead of the new fee source abstraction.

crates/bridge-exec/src/fees.rs
Rajil1213: PS: there should be a lint you can disable to allow referencing private items in doc comments.

Rajil1213:
Went through this PR again. And some of the reviews from the last rounds have still not been addressed. Here is the list:

- Fee-source API shape: conf_target is still stored on BitcoindFeeSource, policy is still stored on MempoolExplorerFeeSource, and FeeSource::estimate() still takes no target/deadline argument.

- Claim-funding refill still uses fee::FEE_RATE in crates/bridge-exec/src/graph/common.rs:188 instead of `cfg.fee_source.current()`.
- unstaking_burn still calls Bitcoin Core directly via estimate_smart_fee(1) in crates/bridge-exec/src/graph/unstaking_burn.rs:152-155.

I still don't fully understand why the base PR for this one is the fireblocks branch. I'd have assumed that the fee rate policy is a standalone change.

PR #572, opened 2026-05-25
feat(bridge-exec): wire CPFP + aggressive RBF for bridge txs (STR-3439)

Zk2u:
## Description

Wires CPFP fee-bumping for **15 of 17** bridge tx kinds via v3 1P1C packages submitted with `submitpackage`. Aggressive RBF on every new block, mempool eviction, and a configurable timer tick (default 30 s).

Both `CpfpStrategy` variants are wired:
- **AnchorBearing** for keyed-anchor parents: claim, stake, unstaking_intent, counterproof, counterproof_ack, contest, bridge_proof_timeout.
- **ParentTxCombined** for payout-output parents: coop_payout, uncontested_payout, contested_payout, unstaking, counterproof_nack (statically-known vout) + WFT, stake funding, slash (script-discovery via `InferGeneralPayout`).

Only **deposit** and **bridge_proof** remain unbumpable — both genuinely lack any operator-owned output and would need tx-graph changes to gain a CPFP hook.

### Type of Change

- [x] New feature/Enhancement

## Notes to Reviewers

**Architecture seam.** `btc-tracker` defines three abstract traits (`CpfpWallet`, `CpfpFeeSource`, `CpfpPackageSubmitter`); `bridge-exec` wires concrete adapters over `OperatorWallet` + bitcoind. `CpfpContext` + `perform_bump` are the single source of truth for one bump attempt. The trait split keeps `btc-tracker` at the bottom of the dependency graph and makes the upcoming Fireblocks backend a pure trait-impl exercise.

**Backend-agnostic signing contract.** `CpfpWallet::build_cpfp_child` may return a fully-unsigned, partially-signed, or fully-signed PSBT. `perform_bump` inspects each input's `final_script_witness` and only signs the unsigned ones (anchor via `anchor_input_signer` keyed to the operator's musig2 key; everything else via `wallet_input_signer` keyed to the general key). This means:
- `NativeGeneralWallet` (descriptor-only) returns unsigned PSBT → bump loop signs everything.
- Upcoming Fireblocks backend (create-and-sign) will return partial PSBT → bump loop only signs the foreign anchor input.

**Notable design choices**, full rationale in commit body:
- `CpfpKind` enum lets each `publish_signed_transaction` call site declare CPFP intent explicitly.
- `parent_fee` is caller-supplied (presigned txs use `FEE_RATE × vsize`; wallet-funded txs use `Psbt::fee()` or `exact_fee_from_prevouts`).
- `CachedFeeSource` wraps `estimatesmartfee` with an atomic background refresh so the bump loop reads constant-time.
- `BumpReason` enum: eager bumps skip on `target ≤ last`; reactive bumps always rebuild to handle silent child eviction.
- Block/tick arms spawn `bump_all_entries` with a `JoinHandle` guard to prevent fan-out pile-up; entries-mutex held briefly for snapshot/writeback only.
- `_covenant_keys_field_audit` in `crates/common/src/params.rs` — compile-time guard so splitting watchtower keys from musig2 keys breaks the build until the CPFP anchor-inference path is taught to dispatch per-anchor.

## Checklist

## Related Issues

STR-3439

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11663,6 +11663,8 @@ dependencies = [
  "operator-wallet",
  "secret-service-client",
  "secret-service-proto",
+ "serde",
+ "serde_json",
  "serial_test",
  "ssz",
  "strata-asm-proto-bridge",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11663,6 +11663,7 @@ dependencies = [
  "operator-wallet",
  "secret-service-client",
  "secret-service-proto",
+ "serial_test",
  "ssz",
  "strata-asm-proto-bridge",
  "strata-asm-proto-bridge-txs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -177,6 +177,9 @@ bitcoin-script = { git = "https://github.com/BitVM/rust-bitcoin-script" }
 bitcoincore-rpc = "0.19.0"
 bitcoind-async-client = { version = "0.14.0", default-features = false, features = [
   "30_2",
+  # `raw_rpc` exposes `call_raw`, which the CPFP bump loop needs for
+  # `gettxspendingprevout`. The typed surface of the client does not cover it.
+  "raw_rpc",
 ] }
 borsh = { version = "1.6.1", features = ["derive"] }
 cfg-if = "1.0"
--- a/bin/strata-bridge/src/config.rs
+++ b/bin/strata-bridge/src/config.rs
@@ -4,6 +4,17 @@
 //! different operators.
 use std::{fmt, net::SocketAddr, path::PathBuf, time::Duration};
 
+/// Default cadence for the background fee-rate refresh task — see
+/// [`Config::fee_refresh_interval`].
+const fn default_fee_refresh_interval() -> Duration {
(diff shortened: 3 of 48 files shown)


Rajil1213:
As the other PR that this one depends on needs changes, will review this one after those have been addressed.

Rajil1213:
This needs a rebase now.

Rajil1213:
@Zk2u this is a pretty long PR and quite complicated at that. Could you first create a design doc for this?

Rajil1213:
@codex review.

barakshani:
> Aggressive RBF on every new block, mempool eviction, and a configurable timer tick (default 30 s).

I got here by "accident" so don't have sufficient understanding of the impl. Why do we need aggressive RBF? What does "aggressive" mean in practice? Does the timer tick mean that every 30s (or whatever the value is set to be) we 're-evaluate'? What actually happens every clock tick?

barakshani (dismissed):
@Zk2u I left some comments on the design doc, as well as one here which is still unasnwered. I am not comfortable with the aggressive approach taken to blindly bump the fees if a tx isn't mined.

Please look into it, so we can get this shipped.

PR #574, opened 2026-05-26
Wire bridge logging and metrics

voidash:
## Description

Wires shared `strata-logging` and `strata-metrics` into `strata-bridge` for STR-3586.

This adds process metrics initialization, Prometheus exporter config for local Docker nodes, and a low-cardinality `strata_bridge_node_info` gauge with static labels for service, mode, network, version, active exporter path, P2P scoring preset, Bitcoin ZMQ status, and FDB TLS status.

### Type of Change

- [x] Documentation update
- [x] Refactor
- [x] New or updated tests
- [x] Dependency Update

## Notes to Reviewers

Pre-PR review found one issue and fixed it: config parsing no longer debug-logs parsed config structs, which could expose values such as RPC passwords.

Dependency boundary is intentional: only `strata-logging` and `strata-metrics` use `strata-common` `v0.1.0-alpha-rc24`; protocol/data crates remain on `v0.1.0-alpha-rc21` to avoid duplicate type conflicts with existing pinned dependencies.

Created and pushed `strata-common` tag `v0.1.0-alpha-rc24` from `main` commit `f93a3b7c4f2381dca87354aa59ebadbc4bc40755`.

Verification run:

- `cargo fmt --all --check`
- `cargo check --workspace --all-targets`
- `cargo test -p strata-bridge`
- `git diff --check`

AI assistance notice: Prepared with Codex.

## Checklist

## Related Issues

STR-3586: https://alpenlabs.atlassian.net/browse/STR-3586


Rajil1213:
@voidash CI needs fixing. You should just be able to run `just lint-fix-ws`.

bin/dev-cli/src/main.rs
```diff
@@ -5,24 +5,40 @@ mod handlers;
...
 use crate::handlers::{bridge_in, checkpoint, claim, contest};
 
 mod cli;
 
 #[tokio::main]
 async fn main() -> Result<(), Error> {
-    logging::init(LoggerConfig::new("dev-cli".to_string()));
+    let service_label = logging::get_service_label_from_env();
+    let otlp_url = logging::get_otlp_url_from_env();
+    init_logging_from_config(LoggingInitConfig {
```
Rajil1213: This is fine but also unnecessary since the `dev-cli` is only used in tests and local runs.

bin/strata-bridge/src/main.rs
```diff
@@ -75,6 +65,50 @@ fn main() {
...
+    emit_node_info_metric(&NodeInfoLabels {
+        service: &service_name,
+        mode: &mode_label,
+        network: &network_label,
+        version: env!("CARGO_PKG_VERSION"),
+        metrics_exporter,
+        p2p_scoring_preset,
+        btc_zmq,
+        fdb_tls,
+    });
```
Rajil1213: Can you move all this to a different module? Let's keep the `main` function succinct so that it is easy to follow.

bin/dev-cli/src/main.rs
Rajil1213: On the other hand, the `secret-service` does need this.

Rajil1213 (changes requested):
Looks good overall. The main change required is to add the observability infra on the secret service as well. This may or may not end up being used in prod but it'll be nice to have that option even if we set the OTLP config to `None`.

Another note is to use rebases instead of merges to resolve conflicts so that the commit history remains linear. In fact, I highly recommend dropping the last commit in this PR which has no code changes.

Rajil1213 (approved):
This looks good to go now.

Rajil1213:
Bypassing the merge queue as codeberg is facing [gateway timeout issues](https://codeberg.org/Codeberg/Community/issues/2596) while fetching `flexitest`.

PR #576, opened 2026-05-27
Bridge Counterproof Guest Program & prover executor integration

MdTeach:
## Description
* Implement Counterproof Guest Program
* Integrate SP1 Host for Counterproof Guest Program
* CI builds the Counterproof Guest Program

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers
Most of the diff is due to the `Cargo.lock`


crates/proofs/bridge-counterproof/src/host/backend.rs
```diff
@@ -0,0 +1,83 @@
+//! Bridge-counterproof host construction.
+//!
+//! Bundles the runtime host enum and the constructor that turns a
+//! [`ProofBackendConfig`] into a ready-to-use [`BridgeCounterproofHost`]. Naming and shape
```
Rajil1213: Looks like this doc is incomplete.

crates/algebra/Cargo.toml
```diff
@@ -7,9 +7,15 @@ edition = "2021"
...
+cfg-if.workspace = true
+futures = { workspace = true, optional = true }
+proptest = { workspace = true, optional = true }
+tokio = { workspace = true, optional = true }
 
 [dev-dependencies]
 tracing.workspace = true
+
+[features]
+default = []
```
Rajil1213: You should make both `async` and `proptest` the default so as to avoid downstream changes. Then in the connectors crate, you can explicitly opt-out.

Rajil1213 (commented):
Manual review didn't turn up anything blocking but I'd like to see a review from @prajwolrg as well.

crates/proofs/bridge-counterproof/src/host/backend.rs
MdTeach: its was re-refactored inn [9dea617](https://github.com/alpenlabs/strata-bridge/pull/576/commits/9dea61790c660464d296f95c678ef767db2a133a)

crates/algebra/Cargo.toml
MdTeach: Amended in [4562491](https://github.com/alpenlabs/strata-bridge/pull/576/commits/4562491022abdc26d247e20d1e060ee8667be0e2)

crates/connectors/Cargo.toml
```diff
@@ -10,7 +10,7 @@ workspace = true
 test_utils = ["dep:bitcoind-async-client", "dep:corepc-node"]
 
 [dependencies]
-strata-bridge-primitives.workspace = true
+strata-bridge-primitives = { path = "../primitives", default-features = false }
```
prajwolrg, line 13: I'm curious on why this is done.

crates/proofs/common/src/host/predicate.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/// [`sp1_groth16_predicate_key`] for a given `elf`, rendered as `Sp1Groth16:<hex>` — the
+/// form the params TOML parser expects.
+pub fn sp1_groth16_predicate_string(elf: &[u8]) -> Result<String> {
+    let vkey_hash = sp1_program_vkey_hash(elf)?;
+    Ok(sp1_groth16_predicate_string_from_key(
+        &sp1_groth16_predicate_key(vkey_hash)?,
+    ))
+}
```
prajwolrg, line 40: nit: serializer of the `PredicateKey` does have similar function already. See https://github.com/alpenlabs/strata-common/blob/main/crates/predicate/src/serde.rs#L40.

prajwolrg (approved):
LGTM! Just a question and an improvement nit.

crates/connectors/Cargo.toml
MdTeach: The reason is that workspace inheritance doesn't reliably honor `default-features = false`.

If the workspace dep is declared without default-features (which is the case for strata-bridge-primitives in our root Cargo.toml), and a member writes:
`strata-bridge-primitives = { workspace = true, default-features = false }`

Cargo emits a warning and silently keeps the defaults on:
```bash
warning: `default-features` is ignored for strata-bridge-primitives, since
`default-features` was not specified for `workspace.dependencies.strata-bridge-primitives`,
this could become a hard error in the future
```

crates/proofs/common/src/host/predicate.rs
MdTeach: Refactored in #579's [16d4927](https://github.com/alpenlabs/strata-bridge/pull/579/commits/16d49279848d533e242dd8db45b2797f4f3b46c6)

PR #580, opened 2026-05-29
feat(operator-wallet): Fireblocks GeneralWallet backend

Zk2u:
## Description

Adds a Fireblocks-custodied general wallet as an alternative to the native BDK backend, selected at startup. When the optional `[operator_wallet.fireblocks]` config section is absent, the native backend is used (default, unchanged behavior). The reserved wallet stays native regardless.

**RAW signing, not TRANSFER.** Fireblocks doesn't build the bitcoin tx for us: the bridge constructs the unsigned v3/TRUC transaction itself (input selection, change, sighashes), sends the sighashes to Fireblocks for ECDSA signing, and assembles the P2WPKH witnesses. This is the only flow that lets us inject a foreign Taproot anchor input, force v3, control exact outputs, and get the PSBT back to package with the parent for CPFP.

All bridge protocol keys (anchors, presigned-tx keys) remain in secret-service. Fireblocks only ever contributes a funding input + optional change to cover fees / top up internal pools, and receives earnings at the vault address.

Key pieces:
- `GeneralWallet` trait gains `payout_descriptor()` and `sign_owned_inputs()`; backend erasure via an `AnyOperatorWallet` enum keeps the executor pipeline non-generic (Fireblocks variant gated behind a `fireblocks` cargo feature).
- CPFP general-payout detection is backend-aware (matches the active backend's receive script: native BIP86 P2TR vs. Fireblocks vault P2WPKH), so CPFP works for both backends.
- All funding/signing call sites honor the signing contract: use the backend-produced witness if present, otherwise secret-service-sign.

Security / robustness:
- Per-request RS256 JWT (`uri`/`nonce`/`iat`/`exp`/`bodyHash`) + `X-API-Key`; API secret read from a file path and redacted in `Debug` (never in the config body or logs).
- RAW-sign responses are matched by echoed sighash content (not order, with a duplicate-sighash guard); each assembled witness is verified to (a) be controlled by the prevout's witness program and (b) verify against the sighash before it's trusted.
- Exact `asset_id`/network validation, P2WPKH deposit-address enforcement, `base_url` trailing-slash normalization, and `deny_unknown_fields` on the wallet config so a typo can't silently downgrade a Fireblocks operator to native.

### Type of Change

## Notes to Reviewers

- **Stacked on #572** — base branch is `azz/str-3439-cpfp-rbf`; review only the top commits.
- Neither maintainer has a Fireblocks account, so the backend is validated against the Fireblocks swagger spec via unit tests (JWT/auth, DTO serde, witness assembly incl. a real BIP-143 sighash round-trip and high-S normalization, and the funding/CPFP selection + dust-boundary math). The live RAW-sign submit→poll path is not covered by an integration test.
- The operator-facing setup guide is intentionally deferred to a follow-up.

## Checklist

## Related Issues

Part of the operator-wallet / Fireblocks epic (STR-3434).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2451,7 +2451,7 @@ dependencies = [
  "bitflags 2.10.0",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -3113,7 +3113,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4076,6 +4076,7 @@ dependencies = [
  "ff",
  "generic-array 0.14.7",
  "group",
+ "hkdf",
  "pem-rfc7468",
  "pkcs8",
  "rand_core 0.6.4",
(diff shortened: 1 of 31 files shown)


Rajil1213:
Why does this PR depend on the CPFP branch? External wallet and CPFP should be orthogonal features (we should be able to fee bump with the native general wallet too).

PR #604, opened 2026-06-04
refactor(docker): fill asm-params in a separate compose step

MdTeach:
## Description
The asm-runner container no longer fills its own params at startup — a small one-shot asm-params-init service handles that against bitcoind first, then asm-runner just reads the file. This also lets us drop Python from the runtime image and the params.toml mount.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update


docker/asm-runner/Dockerfile
```diff
@@ -12,20 +12,14 @@ RUN --mount=type=cache,target=/root/.cargo/git \
     if [ -z "$ASM_REV" ]; then \
         echo "ERROR: failed to extract asm rev from Cargo.toml" >&2; exit 1; \
     fi && \
-    cargo install --git https://github.com/alpenlabs/asm --rev "$ASM_REV" --root /app strata-asm-runner ${ASM_FEATURES}
+    cargo install --locked --git https://github.com/alpenlabs/asm --rev "$ASM_REV" --root /app strata-asm-runner ${ASM_FEATURES}
```
Rajil1213, line 15: This also needs to be done in the functional tests when installing the runner (due to the same compilation issue that this fixes).

docker/README.md
```diff
@@ -148,7 +148,9 @@ done
 
 Use that output to populate `docker/vol/strata-bridge-{1,2,3}/params.toml`.
 
-The ASM runner uses the same operator set by reading `docker/vol/strata-bridge-3/params.toml` at startup.
+The ASM runner's operator set lives in `docker/vol/asm-runner/asm-params-sample.json` and must
+be kept consistent with the bridge `params.toml` by hand — the runner no longer reads
```
Rajil1213: There's no need for this historical context. A reader of this doc will see the current set of instructions oblivious to how things used to be.

Rajil1213 (changes requested):
Just a nit and one more unrelated change just so that we stay consistent.

docker/README.md
MdTeach: Amended in [f7a89a6](https://github.com/alpenlabs/strata-bridge/pull/604/commits/f7a89a6f73282985eb71f20de6a85cf2d421ba95)

docker/asm-runner/Dockerfile
MdTeach: Amended in [f7a89a6](https://github.com/alpenlabs/strata-bridge/pull/604/commits/f7a89a6f73282985eb71f20de6a85cf2d421ba95)

Rajil1213 (approved):

PR #605, opened 2026-06-04
refactor: pass operator table in duty descr instead of output handles

Rajil1213:
## Description

This PR removes the `operator_table` from `OutputHandles` and updates the affected duty descriptions to instead pass in the operator table from the snapshot in their `Context`. This prepares the codebase for supporting operator set changes where the operator table is scoped to the state machines themselves and not static.

Codex was used to implement these changes.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

This does not depend on any other PRs on operator set changes.

## Checklist

## Related Issues

Closes STR-3669


crates/bridge-exec/src/graph/counterproof.rs
```diff
@@ -257,3 +254,61 @@ async fn fetch_counterproof_input(
...
+            .map(|idx| {
+                let secret =
+                    SecretKey::from_slice(&[*idx as u8 + 1; 32]).expect("test key is valid");
+                let btc_key = PublicKey::from_secret_key(&secp, &secret);
+                (*idx, P2POperatorPubKey::from(vec![*idx as u8; 32]), btc_key)
+            })
+            .collect();
+
+        OperatorTable::new(entries, OperatorTable::select_idx(pov_idx))
+            .expect("test operator table is valid")
```
MdTeach: I think we already have this in test-utils. Lets reuse or extend that.

MdTeach (changes requested):
Overall looks good.
Just a nit.

crates/bridge-exec/src/graph/counterproof.rs
Rajil1213: > I think we already have this in test-utils. Lets reuse or extend that.

The one we have is slightly different than required for this particular use case. But the test does not really require that either. So, I've updated the tests to just use the construction from test-utils. Amended in https://github.com/alpenlabs/strata-bridge/commit/cbf27e87f5d40323245ff49b5b3b658cdaaa4c2c.

MdTeach (approved):

PR #621, opened 2026-06-08
refactor(bridge): adopt GameIndex on mosaic client API and align counterproof output ordering

MdTeach:
## Description
* Mosaic client API — replaces `DepositIdx` with a new `GameIndex` newtype (`NonZero<u32>`, equal to `DepositIdx + 1`) on every `MosaicClientApi` method.
* Reorders the fields of `CounterproofOutput` so `operator_pubkey` is committed before `game_idx`, both in the struct definition and in the `commit_ssz` call. This keeps proof public params and ckt public params processing in sync

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

### Note to reviewers
Created [STR-3754](https://alpenlabs.atlassian.net/browse/STR-3754)  to better sync game-index and deposit index

[STR-3754]: https://alpenlabs.atlassian.net/browse/STR-3754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #622, opened 2026-06-08
fix: make funding outpoint reservations idempotent under concurrent workload

Rajil1213:
## Description

This PR fixes an issue where the same graph ends up getting allocated two different funding outpoints. This happened because the duty to generate the graph data got spawned concurrently (one from the block event of observing the DRT and another from a nag from another operator). This caused the mosaic tableset to be initialized with two different sighash value sets, resulting in adaptor verification failing. A similar issue also exists in the fulfillment executor.

The fix implemented in PR updates the database API to never overwrite an existing funding oupoint, and then re-leasing any stale/duplicate leases from the wallet. There is also a regression test (in the second commit) that should fail with `DepositAborted` logs. If run from the tip of this branch, the test should pass.

The regression test is essentially a stress-test and it revealed another issue where the bridge node would panic if it received the `AdaptorsVerified` event too late. So that has been fixed as well. The error has been downgraded from `InvalidEvent` to `Rejected`.

The permissive scoring system in p2p has also been updated to make it as permissive as possible so that peers are not muted. This was primarily needed for the regression test to work (as we're bombarding peers with nags) but even long term, setting the max possible value is closer to what its name represents and so that has been kept around in this PR.

Codex was used to plan and implement these changes.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Notes to Reviewers

Functional tests are failing to install `asm-runner` (should be fixed after #621 is merged and this branch is rebased).

## Checklist

## Related Issues

Closes STR-3749


crates/p2p-service/src/bootstrap.rs
```diff
@@ -91,11 +91,11 @@ fn create_permissive_peer_score_params(topic_name: &str) -> PeerScoreParams {
...
         // Allow peers with scores down to -1000 to participate in gossip
-        gossip_threshold: -1000.0,
+        gossip_threshold: f64::MIN,
         // Allow peers with scores down to -1000 to receive published messages
-        publish_threshold: -1000.0,
+        publish_threshold: f64::MIN,
         // Only graylist peers with extremely negative scores
-        graylist_threshold: -10000.0,
+        graylist_threshold: f64::MIN,
         // Disable opportunistic grafting threshold
```
MdTeach: 🤖  Nit: 

> After switching to f64::MIN, the comments still read "Allow peers with scores down to -1000"

MdTeach (dismissed):
Overall looks good.
Just a nit.

crates/p2p-service/src/bootstrap.rs
Rajil1213: > 🤖  Nit: 
> 
> > After switching to f64::MIN, the comments still read "Allow peers with scores down to -1000"

Nice catch. Amended in https://github.com/alpenlabs/strata-bridge/commit/90efda9de058af722c57e67093e61b5e5e7a685b.

MdTeach (approved):

PR #631, opened 2026-06-12
fix(gsm): do not publish slash from AllNackd state

Rajil1213:
## Description

In the canonical implementation of the `strata-bridge`, we do not want to slash whenever possible regardless of operator fault. This means we must not have this implementation publish a slash transaciton if the payout timelock has expired in the `AllNackd` state as this state is only reached if the operator is _not_ at fault.

AI was not used in this PR.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

Closes STR-3565

### DIFF
--- a/crates/bridge-sm/src/graph/tests/notify_new_block.rs
+++ b/crates/bridge-sm/src/graph/tests/notify_new_block.rs
@@ -1267,9 +1267,9 @@ mod tests {
         );
     }
 
-    /// Non-owner publishes slash after payout timelock expires in AllNackd state.
+    /// Non-owner does not publish slash even after payout timelock expires in AllNackd state.
     #[test]
-    fn all_nackd_nonpov_slash() {
+    fn all_nackd_nonpov_no_slash() {
         let cfg = test_graph_sm_cfg();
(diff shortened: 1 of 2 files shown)


crates/bridge-sm/src/graph/tests/notify_new_block.rs
```diff
@@ -1267,9 +1267,9 @@ mod tests {
         );
     }
 
-    /// Non-owner publishes slash after payout timelock expires in AllNackd state.
+    /// Non-owner does not publish slash even after payout timelock expires in AllNackd state.
```
barakshani, line 1270: this is misleading, you can only publish after the timelock expires. I'd delete "even" and replace "does not" with "should not"

barakshani (approved):

crates/bridge-sm/src/graph/tests/notify_new_block.rs
Rajil1213: This is in the test. It defines behavior of the function under test rather than the protocol. So the wording reflects that.

PR #632, opened 2026-06-12
fix(dsm): tighten reassignment checks

Rajil1213:
## Description

This PR hardens the checks when a withdrawal reassignment event is received in the DSM. This keeps the checks in alignment with those in the GSM.

Codex was used to implement these changes.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Checklist

## Related Issues

Closes STR-3017

### DIFF
--- a/crates/bridge-sm/src/deposit/tests/payout/process_assignment.rs
+++ b/crates/bridge-sm/src/deposit/tests/payout/process_assignment.rs
@@ -78,16 +78,13 @@ mod tests {
     /// and POV operator is the new assignee (should emit FulfillWithdrawalRequest duty).
     #[test]
     fn test_reassignment_to_pov() {
-        let old_desc = random_p2tr_desc();
-        let new_desc = random_p2tr_desc();
-
-        assert_ne!(old_desc, new_desc, "must be diff");
+        let desc = random_p2tr_desc();
 
         let state = DepositState::Assigned {
             last_block_height: INITIAL_BLOCK_HEIGHT,
             assignee: TEST_NONPOV_IDX,
             deadline: LATER_BLOCK_HEIGHT,
-            recipient_desc: old_desc,
+            recipient_desc: desc.clone(),
         };
 
         test_deposit_transition(DepositTransition {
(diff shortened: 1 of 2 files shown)


crates/bridge-sm/src/deposit/tests/payout/process_assignment.rs
```diff
@@ -75,38 +75,35 @@ mod tests {
     }
 
     /// tests correct re-assignment from Assigned state when Assignment event is received
-    /// and POV operator is the new assignee (should emit FulfillWithdrawalRequest duty).
+    /// and POV operator is the reassigned assignee (should emit FulfillWithdrawalRequest duty).
```
barakshani: "new" sounds better to me.

crates/bridge-sm/src/deposit/transitions/payout.rs
```diff
@@ -31,31 +31,33 @@ impl DepositSM {
...
+                if *recipient_desc != assignment.recipient_desc {
+                    return Err(DSMError::rejected(
+                        self.state().clone(),
+                        assignment.into(),
+                        "recipient descriptor cannot be changed for an existing assignment",
+                    ));
                 }
+
+                // Assignment deadline must not be smaller than the existing deadline.
+                if assignment.deadline < *deadline {
```
barakshani, line 52: maybe we should not do it here, but shouldn't we check that `deadline` (the current one, before reassignment) is smaller (or equal) than `last_block_height`? similar to "Assignment deadline must not be smaller than the existing deadline", we should not expect a new assignment if the current deadline has not expired.

crates/bridge-sm/src/deposit/transitions/payout.rs
```diff
@@ -65,14 +67,37 @@ impl DepositSM {
...
+
+        self.state = DepositState::Assigned {
+            last_block_height,
+            assignee: assignment.assignee,
+            deadline: assignment.deadline,
+            recipient_desc: assignment.recipient_desc.clone(),
+        };
+        // Dispatch the duty to fulfill the assignment if the assignee is the pov operator,
+        // otherwise no duties or signals need to be dispatched.
+        if self.context.operator_table().pov_idx() == assignment.assignee {
```
barakshani, line 90: I guess it's fine, but let me double check: shouldn't we check deadline w.r.t. the current bitcoin block?

crates/bridge-sm/src/deposit/transitions/payout.rs
Rajil1213: You're right that it's a bit weird to include this check and not the other checks that you've mentioned. But ultimately, it is the ASM that decides when an assignment must be fulfilled. This condition (included in this PR) is just a sanity check.

crates/bridge-sm/src/deposit/transitions/payout.rs
Rajil1213: That's done by the fulfillment executor.

crates/bridge-sm/src/deposit/tests/payout/process_assignment.rs
Rajil1213: > "new" sounds better to me.

Amended in https://github.com/alpenlabs/strata-bridge/commit/a7c7ba380f123761470e90cc809e58fc457d89ae.

barakshani (approved):

PR #654, opened 2026-06-22
chore(deps): backport asm v0.1-alpha.14 (#638) onto deployed 6ca08588 for staging-v2

prajwolrg:
## Description

Backports the asm dependency bump from #638 onto `6ca08588` — the strata-bridge commit currently deployed on **staging-v2** — so the new asm generation can be exercised on staging in isolation, without pulling in the 111 unrelated commits `main` has accrued since that deploy.

`6ca08588` predates #638's own base by a full dependency generation (strata-common `rc21` / asm `dc6432ff` vs #638's `rc23` start), so #638's per-commit diffs do not apply — this is a backport: the whole coordinated set is re-pinned to #638's final targets and the one call site the new asm API breaks at this base is adapted by hand.

Net change vs the deployed commit: asm `dc6432ff → 31d88a4` (alpha.14 + asm PR #155 phantom-reorg fix), strata-common `rc21 → rc26`, moho `alpha.6 → alpha.9`, zkaleido `beta.2 → beta.5`, ssz `v0.15.0 → v0.16.0`, bitcoin-bosd `v0.10.0 → v0.11.0`, alpen `2c6ae209 → 8589f8e`. `strata-predicate`'s `all` feature became `all-verifiers` and dropped serde/borsh from defaults, so those are requested explicitly. asm merged `AssignmentsApiClient` into `AsmStateApiClient`, ported in `crates/asm-events/src/client.rs` (#638 never touched this file — that code did not exist at its newer base; #638's `statements.rs` change is omitted for the same reason).

`Cargo.lock` is held at the deployed versions except the asm generation above and the SP1 stack, which moves only as far as zkaleido `beta.5`'s `blake3 >= 1.8.5` requires (`sp1-verifier 6.2.0 → 6.3.0`); `jemalloc_pprof` and every other transitive pin are left exactly as staging runs them today.

This PR targets a base branch pinned at the deployed commit (`base/staging-v2-asm-backport`), not `main`, so the diff is exactly the backport and CI exercises it against the deployed bridge. It is **not** intended to merge to `main`.

### Type of Change

- [x] Dependency Update

## Notes to Reviewers

`cargo check --workspace --all-targets` passes locally. The point of this PR is to run the **unit and functional suites** (and `guest-build`) against the new asm on the deployed bridge before the image is promoted to staging-v2 — i.e. CI is the validation here. One deviation from #638 worth watching: a resolve from the deployed lockfile picks `sp1-verifier 6.3.0` rather than #638's `6.2.4`; the host compiles cleanly, so the guest-build / SP1 proving path is the thing to confirm green.

## Checklist

- [x] I have disclosed my use of AI in the body of this PR.

## Related Issues

Backport of #638. Resulting image tag `9ba7c17b` is what the deployments repo would point staging-v2's `strataBridgeClient` and `asmRunner` at.

---

_AI assistance: the dependency re-pinning, conflict resolution, lockfile regeneration, and this description were prepared with Claude Code._

### DIFF
--- a/.github/workflows/functional.yml
+++ b/.github/workflows/functional.yml
@@ -222,7 +222,11 @@ jobs:
         env:
           MOSAIC_REV: ${{ steps.bin-versions.outputs.mosaic-rev }}
         run: |
-          cargo install --git https://github.com/alpenlabs/mosaic --rev "$MOSAIC_REV" --features=reduced-circuits --root "$GITHUB_WORKSPACE/target/mosaic" mosaic
+          # `--locked` forces use of mosaic's committed Cargo.lock so its transitive
+          # ckt-gobble dependency (declared without a rev, so otherwise floating to
+          # ckt `main`) is pinned to the commit mosaic was built against. Without it
+          # a moving ckt `main` yields a binary incompatible with the bridge.
+          cargo install --locked --git https://github.com/alpenlabs/mosaic --rev "$MOSAIC_REV" --features=reduced-circuits --root "$GITHUB_WORKSPACE/target/mosaic" mosaic
 
       - name: Add mosaic binary to PATH
         run: |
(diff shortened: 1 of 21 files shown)


Rajil1213:
I believe this PR can be closed now, right?

prajwolrg:
This doesn't need to be merged anywhere.

PR #702, opened 2026-07-09
feat(ci): env-scoped artifact versions + sha256 sidecars

MdTeach:
## Description
Two enhancements to the Publish Guests + g16 Circuit pipeline:

1. **sha256 sidecars** — every uploaded ELF/vkey/circuit now ships a `<name>.sha256` next to it in `sha256sum c` format, so downloads can be verified. 
2. **env-scoped S3 versions** — every version segment is now prefixed with a deployment environment (`dev`/`staging`/`prod`), keeping each env's artifacts in distinct keys: `bridge/<env>-<genesis>-<sha8>/`, `asm-runner/<env>-<tag>-<sha8>/`, `<env>-<genesis>-<sha8>/g16.v5c`.

The env is chosen once at dispatch, recorded in the guest-build manifest, and read back by both the upload and circuit-gen steps — so the circuit can never be keyed to a different env than the bridge tree it was generated from.

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

### DIFF
--- a/.github/scripts/ci_common.py
+++ b/.github/scripts/ci_common.py
@@ -47,9 +47,22 @@ def write_sha256_sidecar(digest: str, s3_name: str, dest_dir: Path) -> Path:
     return sidecar
 
 
-# `<genesis>-<sha8>` / `<tag>-<sha8>` — the components are numeric/hex/tag chars,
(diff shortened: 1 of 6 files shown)


madan-oss (approved):
LGTM

PR #703, opened 2026-07-09
feat: assert SP1 guest ELF matches params predicate at startup

MdTeach:
## Description
In SP1 mode the guest ELF that *generates* proofs and the params predicate that *verifies* them are configured independently, so a mismatched ELF silently breaks verification. This adds a startup check that the predicate derived from each loaded guest ELF (bridge-proof + counterproof) matches the one in params, aborting startup on mismatch. A new `dev` flag (default `false`) bypasses it.

### DIFF
--- a/bin/strata-bridge/src/config.rs
+++ b/bin/strata-bridge/src/config.rs
@@ -83,6 +83,10 @@ pub(crate) struct Config {
     /// Backend that produces bridge counterproofs.
     pub counterproof: CounterproofBackendConfig,
 
+    /// Developer/experimentation mode. When `true`, the startup consistency checks are skipped.
+    #[serde(default)]
+    pub dev: bool,
+
     /// Configuration for process-level metrics exporters.
     #[serde(default)]
     pub metrics: MetricsConfig,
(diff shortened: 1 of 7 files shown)


bin/strata-bridge/src/config.rs
```diff
@@ -83,6 +83,10 @@ pub(crate) struct Config {
     /// Backend that produces bridge counterproofs.
     pub counterproof: CounterproofBackendConfig,
 
+    /// Developer/experimentation mode. When `true`, the startup consistency checks are skipped.
+    #[serde(default)]
+    pub dev: bool,
```
Rajil1213, line 88: The default should be `false`. Is that the case with this setting?

Rajil1213 (dismissed):
Just one question but otherwise looks good.

bin/strata-bridge/src/config.rs
MdTeach: yes, dev is false unless explicitly configured as true.

Rajil1213 (approved):

PR #734, opened 2026-07-21
feat(orchestrator): add event-pipeline observability (STR-3592)

voidash:
## Summary

Implements the runtime portion of STR-3592: bounded metrics and structured spans across the orchestrator event pipeline, plus the failure-path fixes needed to make those signals trustworthy.

This PR is intentionally limited to the orchestrator. Prometheus/Grafana/Loki configuration, dashboards, alert thresholds, retention, authentication, and scrape wiring belong in the deployment repository and are not included here.

## What changed

- Adds `bridge_event`, `bridge_sm_transition`, and `bridge_duty_execution` spans around the pipeline's causal stages. Transition spans are `INFO` level so they exist under the default filter; periodic transition logs remain `DEBUG` to avoid log noise.
- Adds eight `strata_bridge_*` metric families:
  - end-to-end pipeline event duration and routing outcomes
  - state-machine transition outcomes and duration
  - duty outcomes, in-flight count, and duration
  - persistence duration and outcome
- Keeps metric labels bounded to explicit classifiers such as event kind, state-machine kind/state, result, error class, and duty kind. IDs, transaction data, peer keys, payloads, and raw errors remain in spans or structured logs rather than metric labels.
- Catches detached duty-future panics and logs executor failures with bounded error classes. Duty span context uses the existing redacted `Display` implementations rather than potentially sensitive `Debug` output.
- Fails an entire persistence batch when a tracked state machine is missing instead of silently writing a partial batch.
- Persists every causally linked state-machine batch before dispatching any duty from the event.
- Removes duplicate/no-op transition logging while retaining those outcomes in metrics.

## Operational impact

Operators can correlate a top-level event with its state-machine transitions and detached duties, measure failure rates and latency, and distinguish bounded failure classes without creating identifier-driven Prometheus cardinality.

The `metrics` histogram instruments currently follow the existing recorder configuration. In the current exporter setup they render as summaries unless bucket configuration is added centrally, so consumers must not assume `_bucket` series exist.

## Explicitly out of scope

- Production metrics endpoint/scrape configuration, dashboards, alerts, retention, and authentication (deployment repository).
- Durable per-deposit lifecycle records and timeline RPC (STR-3598).
- Cross-service trace/deposit-ID propagation (STR-1975).
- Durable duty scheduling, retries, and backpressure (STR-2698). The in-flight gauge and outcome metrics make the current detached-execution behavior observable; they do not make it durable.

## Validation

- `cargo fmt --all -- --check`
- `cargo check -p strata-bridge-orchestrator --all-targets`
- `cargo clippy -p strata-bridge-orchestrator --all-targets -- -D warnings`
- `git diff --check main`
- Full linked unit and functional suites run in GitHub CI, which provisions the required FoundationDB client/server.

### DIFF
--- a/crates/orchestrator/src/applicator.rs
+++ b/crates/orchestrator/src/applicator.rs
@@ -12,13 +12,12 @@ use std::collections::VecDeque;
 
 use strata_bridge_primitives::types::{DepositIdx, GraphIdx};
 use strata_bridge_sm::{deposit::machine::DepositSM, graph::machine::GraphSM};
-use tracing::warn;
 
 use crate::{
     errors::PipelineError,
     persister::PersistenceTracker,
     signals_router,
-    sm_registry::{IgnoredEventReason, ProcessOutcome, RegistryInsertError, SMRegistry},
+    sm_registry::{ProcessOutcome, RegistryInsertError, SMRegistry},
     sm_types::{SMEvent, SMId, UnifiedDuty},
 };
 
@@ -160,17 +159,7 @@ impl<'a> Applicator<'a> {
 
                 Ok(())
             }
-            Ok(ProcessOutcome::Ignored { id, event, reason }) => {
-                match reason {
-                    IgnoredEventReason::Duplicate => {
-                        warn!(?id, %event, "duplicate event, skipping");
-                    }
-                    IgnoredEventReason::Rejected(rejected_reason) => {
-                        warn!(?id, %event, %rejected_reason, "event rejected by state machine, skipping");
-                    }
-                }
-                Ok(())
-            }
+            Ok(ProcessOutcome::Ignored { .. }) => Ok(()),
             Err(e) => Err(e.into()),
         }
     }
--- a/crates/orchestrator/src/applicator.rs
+++ b/crates/orchestrator/src/applicator.rs
@@ -130,8 +130,9 @@ impl<'a> Applicator<'a> {
 
     /// Processes a single event through the registry's STF.
     ///
-    /// On success, accumulates duties and enqueues any signal-derived events. Ignored outcomes
-    /// (duplicates, rejections) are non-fatal and logged. Fatal errors are propagated.
+    /// On success, accumulates duties and enqueues any signal-derived events. Ignored outcomes are
+    /// non-fatal: duplicates are debug-logged and rejections are warning-logged by the registry.
+    /// Fatal errors are propagated.
     ///
     /// Persistence tracking follows the state machine's own report: a transition that leaves state
     /// unchanged (e.g. a nag or retry tick) still has its duties accumulated and its signals
--- a/crates/orchestrator/src/applicator.rs
+++ b/crates/orchestrator/src/applicator.rs
@@ -12,12 +12,13 @@ use std::collections::VecDeque;
 
 use strata_bridge_primitives::types::{DepositIdx, GraphIdx};
 use strata_bridge_sm::{deposit::machine::DepositSM, graph::machine::GraphSM};
+use tracing::{debug, warn};
 
 use crate::{
     errors::PipelineError,
     persister::PersistenceTracker,
     signals_router,
-    sm_registry::{ProcessOutcome, RegistryInsertError, SMRegistry},
+    sm_registry::{IgnoredEventReason, ProcessOutcome, RegistryInsertError, SMRegistry},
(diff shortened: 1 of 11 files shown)


crates/orchestrator/src/applicator.rs
```diff
@@ -160,17 +159,7 @@ impl<'a> Applicator<'a> {
...
-                match reason {
-                    IgnoredEventReason::Duplicate => {
-                        warn!(?id, %event, "duplicate event, skipping");
-                    }
-                    IgnoredEventReason::Rejected(rejected_reason) => {
-                        warn!(?id, %event, %rejected_reason, "event rejected by state machine, skipping");
-                    }
-                }
-                Ok(())
-            }
```
barakshani, line 173: why deleting this code?

Rajil1213 (changes requested):
This is an pretty convoluted PR to review -- there is a single commit that makes ALL the changes. I encourage you to split this further out. This PR is also not about observability alone but also makes code changes which are out of scope but acceptable if small. At this point, it's difficult to tell because everything is in a single patch. Please break this commit down into reasonable atomic commits.

Rajil1213 (dismissed):
This looks good to go. 

Much of it will probably need to improve once we test it out in our deployments but the initial ground work looks solid.

crates/orchestrator/src/applicator.rs
voidash: responsibility moved into `SMRegistry::process_event`,  where we can include the SM kind/id,state transition, event, and rejection reason.

crates/orchestrator/src/pipeline.rs
```diff
@@ -94,55 +100,133 @@ impl Pipeline {
...
+                            "classified"
+                        };
+                        observability::record_routing(event_kind, routing_result);
+
+                        if target_count == 0
+                            && !matches!(&event, UnifiedEvent::NagTick | UnifiedEvent::RetryTick)
+                        {
+                            debug!(event_kind, "event did not route to any state machine");
+                        } else if target_count > 0 && classified_count == 0 {
+                            warn!(
```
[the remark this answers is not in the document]
Rajil1213: This feels legit.

crates/orchestrator/src/applicator.rs
Rajil1213: That is the wrong layer of abstraction. The state machine rightfully issues Duplicate/Rejected errors for transitions that are rejected. The orchestrator decides the severity level of these errors.

And either way, moving the code change to another site does not address the original concern from Barak on why this code was necessary in the first place.

Rajil1213 (changes requested):
The mapping from SM-level errors to orchestrator level errors should remain in the orchestrator. And Barak's original question still needs addressing.

crates/orchestrator/src/pipeline.rs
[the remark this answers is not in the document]
Rajil1213: @voidash has this been addressed? Why has this thread been resolved?

crates/orchestrator/src/applicator.rs
voidash: @Rajil1213 Agreed. This is fixed in `0fc123c7`. The earlier deletion was not necessary; it came from an attempted consolidation that put severity policy in the wrong layer. `Applicator` now handles the policy again: `Duplicate` is logged at `DEBUG` and `Rejected` at `WARN`. `SMRegistry` only maps the state-machine errors into `ProcessOutcome` and records transition telemetry. This restores the deleted behavior and addresses Barak’s original concern: the code should not have been removed.

crates/orchestrator/src/pipeline.rs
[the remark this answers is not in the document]
voidash: @Rajil1213 Yes, this is addressed. `a5cfd92c` suppresses the generic warning for `NagRequestExchange`, and `8ed8b042` classifies wrong-recipient and ouroboros filtering as `ExpectedDrop`, so those paths record `expected_drop` instead of `no_classification`. Regression tests cover both peer and ouroboros nags while preserving warnings for genuinely unclassified routed events.

Rajil1213:
This PR needs to be rebased now. And then we can move towards merging.

Rajil1213 (approved):
Commits still need a bit of cleanup but I can live with this. 😅

PR #754, opened 2026-08-03
fix(ci): make RustSec audit fetch reliable

MdTeach:
The RustSec action hid cargo-audit fetch failures behind a JSON parsing error.

This replaces it with pinned cargo-audit 0.22.2, skips registry-index-dependent yanked checks, and updates ruint to 1.20.0 for RUSTSEC-2026-0220.

Validated with actionlint, cargo audit --no-yanked, and cargo check --workspace --all-targets --locked.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -12,8 +12,39 @@ permissions:
   contents: read
 
 jobs:
-  supply-chain:
+  cargo-audit-setup:
+    name: Install cargo-audit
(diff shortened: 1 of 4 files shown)


Rajil1213:
@MdTeach backporting this would be a breaking change/hard fork. Any ideas on how we should handle this change on the release branch? One way would be to not update the guest lockfiles and just update the root lockfile (that CI cares about).

Rajil1213 (approved):

PR #756, opened 2026-08-04
feat: safe harbour activation groundwork

sapinb:
## Description
Second slice of the **Emergency Fund Sweep (Hard Bridge Upgrade)** work ([STR-3937]): the core mechanic that sweeps a deposit UTXO into the frozen safe-harbour address, plus the state-machine groundwork for sweeping **assigned / mid-payout** deposits.

The sweep is mechanically a cooperative payout whose destination is the safe-harbour descriptor, with two deliberate differences: the signing round is **symmetric** (no assignee — every operator publishes nonces and partials, and every operator finalizes and broadcasts the identical deterministic transaction), and the transaction carries an operator-keyed 1-of-N **`MultiAnchor` CPFP anchor** so any operator can fee-bump a stuck sweep later (spending the anchor is a follow-up on top of #572; the base sweep confirms on its own `sweep_fee_rate`).

**This PR is trigger-less by design.** Nothing in production emits `SweepRequested` yet and the
wiring that connects the safe-harbour latch from #733 to the sweep (per-block scan,
withdrawal-duty suppression) lands in follow-up PRs. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency Update

## Checklist

- [ ] I have [disclosed my use of AI](https://github.com/alpenlabs/strata-bridge/blob/main/CONTRIBUTING.md#ai-assistance-notice) in the body of this PR.

## Related Issues

[STR-3937]: https://alpenlabs.atlassian.net/browse/STR-3937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/dev-cli/params.toml
+++ b/bin/dev-cli/params.toml
@@ -31,6 +31,7 @@ magic_bytes = "ALPN"
 deposit_amount = 1_000_000_000
 stake_amount = 100_000_000
 operator_fee = 10_000_000
+sweep_fee_rate = 10
 recovery_delay = 1_008
 contest_timelock = 144
 proof_timelock = 144
--- a/crates/bridge-sm/src/deposit/transitions/payout.rs
+++ b/crates/bridge-sm/src/deposit/transitions/payout.rs
@@ -66,11 +66,19 @@ impl DepositSM {
             | DepositState::PayoutDescriptorReceived { .. }
             | DepositState::PayoutNoncesCollected { .. }
             | DepositState::CooperativePathFailed { .. }
+            // ASM does not gate assignment creation on safe harbour.
+            | DepositState::SweepNoncesPending { .. }
+            | DepositState::SweepNoncesCollected { .. }
             | DepositState::Spent { .. } => {
                 return Err(DSMError::duplicate(self.state.clone(), assignment.into()));
             }
 
-            _ => {
+            // Exhaustive so a new state must be classified rather than inherit the fatal arm.
+            DepositState::Created { .. }
+            | DepositState::GraphGenerated { .. }
+            | DepositState::DepositNoncesCollected { .. }
+            | DepositState::DepositPartialsCollected { .. }
+            | DepositState::Aborted => {
                 return Err(DSMError::invalid_event(
                     self.state.clone(),
                     assignment.into(),
--- a/crates/bridge-sm/src/deposit/transitions/payout.rs
+++ b/crates/bridge-sm/src/deposit/transitions/payout.rs
@@ -569,8 +569,11 @@ impl DepositSM {
         payout_confirmed: &PayoutConfirmedEvent,
     ) -> DSMResult<DSMOutput> {
         let (fulfillment_txid, assignee) = match self.state() {
-            // It must be the sweep transaction in case of a hard upgrade
-            DepositState::Deposited { .. } => (None, None),
+            // It must be the sweep transaction in case of a hard upgrade. The sweep states
+            // observe either their own sweep transaction or a racing payout confirming.
+            DepositState::Deposited { .. }
+            | DepositState::SweepNoncesPending { .. }
+            | DepositState::SweepNoncesCollected { .. } => (None, None),
             // It must be a cooperative payout transaction.
             // The assignee withholds their own partial and broadcasts the payout tx themselves,
             // In this case, we still want other nodes' state machines to transition from
--- a/crates/bridge-sm/src/deposit/transitions/payout.rs
+++ b/crates/bridge-sm/src/deposit/transitions/payout.rs
@@ -563,7 +563,11 @@ impl DepositSM {
     ///   uncontested.
     /// - A contested payout transaction, if the assignee published a claim that was contested but
     ///   not successfully.
-    /// - A sweep transaction in the event of a hard upgrade (migration) of deposited UTXOs.
+    /// - A sweep transaction under safe harbour.
+    ///
+    /// The spend is terminal from every state holding a live deposit UTXO, regardless of which
+    /// of these transactions won; assignee/fulfillment info is preserved where the source state
+    /// carries it, for monitoring only.
     pub(crate) fn process_payout_confirmed(
         &mut self,
         payout_confirmed: &PayoutConfirmedEvent,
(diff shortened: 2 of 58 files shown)


bin/dev-cli/params.toml
```diff
@@ -31,6 +31,7 @@ magic_bytes = "ALPN"
 deposit_amount = 1_000_000_000
 stake_amount = 100_000_000
 operator_fee = 10_000_000
+sweep_fee_rate = 10
```
barakshani, line 34: Is this a floor price, but in practice we adapt the fee based on market rate up to max_fee_rate?

sapinb: This is fixed fee rate to ensure all operators can deterministically create the same transaction. 
Adaptive fee rate will depend on CPFP + RBF. It will be implemented as a followup after #572 is merged.

barakshani: added this to the params doc: https://app.notion.com/p/Bridge-Parameters-2f0901ba000f80f4bf3ef4996283f695

crates/orchestrator/src/events_classifier/offchain.rs
```diff
@@ -335,6 +335,12 @@ pub(crate) fn classify_unsigned_gossip(
                     )
                 })
                 .collect(),
+
+            // TODO(STR-3937): classify into the deposit SM's sweep event once it exists.
```
Rajil1213: This also requires updating the [p2p spec](https://app.notion.com/p/P2P-Schema-2f7901ba000f80ebb496dfe6a8a28339?source=copy_link).

crates/bridge-sm/src/deposit/transitions/sweep.rs
```diff
@@ -0,0 +1,334 @@
...
+                        | DepositState::GraphGenerated { .. }
+                        | DepositState::DepositNoncesCollected { .. }
+                        | DepositState::DepositPartialsCollected { .. }
+                ) {
+                    Err(DSMError::rejected(
+                        self.state().clone(),
+                        sweep_nonce.into(),
+                        "Inapplicable SweepNonce event in pre-deposit state; expected state(s): SweepNoncesPending",
+                    ))
+                } else {
```
Rajil1213, line 223: This will cause a panic if we receive a late `SweepNonce` from any peer. In general, the error policy is to keep the `InvalidEvent` (because it _is_ invalid) but softening it at the orchestrator level (for the time being). At some point in the future, we'll have the state machines take that responsibility.

crates/bridge-sm/src/graph/duties.rs
```diff
@@ -313,6 +313,38 @@ pub enum GraphDuty {
...
+    /// Whether this duty advances the graph owner's claim towards a payout spending the deposit
+    /// UTXO.
+    ///
+    /// Under safe harbour these duties are suppressed at dispatch so no claim advances while
+    /// deposits are being swept. The defensive duties — contest, counterproof, slash, and
+    /// unstaking burn — challenge or punish a claim rather than pursue one and are **never**
+    /// suppressed: they are the only funds defense when a rogue operator stalls the sweep and
+    /// fires its pre-signed claim path. Graph setup duties are also never suppressed, since
+    /// in-flight deposits must finish before they can be swept.
+    pub const fn is_withdrawal_path(&self) -> bool {
```
Rajil1213: Nit: the semantics are a bit different than what the naming of this function would suggest. For example, `PublishContest` maps to `false` when it's also a withdrawal duty, and so is `PublishCounterproofAck` (among others).

Rajil1213: As for in-flight deposit duties, it might make sense to suppress those as well. So a better naming for this function would be `should_suppress_under_sweep_event` (or a shorter version thereof) so that the intent is clearer.

crates/bridge-sm/src/deposit/transitions/payout.rs
```diff
@@ -66,11 +66,19 @@ impl DepositSM {
             | DepositState::PayoutDescriptorReceived { .. }
             | DepositState::PayoutNoncesCollected { .. }
             | DepositState::CooperativePathFailed { .. }
+            // ASM does not gate assignment creation on safe harbour.
+            | DepositState::SweepNoncesPending { .. }
+            | DepositState::SweepNoncesCollected { .. }
```
Rajil1213: It feels like `Rejected` is the right semantics for this error rather than `Duplicate`.

Rajil1213 (changes requested):
Looks good overall. There are a few nits, and a few updates required on the design doc.

Also the commits are beautifully organized. Thanks!

crates/bridge-sm/src/deposit/transitions/sweep.rs
sapinb: InvalidEvent errors from here are softened in a later commit, following existing pattern.

https://github.com/alpenlabs/strata-bridge/pull/756/changes#diff-51d1ff29d78de3b92adc3a49fedb6d2c7bedae0ed386c780901f9f4a045b3e6dR91-R100

crates/bridge-sm/src/deposit/transitions/payout.rs
sapinb: updated.

crates/bridge-sm/src/graph/duties.rs
sapinb: renamed all `is_withdrawal_path ` functions to `should_suppress_under_safe_harbour` 

Inflight deposits are handled separately in the stacked PR #796, so will not make changes for that here.

crates/bridge-sm/src/deposit/tx_classifier.rs
```diff
@@ -69,25 +80,20 @@ impl TxClassifier for DepositSM {
...
+            | DepositState::PayoutNoncesCollected { .. }
             | DepositState::CooperativePathFailed { .. }
-                if is_deposit_spend(self.context().deposit_outpoint, tx) =>
-            {
-                Some(DepositEvent::PayoutConfirmed(PayoutConfirmedEvent {
-                    tx: tx.clone(),
-                }))
-            }
+            | DepositState::SweepNoncesPending { .. }
+            | DepositState::SweepNoncesCollected { .. } => None,
```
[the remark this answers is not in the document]
sapinb: Safe harbour sweep is an emergency measure that ends the bridge's usability. Even if this information is carried over, it will not have any impact on off-protocol reconciliation of potential fronted-but-not-reimbursed withdrawals post sweep.

Rajil1213 (approved):

PR #762, opened 2026-08-06
fix(proofs): anchor moho proof to the trusted genesis state

### LINKED ISSUE #655: Bridge proof doesn't anchor the Moho genesis state a trusted genesis anchor.
### Describe the bug

The current bridge proof takes as hardcoded constants:
1. `moho_vk`
2. `moho_genesis_state`

The above two are bound to vk and are thus our trusted anchors

The bridge proof then has the following inputs:
1. `moho_state` (terminal state that the proof attests to)
2. `moho_proof` (proof along with genesis state)
3. `claim_unlock`
4. `claim_unlock_inclusion_proof`

These are user supplied inputs to the proof and cannot be trusted. 

In the bridge proof, we do the following:
1. verify that the proof's terminal state is the prover supplied `moho_state`. [link to code](https://github.com/alpenlabs/strata-bridge/blob/8c22e56a2794a34d7d87881f84e04162089fcec4/crates/proofs/bridge-proof/src/statements.rs#L80)
2. verify that the `moho_proof`  uses the trusted`moho_vk`. [link to code](https://github.com/alpenlabs/strata-bridge/blob/8c22e56a2794a34d7d87881f84e04162089fcec4/crates/proofs/bridge-proof/src/statements.rs#L86)
3. The claim unlock is included in the MMR of the final state.

It never checks that the trusted `moho_genesis_state` is the same as the initial state that the `moho_proof` user supplied builds on. Hence, the chain is proven sound but the base case can be anything the prover wants. 

A malicious operator can supply arbitrary genesis state where they can add any unlock claims or set an absurdly high PoW and still produce a valid bridge proof. The genesis also contains the `asm_vk` to be used so they can also change that arbitarily.  This allows them to steal funds from the bridge bypassing the entire security measures as the bridge proof is valid and thus cannot be refuted.

### Steps to reproduce

Not relevant.

### Platform(s)

_No response_

### What is the sha256 checksum of the full node you are running?

_No response_

### If you've built the full node from source, provide the full command you used

_No response_

### Code of Conduct

- [x] I agree to follow the Code of Conduct

MdTeach:
## Description

When verifying the recursive Moho proof, the bridge proof compared only the genesis `StateReference` and ignored the genesis state commitment, so the proof could be anchored on any genesis state. `verify_moho_proof` now takes the full `StateRefAttestation` and compares it whole, binding the proof to the genesis state carried by the verifying key. The counterproof's heavier-chain path is updated the same way.

Includes regression tests in both proof crates that keep the genesis reference intact, swap the genesis state commitment, and assert the proof is rejected.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency Update

## Related Issues

closes #655

🤖 This PR was written primarily by Claude Code and reviewed by Codex.

### DIFF
--- a/crates/proofs/bridge-counterproof/src/statements.rs
+++ b/crates/proofs/bridge-counterproof/src/statements.rs
@@ -123,7 +123,7 @@ fn process_counterproof_inner(zkvm: &impl ZkVmEnv, genesis: &BridgeCounterproofG
             verify_moho_proof(
                 &heavier_moho_state,
                 &heavier_moho_proof,
-                genesis.genesis_moho_state.reference(),
+                &genesis.genesis_moho_state,
                 genesis.moho_vk.clone(),
                 "invalid heavier chain: invalid moho proof",
             );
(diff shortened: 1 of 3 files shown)


uncomputable (approved):
ACK 704f619631857b55ea2825121f18813cd2153345

Rajil1213 (approved):

PR #822, opened 2026-09-14
STR-4292 feat(operator-wallet): persist wallet state via BDK SQLite persister

MdTeach:
## Description

Both BDK wallets were created without persistence, so every start rescanned from genesis. They now persist through BDK's SQLite persister, one file per wallet under the new required `operator_wallet.data_dir`, committing every 2000 blocks during sync. A store that fails to open or belongs to another network or key aborts startup and is left untouched.

Replaces #816. Review fixes follow in #827.

### Type of Change

- [x] New or updated tests


crates/operator-wallet/Cargo.toml
```diff
@@ -8,12 +8,17 @@ workspace = true
 
 [dependencies]
 bdk_bitcoind_rpc.workspace = true
-bdk_wallet.workspace = true
+bdk_wallet = { workspace = true, features = ["rusqlite"] }
```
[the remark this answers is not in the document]
MdTeach: SQLite is compiled into the binary, not loaded from the system. `bdk_chain` turns on rusqlite's bundled feature,

Rajil1213:
Looks like the bulk of the changes are in the first commit itself. Can you break down into smaller, more reviewable commits as well?

MdTeach:
> ## 🔒 AI Security Review (claude-opus-5)
> **Medium — persisted change (internal) descriptor is never validated on load, so a tampered store redirects change outputs**
> 
> `crates/operator-wallet/src/persist.rs:79`
> 
> `load_params()` pins only `KeychainKind::External`, plus network and genesis. In `bdk_wallet` 1.2 `LoadParams::descriptor` is per-keychain, so `check_change_descriptor` stays unset and whatever sits in the `change_descriptor` column of `bdk_wallet` is loaded verbatim as the internal keychain. The honest path never writes one (`Wallet::create_single` at `persist.rs:96` leaves it `None`), so the column is a pure injection slot.
> 
> Exploit: an attacker with write access to `<operator_wallet.data_dir>/general-wallet.sqlite` (the host bind mount `docker/vol/strata-bridge-N/data/wallet/` per the new `docker/README.md`, a sidecar on the same volume, or a restored backup) runs `UPDATE bdk_wallet SET change_descriptor = tr(<attacker_key>);`. `descriptor`, `network`, and genesis are untouched, so every check passes. The wallet now has two keychains, so `TxBuilder::finish()` in `build_v3_psbt` (`crates/operator-wallet/src/general/native.rs:135`) routes change to the attacker, and `sign_claim_funding_tx` (`crates/bridge-exec/src/graph/utils.rs:21`) blind-signs every input without inspecting outputs — draining the general wallet per funding tx. This is newly exposed: before this PR both wallets were rebuilt from the secret-service pubkey on every start, so node-disk tampering could not change where the wallet pays. The key-isolation boundary is bypassed without touching any key material.
> 
> Fix: pin the internal keychain explicitly in `load_params()`, e.g. add `.descriptor(KeychainKind::Internal, Option::<ExtendedDescriptor>::None)`, so any persisted change descriptor fails with `LoadMismatch::Descriptor { keychain: Internal, .. }`. Worth a test alongside the existing External-only cases at `persist.rs:207`.

ACK. Resolving in #827

MdTeach:
> ## 🔒 AI Security Review (claude-opus-5)
> **Medium — persisted change (internal) descriptor is never validated on load, so a tampered store can redirect change outputs**
> 
> `crates/operator-wallet/src/persist.rs:81`
> 
> `load_or_create` only pins `KeychainKind::External` (plus network and genesis hash). BDK's `LoadParams` checks a keychain only when you opt in, so `check_change_descriptor` stays unset and any `change_descriptor` row already present in `bdk_wallet` is loaded and installed as `KeychainKind::Internal` without comparison. Someone with write access to `operator_wallet.data_dir` (a bind-mounted host dir in the shipped docker configs, and a directory operators are documented to shuffle files in and out of during store rebuilds) can take the public external descriptor, write a store that carries it plus an attacker-owned change descriptor, and pass every identity check the PR advertises. Once the general wallet has two keychains, `build_v3_psbt` → `tx_builder.finish()` routes the change output to the internal keychain, the secret service signs the inputs without inspecting outputs, and the change is broadcast to the attacker. This undercuts the stated property that a store "belonging to another key aborts startup", since only half the wallet identity is actually pinned.
> 
> Fix: assert the absence of an internal keychain as well —
> 
> ```rust
> Wallet::load()
>     .descriptor(KeychainKind::External, Some(descriptor.clone()))
>     .descriptor(KeychainKind::Internal, None)
>     .check_network(network)
>     .check_genesis_hash(genesis_block(network).block_hash())
> ```
> 
> `Wallet::create_single` never writes a change descriptor, so a genuine store always has `None` here and the extra check is free; a store carrying one is by definition not this wallet's and should be rejected with the file left untouched, exactly like the network/descriptor mismatches already covered in `fn_wallet_store_recovery_test.py`.

ACK. Resolving in https://github.com/alpenlabs/strata-bridge/pull/827

crates/operator-wallet/src/sync.rs
```diff
@@ -101,11 +101,15 @@ async fn sync_wallet_bitcoin_core(
             let start_height = scan_start_height(&last_cp);
             with_bitcoin_core(client, move |client| {
                 let mut emitter = Emitter::new(client, last_cp, start_height);
-                while let Some(ev) = emitter.next_block().unwrap() {
-                    send_update.send(WalletUpdate::NewBlock(ev)).unwrap();
+                while let Some(ev) = emitter.next_block()? {
+                    // A closed channel means the receiver gave up on this attempt. Stop emitting
+                    // instead of panicking on the dropped receiver.
+                    if send_update.send(WalletUpdate::NewBlock(ev)).is_err() {
+                        return Ok(());
```
Rajil1213, line 153: We should log a warning here.

crates/operator-wallet/src/test_utils.rs
```diff
@@ -0,0 +1,100 @@
...
+    sync::{Arc, Mutex, MutexGuard, PoisonError},
+};
+
+use bdk_wallet::{chain::Merge, AsyncWalletPersister, ChangeSet};
+use thiserror::Error;
+
+/// In-memory store. Clones share one store, so a test can rebuild a wallet from a clone to
+/// simulate a restart. Records every persisted changeset and can fail one chosen `persist` call.
+#[derive(Debug, Clone, Default)]
+pub struct MemoryStore {
```
Rajil1213, line 16: Why not mark it with `#[cfg(test)]` or even move it to `test-utils` or `tests` module?

crates/operator-wallet/src/persist.rs
```diff
@@ -54,50 +61,81 @@ impl<E: std::error::Error + 'static> From<LoadWithPersistError<E>> for InitError
...
-    let create_params = Wallet::create_single(descriptor).network(network);
-    let wallet = PersistedWallet::create_async(store, create_params).await?;
-    info!("created wallet at genesis");
-    Ok(wallet)
+    // Build the whole initial state in memory and commit it once. Committing the wallet and the
+    // checkpoint separately would let a failure in between strand the wallet at genesis: the load
+    // path wins on the next start, so the checkpoint would never be applied.
+    let mut wallet = Wallet::create_single(descriptor.clone())
+        .network(network)
+        .create_wallet_no_persist()
```
Rajil1213, line 96: What happens if we were to not call this method?

crates/operator-wallet/src/persist/sqlite.rs
```diff
@@ -0,0 +1,351 @@
...
+};
+use thiserror::Error;
+
+/// Lock wait before a busy connection errors. A safety net: only one process holds a store.
+const BUSY_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Which operator wallet a store belongs to. Selects the file name, so the two wallets can never
+/// open each other's state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum WalletKind {
```
Rajil1213, line 32: Is this not already defined elsewhere?

Rajil1213 (dismissed):
Only a few nits but otherwise good to go! 🚀

crates/operator-wallet/src/persist/sqlite.rs
MdTeach: It seems there is [SignerTarget](https://github.com/alpenlabs/strata-bridge/blob/07f9d8e19b700766278daf8379cbcd80bb214a2a/crates/secret-service-proto/src/v2/wire.rs#L176-L181) in `secret-service-proto`, but it is a wire type with a `Musig2` variant that has no wallet, and it would pull the protocol crate into operator-wallet.

crates/operator-wallet/src/test_utils.rs
MdTeach: Refactored in [5c32250](https://github.com/alpenlabs/strata-bridge/pull/822/commits/5c322507ba56f899f89a2c841173e7d65ec6de63)

crates/operator-wallet/src/sync.rs
MdTeach: Added in [bd40e00](https://github.com/alpenlabs/strata-bridge/pull/822/commits/bd40e007c47ca29354631e0856c5bba56cab395a)

crates/operator-wallet/src/persist.rs
MdTeach: The builder just returns `CreateParams`, so we require this call to get an actual `Wallet` we can seed with the checkpoint before anything is written.

Alternatively there is `create_wallet_async`, but it persists genesis-only state right away, so the bootstrap checkpoint would have to be written separately afterwards, and if that write fails the wallet stays at genesis forever.

crates/operator-wallet/src/sync.rs
```diff
@@ -98,14 +141,22 @@ async fn sync_wallet_bitcoin_core(
     {
         let client = client.clone();
         async move {
-            let start_height = last_cp.height();
+            let start_height = scan_start_height(&last_cp);
             with_bitcoin_core(client, move |client| {
                 let mut emitter = Emitter::new(client, last_cp, start_height);
-                while let Some(ev) = emitter.next_block().unwrap() {
-                    send_update.send(WalletUpdate::NewBlock(ev)).unwrap();
+                while let Some(ev) = emitter.next_block()? {
```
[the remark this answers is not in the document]
MdTeach: Adding this in https://github.com/alpenlabs/strata-bridge/pull/827

Rajil1213 (approved):
This looks good to go now!

PR #831, opened 2026-09-15
fix(bridge-sm): reject claims that confirm before the graph is signed

MdTeach:
## Description

The claim tx is funded by the graph owner alone, so it can confirm before the rest of the federation has finished signing the graph. The classifier emits `ClaimConfirmed` from the pre-signing states, but `process_claim` fell through to `InvalidEvent` there, which the orchestrator escalates to a fatal `InvariantViolation`: every node lagging behind the owner exited on the same block, and again on restart. Those states now reject the claim and log it.

Zellic finding 3.8, [SECFIND-794](https://alpenlabs.atlassian.net/browse/SECFIND-794).

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

A node still pre-signing holds no signatures, so it cannot contest the claim it just rejected. Whether to record the claim and contest on reaching `GraphSigned`, or to refuse to finish signing a graph whose claim has confirmed, is a protocol decision left for a follow-up.


Rajil1213 (approved):
Looks good but neither the audit finding nor this fix ultimately affects the security of the system as neither the graph nor the deposit will be signed if any of the associated claim transactions are already on chain, and after the graph _is_ signed, the claim will be recorded.

PR #834, opened 2026-09-16
feat(fntests): add manually run full_mosaic group on the real g16 circuit

MdTeach:
## Description

Every functional test runs Mosaic on the bundled toy circuit, whose verdict is just the game-index parity, so nothing in the suite ever shows the garbled circuit adjudicating a counterproof on its merits. This adds a manually run `full_mosaic` group that generates the real g16 SP1-Groth16 verifier circuit in-run from the run's own counterproof vkey and pairs two tests at deposit index 0, where the toy circuit can never NACK: a genuine counterproof that is ACKed and slashed, and one forged from a stale ELF and predicate that the circuit rejects so the operator NACKs it. The circuit artifact and the cut-and-choose parameters are independent knobs. `fn_invalid_counterproof_nackd` passed locally under full parameters (N=181/K=174) in 3h32m, with the honest operator clearing `verify_mosaic_vkey` with dev mode off, which is what STR-3889 was waiting on.

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

The group is excluded from the default sweep and the CI matrix; it needs `sp1-env.bash`, an external regtest bitcoind and about 1 TB of transient disk. The control test `fn_valid_counterproof_acked` has not been run yet.


MdTeach:
@codex review

